PR #420, opened 2022-03-23
UI adjustments

nekhaly:
This PR should cover most of the comments about the UI and its responsiveness.

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -1,15 +1,15 @@
 <template>
-  <div class="flex justify-center items-center pt-16 pb-6">
-    <img class="w-[520px]" src="@/assets/images/logo.svg" alt="logo" />
+  <div class="flex justify-center items-center py-12">
+    <img class="w-[24rem]" src="@/assets/images/logo.svg" alt="logo" />
   </div>
-  <router-view v-if="config" class="flex-auto m-3.5 z-10" />
+  <router-view v-if="config" class="flex-auto z-10" />
   <div v-else class="flex-auto flex flex-col items-center justify-center">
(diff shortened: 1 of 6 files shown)


frontend/src/App.vue
```diff
@@ -1,17 +1,15 @@
 <template>
-  <img
-    class="absolute top-[112px] left-[112px] w-[458px]"
-    src="@/assets/images/logo.svg"
-    alt="logo"
-  />
-  <router-view v-if="config" class="flex-auto m-3.5 z-10" />
+  <div class="flex justify-center items-center py-12">
```
manuelwedler, line 2: I think another breakpoint where decrease the padding of the logo container might be useful for a full hd screen. At the moment a user might still need to scroll

manuelwedler (commented):
Did you check another time with Olympe if the breakpoints and other changes are fine? If so this should be fine

nekhaly:
> Did you check another time with Olympe if the breakpoints and other changes are fine? If so this should be fine

I will have another review shot with her today. We agreed on most of the changes already, but didn't make a final review after merging with your approach!

manuelwedler:
Perfect :)

manuelwedler:
There also seems to be a problem with the CI. I tries to run `yarn lint` in the root folder and not in the `frontend` folder.

frontend/src/App.vue
nekhaly: I made a re-adjustment to avoid the scrolling and reviewed it with @olympe-silve. So everything is good with the current state

manuelwedler (approved):
Perfect :) Ready for the release

PR #445, opened 2022-03-25
Putback chatbot tracking params again

nekhaly:
We removed the tracking utm values from feedback chatbot, and we put  
them again here.


manuelwedler (approved):
LGTM

PR #453, opened 2022-03-28
deploy: Store deployment args in deployment file

palango:
This will make it easier to check depolyments later.

⚠️ As this changes the format of the deployment file, this will make the agent not work until a new depolyment.


fredo:
sounds like a good idea

fredo (approved):

PR #458, opened 2022-03-28
Minor docs tweaks

istankovic:


palango:
What line length are you limiting to? I still see lines above 120 chars.

istankovic:
> What line length are you limiting to? I still see lines above 120 chars.

I did not want to reformat everything as I am not aware of any agreed-upon line length and did not want to get into bikeshedding around max line length. So I just modified what was appeared "too long" to me.

palango (approved):

PR #459, opened 2022-03-28
contracts: mapping for finalization time per target chain id

fredo:
introduce a mapping from for finalization times per chain id.
If not contained target chain id is not supported by the protocol.

### DIFF
--- a/contracts/contracts/RequestManager.sol
+++ b/contracts/contracts/RequestManager.sol
@@ -76,7 +76,6 @@ contract RequestManager is Ownable {
     // Constants
     uint256 public claimStake;
     uint256 public claimPeriod;
-    uint256 public challengePeriod;
     uint256 public challengePeriodExtension;
 
     uint256 public constant MIN_VALIDITY_PERIOD = 5 minutes;
@@ -86,6 +85,8 @@ contract RequestManager is Ownable {
     uint256 public requestCounter;
     uint256 public claimCounter;
     ResolutionRegistry public resolutionRegistry;
+
+    mapping (uint256 => uint256) public finalizationTimes; // target rollup chain id => finalization time
 
     mapping (uint256 => Request) public requests;
     mapping (uint256 => Claim) public claims;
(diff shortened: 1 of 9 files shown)


scripts/deployment/main.py
```diff
@@ -89,7 +89,9 @@ def deploy_l1(web3: Web3) -> dict[str, DeployedContract]:
     return {"Resolver": resolver}
 
 
-def deploy_l2(web3: Web3, resolver: Contract) -> dict[str, DeployedContract]:
+def deploy_l2(
+    web3: Web3, resolver: Contract, supported_target_chains: list[tuple[int, int]]
```
palango, line 93: why not use a simple dict here?

contracts/contracts/RequestManager.sol
```diff
@@ -391,4 +393,9 @@ contract RequestManager is Ownable {
         gasPrice = newGasPrice;
         serviceFeePPM = newServiceFeePPM;
     }
+
+    function addTargetChain(uint256 targetChainId, uint256 finalizationTime) external onlyOwner {
+        require(finalizationTime > 0, "Finalization time of roll up must be greater 0");
```
istankovic, line 398: Nitpick: roll up -> rollup

contracts/contracts/RequestManager.sol
```diff
@@ -391,4 +393,9 @@ contract RequestManager is Ownable {
...
+        finalizationTimes[targetChainId] = finalizationTime;
+    }
```
istankovic, line 400: As this function can be called multiple times for the same chain ID, I would rename it to something like `setFinalizationTime`.

contracts/contracts/RequestManager.sol
```diff
@@ -391,4 +392,9 @@ contract RequestManager is Ownable {
         gasPrice = newGasPrice;
         serviceFeePPM = newServiceFeePPM;
     }
+
+    function setFinalizationTime(uint256 targetChainId, uint256 finalizationTime) external onlyOwner {
+        require(finalizationTime > 0, "Finalization time of rollup must be greater than 0");
```
palango, line 397: Maybe something like "Rollup is not registered" is better? The current error is not really actionable for the user.

fredo: Isn't the only user the owner of the contract, hence we?

Issue #464, opened 2022-03-28
Add imprint text to frontend


nekhaly:
It's required for legal reasons to add imprint text to the app screen.

Imprint text can be copied from here:
https://www.brainbot.com/imprint.html

PR #465, opened 2022-03-29
frontend: Add imprint dialog box to frontend

nekhaly:
Resolves:  
#464

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -9,7 +9,11 @@
     </div>
   </div>
   <feedback></feedback>
-  <footer class="my-8 text-lg text-center text-teal-light">Powered by Beamer</footer>
+  <imprint-modal></imprint-modal>
+  <footer class="my-8 text-lg text-center text-teal-light">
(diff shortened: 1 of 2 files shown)


frontend/src/App.vue
```diff
@@ -9,7 +9,11 @@
     </div>
   </div>
   <feedback></feedback>
-  <footer class="my-8 text-lg text-center text-teal-light">Powered by Beamer</footer>
+  <imprint-modal></imprint-modal>
+  <footer class="my-8 text-lg text-center text-teal-light">
+    Powered by Beamer .
```
istankovic, line 14: Is it intentional that the `.` is here, also with space before it?

istankovic (approved):

frontend/src/App.vue
nekhaly: Yes! At least that is what I have understood from @olympe-silve yesterday

istankovic: @olympe-silve could you please clarify whether this is intended?

PR #504, opened 2022-04-01
Fix uncompleted request on frontend

nekhaly:
After changing configuration it appears that we were using the same  
token address of source rollup for both source and target token address,  
as they were equal anyway, so the issue didn't appear before they are  
different in new condiguration. As there's no way to map source and  
target token addresses, we used the token symbol temporarily for this  
mapping until we figure out another proepr way to make this.  
Also the unnecassry event check at the request end was causing issue,  
it has been removed anyway.

### DIFF
--- a/frontend/src/components/RequestDialog.vue
+++ b/frontend/src/components/RequestDialog.vue
@@ -106,6 +106,11 @@ const { requestTransactionActive, requestState, transactionError, executeRequest
 const { executeWaitFulfilled } = useWaitRequestFilled(beamerConfig);
 const { requestSigner, requestSignerActive, requestSignerError } =
   useRequestSigner(ethereumProvider);
+const getTargetTokenAddress = (targetChainId: any, tokenSymbol: string) => {
+  return beamerConfig.value.chains[targetChainId].tokens.find(
+    (token) => token.symbol === tokenSymbol,
+  )?.address as string;
+};
 
 const feesEther = computed(() => {
   if (fee.value) {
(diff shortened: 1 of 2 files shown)


frontend/src/components/RequestDialog.vue
```diff
@@ -132,7 +137,10 @@ const submitRequestTransaction = async (formResult: {
     targetChainId: Number(formResult.toChainId.value),
     sourceTokenAddress: formResult.tokenAddress.value,
     sourceChainId: Number(formResult.fromChainId.value),
-    targetTokenAddress: formResult.tokenAddress.value,
+    targetTokenAddress: getTargetTokenAddress(
+      formResult.toChainId.value,
+      formResult.tokenAddress.label,
```
palango, line 142: the function takes the symbol, not the address as parameter

nekhaly: this's a select box and the token symbol is the `label`, the address would be `formResult.tokenAddress.value` instead!

palango (approved):

frontend/src/services/transactions/request-manager.ts
```diff
@@ -68,12 +57,6 @@ export async function sendRequestTransaction(
 
   const transactionReceipt = await transaction.wait();
   request.receipt = transactionReceipt;
-  //events is added dynamically when wait method called!
-  // TODO verification of other args?
-  const event = findFirstEvent(request.receipt, 'RequestCreated');
-  if (!event) {
-    //TODO error
-  }
-  request.requestId = event.args.requestId;
```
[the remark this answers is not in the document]
nekhaly: This code has been left from the old implementation.. currently, it's useless because we use another way of detecting fulfillment by listening to the events on the target chain directly. This code was from the old concept of waiting user to change between chains, so no longer needed!

PR #508, opened 2022-04-04
ci: Fix caching

palango:


weilbith (approved):
The CI run proves that it works. Seems like a valid solution. 👍🏾

PR #511, opened 2022-04-04
agent: Handle log range limiting on boba

### LINKED ISSUE #509: exceed maximum block range: 5000 while fetching events
The agent runs into the following error while fetching events

```
ValueError: {'code': -32000, 'message': 'exceed maximum block range: 5000'}
```

Stacktrace comes from `Boba` with this command

```
2022-04-04 09:14:20.307118 [debug    ] Fetching events                [EventFetcher] chain_id=28 contract=RequestManager from_block=42621 to_block=50621
```
We need to reduce the maximum number of blocks to fetch down to 5000

palango:
Resolves #509 

I'm not super happy with the implementation, if you have any idea how to do this in a nicer way let me know.

### DIFF
--- a/beamer/events.py
+++ b/beamer/events.py
@@ -157,15 +157,20 @@ class EventFetcher:
             )
             logs = self._contract.web3.eth.get_logs(params)
             after_query = time.monotonic()
-        except requests.exceptions.ReadTimeout:
+
+        # Boba limits the range to 5000 blocks
+        # 'ValueError: {'code': -32000, 'message': 'exceed maximum block range: 5000'}'
+        except (requests.exceptions.ReadTimeout, ValueError):
             old = self._blocks_to_fetch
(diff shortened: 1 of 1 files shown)


fredo:
I think it is safe to remove the check against the exact `ValueError`.

At least when handled with a `return None`. It doesn't hurt to reduce the number of blocks to fetch, so IMO it's okay to simply reduce based on a Value error.
If we were to differentiate between the exact error, then we should reraise it instead. Like `ValueErrors` are currently handled.

fredo (approved):

palango:
> I think it is safe to remove the check against the exact `ValueError`.
> 
> At least when handled with a `return None`. It doesn't hurt to reduce the number of blocks to fetch, so IMO it's okay to simply reduce based on a Value error. If we were to differentiate between the exact error, then we should reraise it instead. Like `ValueErrors` are currently handled.

I agree, fixed it.

Issue #332, opened 2022-03-11
Distorted select box on request rejection


nekhaly:
After I reject a transaction in Metamask, I get this:

![rejected-tx](https://user-images.githubusercontent.com/1312091/156349468-d9cdf682-7a06-445e-bbb5-6abb695501e4.png)

_Originally posted by @istankovic in https://github.com/beamer-bridge/beamer/issues/249#issuecomment-1056795357_

Issue #507, opened 2022-04-04
Error handling in frontend


nekhaly:
# Rationale

There should be proper error handling for app including errors that happen by user actions or errors that emerged from requests or event listeners.

# Implementation

All errors from different parts of the app should be thrown by a proper `try-catch` statement. There should be a `Map` data structure in the app that contains a code and a user-friendly message if possible that will be shown to the user. In the future, errors will be integrated into the notification system.

weilbith:
The error handling has slightly improved. Though not much.
@nekhaly what are the criteria to call this issue done?

nekhaly:
@weilbith updated the issue with what I imagine as acceptance criteria. Feel free to add/update what fits within this scope! If you don't like it, let's discuss it in the next planning meeting!

taleldayekh:
@manuelwedler can you sit down with @GabrielBuragev and see if this issue still makes sense?

GabrielBuragev:
@taleldayekh
A brief sync-up on the discussion with Manuel.
We believe that this issue still makes sense and should be addressed properly.
We will need to figure out a good approach to generalize all the different wallet provider error messages and on top of that add our own error messages (protocol related).
Once we decide on the specifications of the implementation we will write it down here.

PR #513, opened 2022-04-04
Fix distorted select boxes on rejection

nekhaly:
When user rejects a transaction, select boxes appear distorted. This was  
happening because we were not handling the error properly and ui  
rendring stops before completion.

Resolves:  
#332

### DIFF
--- a/frontend/src/components/RequestDialog.vue
+++ b/frontend/src/components/RequestDialog.vue
@@ -100,8 +100,10 @@ const beamerConfig = injectStrict(BeamerConfigKey);
 const requestMetadata = ref<RequestMetadata>();
 const requestForm = ref<FormKitFrameworkContext>();
 
+const transactionError = ref('');
+
 const { fee, executeGetFee } = useGetFee(ethereumProvider, beamerConfig);
-const { requestTransactionActive, requestState, transactionError, executeRequestTransaction } =
+const { requestTransactionActive, requestState, executeRequestTransaction } =
   useRequestTransaction(ethereumProvider, beamerConfig);
 const { executeWaitFulfilled } = useWaitRequestFilled(beamerConfig);
 const { requestSigner, requestSignerActive, requestSignerError } =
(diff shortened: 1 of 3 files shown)


frontend/src/components/RequestDialog.vue
```diff
@@ -133,32 +135,38 @@ const submitRequestTransaction = async (formResult: {
...
-    amount: formResult.amount,
-    fee: feesEther.value,
-  };
-  request.fee = fee.value;
-
-  await executeRequestTransaction(request, ethereumProvider.value.signer.value);
-
-  await executeWaitFulfilled(request, requestState);
+
+  try {
```
weilbith, line 139: Nitpick: Why did you put so much into the `try` block? Wouldn't it be sufficient to put the `executeRequestTransaction` and `executeWaitFulfilled` in it? In theory the other code should never fail. Buf if it does, I don't think we want to catch it and declare as simple transaction error without a stack trace etc.

frontend/src/utils/create-async-process.ts
```diff
@@ -13,10 +13,15 @@ export default function createAsyncProcess<T extends (...args: any[]) => any>(
   const active: CreateAsyncProcessReturn<T>['active'] = ref(false);
 
   const run: CreateAsyncProcessReturn<T>['run'] = async (...args) => {
-    active.value = true;
-    const result = await fn(...args);
-    active.value = false;
-    return result;
+    try {
```
weilbith, line 16: Suggestion: catching an error can be cumbersome in terms of: who in the call hierarchy does it and when. And alternative approach would be to extend the state of an async process and add an error field. So you can check if the process has finished (not active anymore) and if there is an error. Might be beneficial and possibility allows to have more clean code. Please think about out. Maybe I'm completely off and have no clue what I'm talking about. In this case just resolve the comment thread silently. 😜

PS: if we get started with this, there are probably more improvements here. 😬

weilbith (approved):
I tested it and it works very well. Thank you very much for fixing this unpleasant issue. 🙂 
I have a nitpick and something to think about. But nothing I would categorize as blocking this PR from getting merged. Thereby an optimistic approval that you will act in best manner. 😊

frontend/src/components/RequestDialog.vue
nekhaly: Nothing specific! I have just surrounded the code inside the function with `try-catch` because I used to use it on the whole body and not for the certain parts that may fail. Not a good or a bad habit :D, but wrapped the async calls only as this may make the code more readable!

frontend/src/utils/create-async-process.ts
nekhaly: I cannot say that I'm fully understanding this but in general, I have something against this `createAsyncProcess` that I'm not sure what is it exactly! maybe, in simple words, I feel it makes things over-complicated for my poor mind!  
Anyway, we can discuss all the options that make handling errors a good experience for both developers and users while working on issue #507

frontend/src/components/RequestDialog.vue
weilbith: I don't get the last part. But I assume you made the best decision.

PR #516, opened 2022-04-05
frontend: Remove traget chain event listener after finishing

nekhaly:
We turn on event listener on target chain after a successful transaction  
to listen to fulfillment event, event listener keeps working and send  
unnecessary requests to RPC. It is now removed after sucessful event  
listening or timout.


weilbith (approved):
Simple and efficient. My most favorite PRs. 🚀

PR #519, opened 2022-04-05
agent: Gracefully handle events for ignored claims

palango:


weilbith (approved):
Seems simple enough that I can review this already.
It would be interesting for me what happens with returned boolean. Who cares about them?
I'm optimistic that this change can be merged like that.

palango:
> Seems simple enough that I can review this already. It would be interesting for me what happens with returned boolean. Who cares about them? I'm optimistic that this change can be merged like that.

They are used here: https://github.com/beamer-bridge/beamer/blob/5148323511c6c5c77e5a8914db683886a3d24027/beamer/chain.py#L162-L165=

But that something we need to clean up.

Issue #518, opened 2022-04-05
TransitionNotAllowed: Can't ignore when in Filled.


palango:
```
2022-04-05 08:19:26.861441 [debug    ] Processing request             [EventProcessor] request=<Request id=1 state=filled filler=0x6626079BCF8c3241b082C73B74DFea46CeFA4f02>
2022-04-05 08:19:26.972937 [info     ] Request expired, ignoring      [beamer.chain] request=<Request id=1 state=filled filler=0x6626079BCF8c3241b082C73B74DFea46CeFA4f02>
Traceback (most recent call last):
  File "/root/.shiv/beamer-agent_22e02403e58b1140e099c568bffceb554340aa00312cc080cf4f5237a8237418/site-packages/beamer/chain.py", line 40, in wrapper
    return func(*args, **kwargs)
  File "/root/.shiv/beamer-agent_22e02403e58b1140e099c568bffceb554340aa00312cc080cf4f5237a8237418/site-packages/beamer/chain.py", line 149, in _thread_func
    self._process_requests()
  File "/root/.shiv/beamer-agent_22e02403e58b1140e099c568bffceb554340aa00312cc080cf4f5237a8237418/site-packages/beamer/chain.py", line 198, in _process_requests
    claim_request(request, self._context)
  File "/root/.shiv/beamer-agent_22e02403e58b1140e099c568bffceb554340aa00312cc080cf4f5237a8237418/site-packages/beamer/chain.py", line 327, in claim_request
    request.ignore()
  File "/root/.shiv/beamer-agent_22e02403e58b1140e099c568bffceb554340aa00312cc080cf4f5237a8237418/site-packages/statemachine/statemachine.py", line 61, in __call__
    return self.func(*args, **kwargs)
  File "/root/.shiv/beamer-agent_22e02403e58b1140e099c568bffceb554340aa00312cc080cf4f5237a8237418/site-packages/statemachine/statemachine.py", line 85, in transition_callback
    return self._run(machine, *args, **kwargs)
  File "/root/.shiv/beamer-agent_22e02403e58b1140e099c568bffceb554340aa00312cc080cf4f5237a8237418/site-packages/statemachine/statemachine.py", line 112, in _run
    self._verify_can_run(machine)
  File "/root/.shiv/beamer-agent_22e02403e58b1140e099c568bffceb554340aa00312cc080cf4f5237a8237418/site-packages/statemachine/statemachine.py", line 108, in _verify_can_run
[...]

PR #520, opened 2022-04-05
agent: Allow request transition from filled to ignored 

palango:
Resolves https://github.com/beamer-bridge/beamer/issues/518

This is a strange case where the agent filled a request but didn't claim
in time.
This has been seen on testnet.


nekhaly (approved):
As it looks direct to the goal, while I'm not involved in the development of this, I can say that this can be approved!

PR #522, opened 2022-04-05
frontend: Show error when fulfillment fails

nekhaly:
If agents are down, progress screen was showing success in all cases
including the timeout for fulfillment event listening. We show a proper
error right now if the fulfillment didn't go through to let the use know
 that something wrong has happened.


weilbith (commented):
I think the changes are okay. Is a bit more "involved". But I would really like to test this. At least manually. How do I do this best without shutting down the testnet agents? 😬

nekhaly:
I discovered a defect related to the request id, so I'll push an update for this PR later today!

frontend/src/components/RequestDialog.vue
```diff
@@ -183,7 +183,7 @@ const isNewTransferDisabled = computed(() => {
 
 const shownError = () => {
   const error = requestSignerError.value || transactionError.value;
-  if (error) {
+  if (error && requestState.value !== RequestState.RequestFailed) {
```
manuelwedler, line 186: Shouldn't the user be able to start a new transfer in the error case? If I'm not mistaken, the `NewTransfer` button stays disabled now.

And not related to this PR, but I think this function should not modify state. It should actually be a computed property to make use of vue's reactivity system. To modify the state below it would be better to have a separate watcher.

frontend/src/composables/useRequestTransaction.ts
```diff
@@ -122,8 +122,8 @@ export function useWaitRequestFilled(beamerConfig: Ref<Readonly<BeamerConfig>>)
       await waitOnFulfillment();
       requestState.value = RequestState.RequestSuccessful;
     } catch (error) {
-      console.log(error);
```
manuelwedler, line 125: I think it is helpful to log the full error, but maybe not here and instead where it is actually caught. And `console.error` should rather be used.

frontend/src/services/transactions/fill-manager.ts
```diff
@@ -30,13 +30,13 @@ export async function listenOnFulfillment(
...
     setTimeout(() => {
-      reject(new Error('Timeout!'));
+      reject(new Error('Timeout while waiting for fulfillment acknowledgement!'));
     }, eventListeningTimeout);
   });
 
   return Promise.race([fulfillmentPromise, timeoutPromise])
-    .then((fullRequestId) => {
-      return fullRequestId;
+    .then((fillRequestId) => {
```
manuelwedler, line 38: What's the point of this `then` call?

frontend/src/services/transactions/request-manager.ts
```diff
@@ -14,13 +14,18 @@ export async function getRequestFee(
   const connectedContract = ethereumProvider.connectContract(requestManagerContract);
   return await connectedContract.totalFee();
 }
+interface TransactionReceiptInterface extends TransactionReceipt {
+  events: Array<any>;
```
manuelwedler, line 18: Is this really working? According to the ethers documentation, there should be a `logs` property on the `TransactionReceipt`.

manuelwedler (commented):
Just a couple of comments. I didn't test this though.

frontend/src/components/RequestDialog.vue
nekhaly: > Shouldn't the user be able to start a new transfer in the error case? If I'm not mistaken, the `NewTransfer` button stays disabled now.
You are right.. Didn't notice that while implementing.. TY

> And not related to this PR, but I think this function should not modify state. It should actually be a computed property to make use of vue's reactivity system. To modify the state below it would be better to have a separate watcher.

I am not sure if I got exactly what do u mean but talking about Vue, then you are the expert here 😁!
I think you are talking about getting rid of `showError` at all as a function and converting it into a computed property that will be watched to change `requestState` there. Correct me if I'm wrong.
Anyway, I have just used the current implementation without adding fancy things.. So just added a condition to prevent go to `Init` state if a failure is there because in future we will have information in progress screen like transaction Id and such on that the user will need to be there to trap them for further support!

manuelwedler: Ok, forget about the button. I misread the `isNewTransferDisabled` condition above. The button shouldn't be disabled. So it's fine.

I would wrap the function like this:
`const shownError = computed(() => ...)`

And remove the check from there and add a watcher:
`ẁatch([shownError], () => <here the check and state change>`

(Computed properties are not allowed to modify state.)

frontend/src/services/transactions/request-manager.ts
nekhaly: `logs` doesn't contain events information, `events`  property does in its `args` and unfortunately it isn't available in  `TransactionReceipt` interface!

frontend/src/services/transactions/fill-manager.ts
nekhaly: Nothing! just the default `Promise` tree of `then-catch-finally` because I need `catch` and `finally` implementations! I removed it!

frontend/src/composables/useRequestTransaction.ts
nekhaly: I think it will be logged anyway in the console if it's a Metamask error (which is the current case here)!

weilbith (approved):
I think this is good for now. As already discussed this will be replaced with a better solution for the user once we have a request history available.
Good job. Thank you. 👍🏾

PR #524, opened 2022-04-06
Move code

palango:
Based on #515

Improve code structure


weilbith (approved):
Looks simple. I can't see any "mistakes".
If you are already on it: maybe also split `models.py` for `claim.py` and `request.py` and put them in a shared `models/` directory? 🤷🏾

Issue #356, opened 2022-03-16
Agent: Garbage collect and filter unused requests and claims


fredo:
We started with removing unfillable requests from the `RequestTracker`. We can improve on that by removing all information which are not of interest for the agent.
I would like to raise this as its own issue because we need to be careful about what information are not of interest for the agent. I.e. an expired request can still be relevant if there are challenged (or even pending) claims to it where the informaiton might be interesting for the agent (part of the challenge).
We can use this issue to come up with a good strategy.

### Todo
- [x] Think through that anything that is removed won't be necessary for the agent, e.g. never remove requests when there is an active claim.
- [x] Remove "all information which are not of interest for the agent".

fredo:
My concerns in the refinement meeting are derived from the architecture @istankovic and me discussed last week back in the office.
We agreed that the state machines of claims and requests should be viewed as two separate state machines running in parallel completely detached from each other. Since claims are determined by other users, theoretically claims can arrive at any state of the request (but withdrawn). Although the request is already withdrawn, there could still be active claims, which need the requests internal information.

# Rules of the interaction of claims and requests
In order to keep control over the state machines we determined a few rules which we must follow in the future.

1) Claims have read access to requests' states and vice versa.
2) Inside of processing one state machine we should never write and/or trigger state changes of the other. We decided that to be able to retrace how a claim or request arrived in this state. Ideally in the future, state changes should only be triggered by events. Currently we accept the exception that we allow internal state changes when processing requests or states.


### Why we need to be cautious
Although, we have a final state for requests and claims (namely the withdrawn state), we do allow read access between the state machines (as described above). Thus, we cannot simply remove an object when it arrives in its final state, because the other state machine might need to read from this object still.

#### Example: Active claims
In theory, we can have an infinite number of active claims for a given request. While the request can be withdrawn by a successful claim OR L1 resolution, this does not withdraw all other active claims. In the agent's internal state this means, that there can exist active claims, which are not yet in the withdrawn state, but the request is already withdrawn and in its final state.
Those active claims might need to read the state of the finalized request still. We need to respect this when it comes to removing information from the internal state.

### What to do now?
We should take a close look to the claim and request state machine and check the times when there is read access between them. Then, formulate conditions when removal of information is safe. Implement these conditions when it comes to removal (final state).

palango:
For claims I only find this reference on the request, which is why we need to wait with removal of a certain request until all related claims are removed.
[...]

PR #525, opened 2022-04-06
Testability

palango:
Based on #515

- Make some functions free, so that they can be used in tests.
- Use those functions to test the removal of old requests and claims

Resolves https://github.com/beamer-bridge/beamer/issues/356

### DIFF
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -186,6 +186,7 @@ class EventProcessor:
     # TODO: pull this out of the event processor, so it can be tested independently
     def _process_requests(self) -> None:
         assert self._synced, "Not synced yet"
+        self._log.info("Processing requests", num_requests=len(self._context.requests))
 
         to_remove = []
         for request in self._context.requests:
@@ -210,16 +211,13 @@ class EventProcessor:
 
     # TODO: pull this out of the event processor, so it can be tested independently
     def _process_claims(self) -> None:
-        with self._lock:
-            if self._num_syncs_done < 2:
-                # We need to wait until we are synced with both chains.
-                return
-
-        to_remove = []
+        assert self._synced, "Not synced yet"
+        self._log.info("Processing claims", num_claims=len(self._context.claims))
 
         block = self._context.request_manager.web3.eth.get_block("latest")
         latest_timestamp = block["timestamp"]
 
+        to_remove = []
         for claim in self._context.claims:
             self._log.debug("Processing claim", claim=claim)
 
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -146,8 +146,9 @@ class EventProcessor:
                 self._process_events()
 
             if self._synced:
-                self._process_requests()
-                self._process_claims()
+                process_requests(self._context)
+                process_claims(self._context)
+
         self._log.info("EventProcessor stopped")
 
     def _process_events(self) -> None:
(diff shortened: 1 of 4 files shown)


weilbith (commented):
Mistake...

beamer/chain.py
```diff
@@ -183,97 +184,85 @@ def _process_events(self) -> None:
...
-                    to_remove.append(request.id)
 
-        for request_id in to_remove:
-            self._context.requests.remove(request_id)
+def _transact(func: web3.contract.ContractFunction, **kwargs: Any) -> web3.types.HexBytes:
+    try:
+        tx_hash = func.transact(cast(Optional[TxParams], kwargs))
+    except (web3.exceptions.ContractLogicError, requests.exceptions.RequestException) as exc:
+        raise _TransactionFailed() from exc
+    return tx_hash
```
weilbith, line 204: Nitpick:

```suggestion
   try:
        return func.transact(cast(Optional[TxParams], kwargs))
    except (web3.exceptions.ContractLogicError, requests.exceptions.RequestException) as exc:
        raise _TransactionFailed() from exc
```

beamer/chain.py
```diff
@@ -183,97 +184,85 @@ def _process_events(self) -> None:
             if not any_state_changed:
                 break
 
-    # TODO: pull this out of the event processor, so it can be tested independently
-    def _process_requests(self) -> None:
-        assert self._synced, "Not synced yet"
```
weilbith, line 188: This line is missing from the new function. But honestly I don't even get why we added it, if the caller ensures we are synced anyways. So was that the reason you removed it?

beamer/chain.py
```diff
@@ -183,97 +184,85 @@ def _process_events(self) -> None:
...
 
-    def __str__(self) -> str:
-        return "transaction failed: %s" % self.cause()
+        if claim.transaction_pending:
+            continue
 
-    def cause(self) -> str:
-        return str(self.__cause__)
+        block = context.request_manager.web3.eth.get_block("latest")
+        if claim.termination <= block["timestamp"]:
```
weilbith, line 258: It was `latest_timestamp >= claim.termination` before and now `claim.termination <= block["timestamp"]`. So I think to match this you need to remove the `=`.

weilbith (commented):
TESTS 🥳 
Nice job 👍🏾 

Depending on the outcome of [this comment](https://github.com/beamer-bridge/beamer/pull/515#discussion_r843986815) of the "parent" PR, the tests need to change. Though I'm optimistic that this will not be the case. 😆

beamer/chain.py
```diff
@@ -183,97 +184,85 @@ def _process_events(self) -> None:
...
 
-    # TODO: pull this out of the event processor, so it can be tested independently
-    def _process_requests(self) -> None:
-        assert self._synced, "Not synced yet"
-        self._log.info("Processing requests", num_requests=len(self._context.requests))
-
-        to_remove = []
-        for request in self._context.requests:
-            if self._stop:
-                break
```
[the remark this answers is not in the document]
palango: Yeah, sharp eyes. The issue this solved for me is that when the loop for processing requests and claims, I cannot stop the agent.

so we could pass in the event to make this possible again. do you thing this is worth it?

beamer/chain.py
palango: 👁️

beamer/chain.py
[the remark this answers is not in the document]
palango: >  I would expect it to just take longer until the loop is done.

Exactly what I mean. But with some caching and other improvements this time should go down a lot.

weilbith (approved):
I think it should work for the moment. 👍🏾

Issue #477, opened 2022-03-30
Add caching to agent


palango:
Currently eth rpc requests are blocking. a simple solution to increase performance is thus to use more aggressive caching,

palango:
Fixed for now by #532 and #514

PR #532, opened 2022-04-07
agent: Cache latest block data

palango:
Currently black data are queried from RPC in handlers when required.
This is ineffective, as most of these requests will get the same return
value.

Instead, we add a mapping from chain id to block data to the `Context`,
in which the latest known block for each chain is stored. It is updated
via the new `LatestBlockUpdated` event.

Part of https://github.com/beamer-bridge/beamer/issues/477

### DIFF
--- a/beamer/events.py
+++ b/beamer/events.py
@@ -8,7 +8,7 @@ import web3
 from eth_abi.codec import ABICodec
 from eth_utils.abi import event_abi_to_log_topic
 from web3.contract import Contract, get_event_data
-from web3.types import ABIEvent, ChecksumAddress, FilterParams, LogReceipt, Wei
+from web3.types import ABIEvent, BlockData, ChecksumAddress, FilterParams, LogReceipt, Wei
 
 from beamer.typing import (
     BlockNumber,
@@ -24,6 +24,11 @@ from beamer.typing import (
 @dataclass(frozen=True)
 class Event:
     chain_id: ChainId
+
+
+@dataclass(frozen=True)
+class LatestBlockUpdatedEvent(Event):
+    block_data: BlockData
 
 
 @dataclass(frozen=True)
@@ -177,6 +182,7 @@ class EventFetcher:
             self._log.error("Connection error", url=url, exc=exc)
             # Propagate the exception upwards so we don't make further attempts.
             raise exc
+
         else:
             duration = after_query - before_query
             if duration < EventFetcher._ETH_GET_LOGS_THRESHOLD_FAST:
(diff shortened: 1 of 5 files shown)


beamer/events.py
```diff
@@ -185,6 +191,11 @@ def _fetch_range(
                 self._blocks_to_fetch = max(EventFetcher._MIN_BLOCKS, self._blocks_to_fetch // 2)
             codec = self._contract.web3.codec
             events = _decode_events(logs, codec, self._chain_id, self._event_abis)
+            events.append(
```
fredo, line 194: Shouldn't it be sufficient to have one `LatestBlockUpdated` per `fetch()` call? By having this in `fetch_range()` it is possible to have multiple calls per fetch(). It wouldn't harm logic wise but is less efficient.
I was thinking putting it somewhere around [this line](https://github.com/beamer-bridge/beamer/blob/5148323511c6c5c77e5a8914db683886a3d24027/beamer/events.py#L209)

fredo (commented):
LGTM, should solve most of our current problems. I have some suggestions below.

beamer/events.py
fredo: And additionally, whereever you place this call we need to make sure to also handle exceptions during this call. I think currently it would bubble up until the very top of the thread_func. 
We should handle it like the other calls, maybe take a look at https://github.com/beamer-bridge/beamer/blob/5148323511c6c5c77e5a8914db683886a3d24027/beamer/events.py#L191. 

What is important to me is that we guarantee that this event is in the list of events. Otherwise we break an invariant that we always need to have `latest_blocks` set before each iteration over claims and requests.

palango: Good point, I fixed this by adding the block data request to the inital block number request. This way we do not change internal state when the request fails/

beamer/events.py
```diff
@@ -207,5 +214,13 @@ def fetch(self) -> list[Event]:
                     from_block = BlockNumber(to_block + 1)
 
             self._next_block_number = from_block
+
+            result.append(
```
fredo, line 218: thanks for the change. couldn't we move this directly into line 204?

beamer/events.py
```diff
@@ -207,5 +214,13 @@ def fetch(self) -> list[Event]:
...
+                LatestBlockUpdatedEvent(
+                    chain_id=self._chain_id,
+                    block_data=block_data,
+                )
+            )
```
fredo, line 223: in line 204

```suggestion
           result = [LatestBlockUpdatedEvent(
                    chain_id=self._chain_id,
                    block_data=block_data,
                )
            ]
```

beamer/events.py
palango: I'd rather do the update of the latest block after all events that happened before are processed. I don't see any problem right now, but think that's the more sensible approach.

palango: Merging for now, if you disagree we can fix this later.

PR #545, opened 2022-04-11
agent: Query block data later when getting events

palango:
This will save some RPC requests when no new blocks are available.

As commented from @istankovic

### DIFF
--- a/beamer/events.py
+++ b/beamer/events.py
@@ -195,7 +195,6 @@ class EventFetcher:
     def fetch(self) -> list[Event]:
         try:
             block_number = self._contract.web3.eth.block_number
-            block_data = self._contract.web3.eth.get_block(block_number)
         except requests.exceptions.RequestException:
             return []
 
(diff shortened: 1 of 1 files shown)


beamer/events.py
```diff
@@ -214,8 +213,13 @@ def fetch(self) -> list[Event]:
                 result.extend(events)
                 from_block = BlockNumber(to_block + 1)
 
-        self._next_block_number = from_block
+        try:
+            # Block number needs to be decremented here, because it is already incremented above
+            block_data = self._contract.web3.eth.get_block(from_block - 1)
```
istankovic, line 218: We can instead use just `block_number`, which stays constant and is the loop end condition.

palango: No, because we might `break` in the while loop, this would be incorrect.

beamer/events.py
```diff
@@ -214,8 +213,13 @@ def fetch(self) -> list[Event]:
...
+        except requests.exceptions.RequestException:
+            return []
 
+        self._next_block_number = from_block
         result.append(
             LatestBlockUpdatedEvent(
                 chain_id=self._chain_id,
```
istankovic, line 225: I would rather do something like this:

```
        self._next_block_number = from_block
        try:
            block_data = self._contract.web3.eth.get_block(block_number)
        except requests.exceptions.RequestException:
            return result
        else:
            result.append(LatestBlockUpdatedEvent(...))
        return result
```

So that we return all the collected events even if the last RPC call to get block data fails.

palango: 👍

istankovic (approved):

Issue #497, opened 2022-04-01
Implement proper state management for frontend


nekhaly:
# Problem

For app performance's sake, in case of, for example, we would like to retrieve historical events for the user to allow checking old transactions, a cached state for the events will help much in decreasing querying blockchain for those old ones. A proper state management tool would help in that!

#Solution

`Vuex` is the direct suggestion to achieve this, but we may look also for other solutions

manuelwedler:
Just a remark: For vue3, [Pinia](https://pinia.vuejs.org/) is the go-to state management library now. Vuex is in maintenance mode.

Issue #384, opened 2022-03-17
frontend: Add full test setup and tests


manuelwedler:
We don't have any testing at the moment for the frontend. As a bare minimum we should have Unit Tests for all components, modules, etc.
We should use [Vitest](https://vitest.dev/) as a framework because it is recommended for Vue 3.

### Acceptance Criteria

Should have setup for:

- [ ] Unit testing
- [ ] Vitest, component testing
- [ ] Cypress, e2e testing

weilbith:
We have some goodish experience with Cypress for end-to-end tests.

taleldayekh:
@weilbith Can you split into separate issue.

nekhaly:
@weilbith I think this issue acts as an epic for issues you have opened, so I think we should add issues to it, and then it will close automatically (not sure about this from Zenhub) when all child issues are closed!

PR #541, opened 2022-04-08
chore: Add linting to frontend CI checks

manuelwedler:

PR #548, opened 2022-04-11
Frontend/feature/replace configuration provider with store

weilbith:
Related to #497 and #384 

This adds the first usage of a store for the configuration of the app. First we intend to replace the current provide-inject usage. Later we will use stores to persists data for sessions or permanently and also add much more reactive, shared state.

Furthermore this finally makes the initial setup for the unit tests. So far the only tests are for the new configuration store. More configuration for coverage etc. will come.


weilbith:
I should move the `.gitiginore` changes into the `frontend` it's local file.

weilbith:
And I'm still unsure about the composition usage. Unfortunately I haven't worked too much with it yet. So I'm lacking a bit in best practice and deep knowledge. So my current question is: where should the dirty glue code go? Should mean, it is better to keep the `composables/*` clean from the stores and other composables and provide everything as argument instead (kinda as before). But then the ugly glue code is within the components. And this also means that the nice to reuse composable code is not so easy to use anymore as the components have to use all other composables and stores too and it has to be repeated everywhere. That is why I'm currently having it like it is right now, keeping the components clean. They will be tough enough to test anyway. And the mocking of the stores will be the same for the composables, just that we have to do it way less. 
I hope the others can agree with me or provide some nice input for best practices please. 🙂

weilbith:
I'll now rebase to resolve the issues which got introduced by the merging of #541

weilbith:
Regarding my above comment:
This morning I had a call with @nekhaly where we discussed the topic. We both agreed to stick with the current approach and continue with it as there are good enough arguments for it. But we will have another discussion together with @manuelwedler about this. He has written the former code and seems to be much more familiar with the recent changes in the community and best practices. In case he has much more good arguments for the opposite approach, the changes are very limited and not too much. Therefore it okay to continue with the current state to work around Manuel his working times.

weilbith:
Due to the rebase on `main`, the CI checks now fail due to linting warnings. I'll fix them and split them into fixup commits.

frontend/tests/utils/data_generators.ts
```diff
@@ -0,0 +1,44 @@
...
+export function generateChainConfiguration(
+  partialConfiguration?: Partial<ChainConfig>,
+): ChainConfig {
+  return {
+    requestManagerAddress: getRandomEthereumAddress(),
+    fillManagerAddress: getRandomEthereumAddress(),
+    explorerTransactionUrl: getRandomUrl('explorer'),
+    rpcUrl: getRandomUrl('rpc'),
+    name: getRandomString(ALPHABET_CHARACTERS, 8, 'name-'),
+    token: generateToken(),
```
nekhaly, line 41: A little bit tricky here, but shouldn't that be an array of `tokens` to match configuration? something like `tokens: [generateToken()]`

nekhaly (approved):
Looks amazing and clean. Approved with a small comment about unit testing data!

nekhaly:
Unfortunately, I couldn't rebase that locally to test with the latest on `main`. So you can test again after rebasing that everything is good before merging!

frontend/tests/utils/data_generators.ts
weilbith: ✔️

weilbith:
The review feedback got addresses (thanks @nekhaly for detecting this) and pushed. I'll now rebase on `main` to resolve the conflict. While doing that I'll also resolve all the fixup commits I made as I'm optimistic that after the required test, it is good to merge then.

weilbith:
Alright. I completed everything. The manual test worked too. I additionally used the respective roll-up explorers to verify the transactions and token balances. Once the CI pipeline has completed (somehow takes long right now), I'll go and merge this PR.

PR #551, opened 2022-04-12
add new base workflow to block fixup commit merges

weilbith:
Using Git's feature of fixup commits is a good practice for pull requests. But all fixup commits must be always resolved first before merging the pull request finally. To ensure nobody can forget this, it is always very useful to have the continuous integration pipeline to verify this.

I'm completely open to rename the workflow from `base` to anything else. No strong opinion here. I thought about something like `shared` or so. 🤷🏾

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -1,0 +1,14 @@
+name: Base CI
+
+on:
+  pull_request:
+    branches: 
+      - main
+
+jobs:
+  block-fixup-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.0.0
+      - name: 'Block Merging Fixup Commits'
+        uses: 13rac1/block-fixup-merge-action@v2.0.0


palango (approved):
Nice

istankovic:
Thanks @weilbith for submitting this.

As far as I understand, this action will only prevent a subset of "fixup" commits -- the ones that were done via `git commit --fixup`. I'm not sure if I ever saw those in our PRs -- I think fixup commits made in a regular `git commit` fashion are much more common, and there is no way to automate checking against those. In those cases, I typically tend to point that out during the review process.

weilbith:
@istankovic ofc it can only apply for "proper" fixup commits. The practice of developers to do "manual" fixup commits with messages like "improve function by pull request review" needs to be tackled on a different level and can barely be solved by a technical solution. One thing that could help is to force the usage of [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) (which are interesting in general) and an according [linter](https://github.com/conventional-changelog/commitlint) to enforce this.

istankovic:
Yeah, I agree in general, though I'd be wary of being very strict about everything -- I think discussing _why_ we should do certain things and not others is rather more useful in helping everyone understand and is in practice more pleasant. And yes, as you said, this is not a thing to be solved on a technical level.

weilbith:
Sounds good. This is not yet an actual open source project with a community. In such a case technical restrictions where possible are very helpful to have continuous quality guarantees without too much human management effort. But that is not (yet) the case. So talking in the team and making an agreement can be good enough.

PR #552, opened 2022-04-12
agent: GC ignored requests as well

palango:
This was forgotten when implementing GC and brings us from 3000 active
requests down to 20 requests on testnet.

### DIFF
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -216,10 +216,10 @@ def process_requests(context: Context) -> None:
         elif request.is_filled:
             claim_request(request, context)
 
-        elif request.is_withdrawn:
+        elif request.is_withdrawn or request.is_ignored:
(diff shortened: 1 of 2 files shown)


beamer/chain.py
```diff
@@ -216,7 +216,7 @@ def process_requests(context: Context) -> None:
         elif request.is_filled:
             claim_request(request, context)
 
-        elif request.is_withdrawn:
+        elif request.is_withdrawn or request.is_ignored:
             active_claims = any(claim.request_id == request.id for claim in context.claims)
             if not active_claims:
                 log.debug("Removing withdrawn request", request=request)
```
istankovic, line 222: Nitpick: we should consider updating the log message now that also ignored requests are removed.

istankovic (approved):

Issue #492, opened 2022-03-31
validUntilBlock and validUntil


fredo:
Outsourced discussion from #346 

# Summary
The idea is to add a `validUntilBlock` parameter to the request. `validUntil` must stay in order to be able to withdraw expired requests by the user. The problem is that we cannot validate the `validUntilBlock` upon creating the request so it could be any value even in the far future.
It still makes sense to have it to prevent an attack scenario.

# Functionality
We can allow a value of `validUntilBlock` for a maximum of N blocks. If a misbehaving user submits a block number which `validUntilBlock > block.number + N` on the target rollup we can invalidate this variable. It means the user misbehaved.
The request can only be filled in `fillPeriod = validUntilBlock - N < x < validUntilBlock`. 
The request is going to be invalidated if I can trigger it before `validUntilBlock - N`. The request cannot be filled anymore.

# Attack scenario
The functionality enforces protection for a misbehaving user.

1) Alice accidentally creates a request with `validUntilBlock = targetRollupBlock.number + 1000000`
2) Charlie claims it with `filledBlock = targetRollupBlock.number + 1000000 - 1`
3) None can send a Non-filled Proof + None can even fill it
4) If Charlie remains the highest bidder he wins the deposit

By invalidating the request we can send a `invalidRequestMessage` via L1 and it behaves like the Non-Fill Message. I have not found a way to make it possible that someone can fill it correctly once the fill period started. But I think as soon as the user misbehaves it is in the best interest for all to simply invalidate the request at all.

# Choosing correct values
I think we do not have race conditions if the `validUntilBlock` is small enough. If Charlie sends the "future fill" right before the fill period starts and Bob is not able to invalidate anymore then he must be able to send the Non-fill Proof. Since Charlie has the following constraint checked by the RequestManager `filledBlock < validUntilBlock`, the maximum waiting time for the Non-FillProof will be `N` blocks (maybe a couple of more due to the race condition). If N is much smaller than `ChallengePeriodExtension` this will leave enough time for Bob to trigger the Non-Fill proof. The constraint is `N_target_blocks << ChallengePeriodExtension`.

fredo:
Result from discussion wit @cducrest:

We need a validity period to reason about the non-fill proof. Anyone could claim a block he filled a request which is very far in the future. Even that far away that the non-fill proof will not arrive in time upon end of a challenge of a false claim.

[...]

Issue #346, opened 2022-03-15
L1 non-fill proof


cducrest:
We had this discussion with Fred about the following problem:

> What happens if `Charles` self-challenges a false claim on `Alice's` request that cannot be filled (e.g. she request a transfer for an unsupported token / too high a value / off-chain sequencer ....). Then no correct claim will be able to terminate before `Charles's` claim and he will be able to withdraw `Alice's` deposit.

A potential solution is to have some non-fill proof submitted from destination roll-up to source roll-up via L1 proving that no fill for a corresponding request hash was done. Here's how it would look:

1) Alice sees Charles’ false claim. Observers knows that the fill ID is false, and the the request is not filled by anyone else.
2) Bob triggers an L2 -> L1 message which states that the request is not filled yet. This needs to be triggered `finalization_time` of target roll-up before Charles' claim end.
3) The non-fill proof submitted to source roll-up should enable Bob to earn the deposit of Charles, to incentivize Bob's participation.
4) Charles should not be able to withdraw its stake or Alice's deposit.
5) Alice can withdraw her deposit if the `expiry` time on her request ended.

Bob's intervention is only necessary if Charles challenged and counter-challenged himself, meaning that the result would be that his claim over Alice's deposit is successful. This means that we need to enforce that the counter-challenge stake should be sufficient to incentivize Bob's participation to submit the L1 non-fill proof.

We need to either find a better idea or implement it.

cducrest:
Either we allow non-fill proofs to be submitted at any time, and we don't add the `valid until block` field to the request hash. The proof will need to have the block number at which it was submitted, and it will need to be registered in the resolution registry for the request manager to check whether the non-fill proof was submitted too early.

Or non-fill proofs cannot be submitted before `valid until block` as verified in the fill manager, so the `valid until block` must be part of the request hash (and the fill call).

Second solution seems easier to implement, you are right (albeit more gas intensive in the optimist case, but I don't think we care).

fredo:
Is it really more gas efficient? 

I understand that we must add the `valid until block` in 1) but at the same time we would need to add the `filled block number` upon any claim. This would be necessary in order for an honest claim not to be blocked by the non-fill-proof message. Furthermore we would need to pass the non-fill block number via L1 resolution. 

[...]

PR #555, opened 2022-04-12
Add non-fill proofs to whitepaper

cducrest:
See issue #346 and #492 for discussion

Idk if we should keep the idea of parallel claims to solve the "self-challenge" problem in the case the request was honestly filled. We could think about replacing it with the honest filler joining the challenge game as a third party. I'm fine with w/e.

### DIFF
--- a/docs/source/whitepaper.rst
+++ b/docs/source/whitepaper.rst
@@ -111,8 +111,8 @@ When filling the request, a `fill ID` is also computed, that serves to identify 
 `fill ID` can be found in section :ref:`fill_id`.
 
 After filling the request, Bob will immediately claim the refund on rollup A. In doing so, he submits the `request ID`,
-and `fill ID` to rollup A. This initiates a `claim period` during which the validity of the fill by Bob
-can be contested. Bob is required to send a deposit `claim stake` in rollup A's coin, that will be lost if
+and `fill ID` to rollup A. This initiates a `claim period` during which the validity of the fill by
+Bob can be contested. Bob is required to send a deposit `claim stake` in rollup A's coin, that will be lost if
 his claim is proven to be eventually incorrect.
 
 If any participant contests Bob's claim, the protocol will immediately enter the challenge game. This
@@ -159,9 +159,9 @@ him and the challenger, Charles. Charles needs to stake a deposit higher than `c
 The challenge will be on-going until the end of the `challenge period`.
 
 During the challenge, the contested participant (in turn Bob, then Charles), can submit a transaction to confirm its
-position and contest the other party. It is required that the new stake of the participant is higher than the current stake of the opponent.
-Everytime a participant responds to the challenge, the termination time of the challenge and underlying claim is extended to be at least
-`challenge period extension`, to give time for the other party to respond.
+position and contest the other party. It is required that the new stake of the participant is higher than the current
+stake of the opponent. Everytime a participant responds to the challenge, the termination time of the challenge and
+underlying claim is extended to be at least `challenge period extension`, to give time for the other party to respond.
 
 At the end of the challenge period, the last non-contested participant, and thus the participant with the highest stake, wins. The claim
 will be seen as valid if the winner of the challenge game is the original claimer. This means that he will be able to
@@ -205,8 +205,8 @@ rollup B on L1. This proof is a call to a `resolver` contract on L1 and contains
 To trigger L1 resolution is to apply this call on L1 using the data from the rollup B's outbox. This will forward the
 information from the resolver to the inbox of rollup A in the form of a call to the `resolution registry`.
 This registry will store in its state a mapping from `fill hash` to `Bob`, allowing the `request manager`
-to verify that a claim to fill a certain request with a certain fill ID is honest. Rollup A's chain ID is necessary for the
-`resolver` contract to know to which `resolution registry` to forward the proof to. Rollup B's chain ID is used to
+to verify that a claim to fill a certain request with a certain fill ID is honest. Rollup A's chain ID is necessary for
+the `resolver` contract to know to which `resolution registry` to forward the proof to. Rollup B's chain ID is used to
 restrict the call to authenticated `fill manager` and `cross domain messenger` contracts.
 
 After L1 resolution has transferred the fill information from rollup B to rollup A, Bob can directly call `withdraw` on
@@ -245,11 +245,11 @@ the challenge and receive his stake as well as Charles' stake, the tokens locked
 Why do we need the fill ID?
 +++++++++++++++++++++++++++
 
-The reason a claimer needs to submit a `fill ID` is to make a statement of when the related request was filled. It is
+The reason a claimer needs to submit a `fill ID` is to make a statement as to when the related request was filled. It is
 returned by the `FillManager` contract on rollup B and there will always be only one valid `fill ID` to a fill of a
 requests. Enforcing a submission of an ID, certain attacks on honest challengers are prevented. Without this ID, an
 evildoer could claim an unfilled request and only fill it once its claim is challenged thus turning it into a rightful
-claim and gaining the challenger's stake. The `fill ID` is defined as
+claim and gaining the challenger's stake. The `fill ID` is defined as:
 
 ::
 
@@ -322,10 +322,9 @@ In any case, this condition will always be fulfilled if Bob fills the request be
 Self challenges
 +++++++++++++++
 
-To make the protocol easier to reason about and implement, only two actors can participate in a challenge: the original
-claimer, and the initial challenger. This raises the concern that, after submitting his false claim, Charles could challenge
-himself to prevent anyone from challenging him. This would let Charlie control the state of his challenge and he would be able to
-let it expire with his claim successful as an outcome.
+In the previous explanation, we only talked about two participants to the challenges, Bob and Charles. What if, after
+submitting his false claim, Charles challenges himself. This would let Charles control the state of his challenge and
+he would be able to let it expire with his claim successful as an outcome.
 
 To prevent this successful `self-challenged claim` to allow Charles to withdraw Alice's deposit, Bob can fill Alice's request
 and do his own claim in parallel. If Bob's claim is not challenged and `claim period` is lower than the `challenge period`,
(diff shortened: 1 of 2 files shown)


docs/source/whitepaper.rst
```diff
@@ -335,20 +334,82 @@ Charles can attempt to delay Bob's withdrawal by challenging Bob's rightful clai
...
+
+There is no need for Bob to enter the challenge if the claimer is in a losing position, we only allow new challengers if
+the claimer is ahead in the challenge. This is done by a bookkeeping mechanism of bidders and bids. We store in a mapping
+who bid which amount in total and who was the last bidder.
+
+The rationale behind the book keeping is that anybody who entered into a challenge and put in some stake should be
+rewarded when his position wins, depending on the amount they had to put in. When the challenge ends, if the claimer is
+ahead, he will earn the stakes of every challenger. If a challenger is ahead, each non-last challenger earns a value
+equal to their total stake. The stake of the last challenger being only partially covered by the claimer, he will only
+earn `stake claimer - stake other challengers`, i.e. the remaining tokens.
```
istankovic, line 384: I assume your intention here was to make this monospace, not italic, right?
If so, could you please use \`\`, instead of \`, not just here, but everywhere else?

fredo:
> See issue #346 and #492 for discussion
> 
> Idk if we should keep the idea of parallel claims to solve the "self-challenge" problem in the case the request was honestly filled. We could think about replacing it with the honest filler joining the challenge game as a third party. I'm fine with w/e.

I think we can keep it as an alternative if the request was honestly filled. From an honest filler's perspective I would argue that doing a parallel claim for his own is the most incentivized in the first place. 
But you might be right, that if the honest claim gets also challenged it might be less effort and economically more incentivizing to challenge the false claim as a third party. All this is about to collect funds for L1 resolution and this would be the fastest and easiest way.

docs/source/whitepaper.rst
```diff
@@ -322,10 +322,9 @@ In any case, this condition will always be fulfilled if Bob fills the request be
 Self challenges
 +++++++++++++++
 
-To make the protocol easier to reason about and implement, only two actors can participate in a challenge: the original
-claimer, and the initial challenger. This raises the concern that, after submitting his false claim, Charles could challenge
-himself to prevent anyone from challenging him. This would let Charlie control the state of his challenge and he would be able to
-let it expire with his claim successful as an outcome.
+In the previous explanation, we only talked about two participants to the challenges, Bob and Charles. What if, after
+submitting his false claim, Charles challenges himself. This would let Charles control the state of his challenge and
```
fredo, line 326:
```suggestion
submitting his false claim, Charles challenges himself? This would let Charles control the state of his challenge and
```

docs/source/whitepaper.rst
```diff
@@ -322,10 +322,9 @@ In any case, this condition will always be fulfilled if Bob fills the request be
...
+he would be able to let it expire with his claim successful as an outcome.
 
 To prevent this successful `self-challenged claim` to allow Charles to withdraw Alice's deposit, Bob can fill Alice's request
```
fredo, line 329:
```suggestion
To prevent this successful `self-challenged claim` allowing Charles to withdraw Alice's deposit, Bob can fill Alice's request
```

docs/source/whitepaper.rst
```diff
@@ -335,20 +334,82 @@ Charles can attempt to delay Bob's withdrawal by challenging Bob's rightful clai
...
-Charles' claim reach the end of its period.
-The time constraint described in the previous example also holds in this case.
+Charles' claim reach the end of its period. The time constraint described in the previous example also holds in this case.
 
 Claims that cannot be filled
 ++++++++++++++++++++++++++++
 
 In both the regular `false claim` and `self-challenge claim` cases, we assumed that Bob could fill Alice's request in
 order to prove that the false claimer Charles was not the correct filler. However, If Alice's request cannot be filled
 for any reason (e.g. transfer value too high), instead of proving that someone other than Charles filled a request,
```
fredo, line 344:
```suggestion
order to prove that the false claimer Charles was not the correct filler. However, Alice's request might not be able to be filled for any reason (e.g. transfer value too high). Instead of proving that someone other than Charles filled a request,
```

docs/source/whitepaper.rst
```diff
@@ -335,20 +334,82 @@ Charles can attempt to delay Bob's withdrawal by challenging Bob's rightful clai
...
-Bob will need to prove that no one filled the request before a certain block height. For that, Bob needs to create and
-submit an `L1 non-fill proof` from rollup B to rollup A.
+Bob will need to prove that Charles did not fill the request as claimed. For that, Bob needs to create and submit an
+`L1 non-fill proof` from rollup B to rollup A.
 
-.. todo::
-    Exact specification TBD: https://github.com/beamer-bridge/beamer/issues/346
+When called, the fill manager contract on rollup B recomputes the fill hash from the request hash, and fill ID,
```
fredo, line 348: I would explicitly state, that roll up B will recompute the fillhash from reh request hash and fill ID which was attached to the claim. Basically that the information come from the claim.

docs/source/whitepaper.rst
```diff
@@ -335,20 +334,82 @@ Charles can attempt to delay Bob's withdrawal by challenging Bob's rightful clai
...
+and checks that no fills exists for the corresponding request hash and fill hash. It then submits a proof
+to the outbox of rollup B indicating that the fill hash is invalid, i.e. that the request hash cannot be mapped to the
+fill hash.
+
+Similarly to the filled L1 resolution case, Bob can then trigger a call on L1 to forward this message to rollup A. This
+message will store in the `resolution registry` that the `fill hash` is invalid and make any claim from Charles with
+the corresponding hash an invalid claim.
```
fredo, line 355:
```suggestion
message will store a flag in the `resolution registry` stating that the `fill hash` is invalid. This invalidates any claim with the corresponding ``fill hash``.
```

docs/source/whitepaper.rst
```diff
@@ -335,20 +334,82 @@ Charles can attempt to delay Bob's withdrawal by challenging Bob's rightful clai
...
+
+To make sure the proof arrives in time on rollup A, Bob will need to call the `fill manager` as soon as he notices a
+false claim for a non-filled request. It takes `finalization time of rollup B` after Bob's call to be able to send the
+proof, while the challenge period is defined to be `finalization time of rollup B + challenge period extension`.
```
fredo, line 359:
```suggestion
proof to the resolution registry, while the challenge period is defined to be `finalization time of rollup B + challenge period extension`.
```

docs/source/whitepaper.rst
```diff
@@ -335,20 +334,82 @@ Charles can attempt to delay Bob's withdrawal by challenging Bob's rightful clai
...
+
+In the case where someone challenges Charles on the false claim at the same time as Bob sends the transaction for the
+proof on rollup B, Bob will not be able to challenge Charles and will not receive any financial reward from having sent
+this transaction. The situation being unlikely to happen and the costs of rollup transactions being low, we believe this
+not to be too big of a problem.
+
+The costly L1 transaction can however be sent only when Bob is properly incentivized, that is he is at stake in the
+challenge against Charles.
```
fredo, line 367: What about rewriting it that the L1 transaction might not be properly incentivized, but the L1 transaction can always be executed if time has come.
It sounds a bit like it was not possible to do so which might be confusing to the reader.

docs/source/whitepaper.rst
```diff
@@ -335,20 +334,82 @@ Charles can attempt to delay Bob's withdrawal by challenging Bob's rightful clai
...
+Similarly to the filled L1 resolution case, Bob can then trigger a call on L1 to forward this message to rollup A. This
+message will store in the `resolution registry` that the `fill hash` is invalid and make any claim from Charles with
+the corresponding hash an invalid claim.
+
+To make sure the proof arrives in time on rollup A, Bob will need to call the `fill manager` as soon as he notices a
+false claim for a non-filled request. It takes `finalization time of rollup B` after Bob's call to be able to send the
+proof, while the challenge period is defined to be `finalization time of rollup B + challenge period extension`.
+
+In the case where someone challenges Charles on the false claim at the same time as Bob sends the transaction for the
+proof on rollup B, Bob will not be able to challenge Charles and will not receive any financial reward from having sent
```
fredo, line 362: Maybe it makes sense to rewrite this in the future because I think we should explain the new challenge mechanism when claims are explained. If the reader is aware of how challenge games work at this point, we can rewrite it a bit that he might not be financially rewarded if he is not part of the challenge at any point of the challenge.

docs/source/whitepaper.rst
```diff
@@ -335,20 +334,82 @@ Charles can attempt to delay Bob's withdrawal by challenging Bob's rightful clai
...
+
+The costly L1 transaction can however be sent only when Bob is properly incentivized, that is he is at stake in the
+challenge against Charles.
+
+Non-fill proofs and self challenges
++++++++++++++++++++++++++++++++++++
+
+We explained in a previous part a strategy for honest fillers to counter wrongful self-challenged claims. To be properly
+incentivized for the submission of the non-fill proof on L1, we need to allow Bob to join a self-challenged claim
+initiated by Charles as an additional challenger.
```
fredo, line 374: Also if the challenge mechanism is described already above when the challenge game is explained, we can here just state that the mechanism of being able to join as a third party prevents the described problem here

docs/source/whitepaper.rst
```diff
@@ -335,20 +334,82 @@ Charles can attempt to delay Bob's withdrawal by challenging Bob's rightful clai
...
+Bob bids 5, the stakes look like so:
+
+=========  =======  ======
+ Charles    David    Bob
+=========  =======  ======
+    10        6       5
+=========  =======  ======
+
+If the challenge end at this point, David would earn 6 coins from Charles' stake, and Bob only 4. However, if Charles is
+proven via L1 resolutions to be the correct filler, he will earn 11 coins from the cumulated stakes of David and Bob.
```
fredo, line 412: This would probably makes sense to describe it in challenge game

fredo (changes requested):
Some things to collect in general:

- Describe the new challenge game under section challenge game and rather explain in the problems section which parts of the mechanism are leveraged to prevent the problem. 
- Describe somewhere that the minimum amount of the claimer to counter challenge must be ~L1 cost

docs/source/whitepaper.rst
cducrest: Makes sense

docs/source/whitepaper.rst
cducrest: What you want to prevent is the allowance of Charles to withdraw, not the claim. So `to allow` is correct.

fredo: I did not know that it changes meaning if you change from to+infinitive to a gerund. But I'm certainly not a grammar expert.
The sentence still sounds weird to me

cducrest:
> Describe somewhere that the minimum amount of the claimer to counter challenge must be ~L1 cost

I think we should extend the part about the protocol parameters later to make explicit our timing assumptions and values of the stake, etc ... I would add that to this part in another PR then.

fredo (approved):
looks good

PR #556, opened 2022-04-12
Set up a Prometheus metrics endpoint

### LINKED ISSUE #468: Set up a Prometheus endpoint
We should do the following:
- set up a Prometheus endpoint within the agent so that the agent can be polled
- add some minimal stats to the data returned by the endpoint:
  - the agent's source and target rollups
  - the agent's uptime
  - the total number of requests filled on the target rollup (regardless of the filler)
  - the number of requests the agent filled on the target rollup during agent's uptime
  - the total number of requests created on the source rollup
- configure the alert mechanism on `metrics.brainbot.com` so that if an agent goes down, the team is notified by e-mail

istankovic:
This uses the `prometheus_client` Python package to set up an endpoint that serves some simple data about the running agent,
as outlined in #468.

Part of #468.

I've tested the endpoint locally and it seems to work alright. The next step would be to configure our Prometheus server to poll our agents' endpoints periodically. Once that is done, we can close #468.

### DIFF
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -1,6 +1,7 @@
 import threading
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Optional
 
 import structlog
 import web3
@@ -9,6 +10,7 @@ from eth_typing import Address
 from web3.gas_strategies.rpc import rpc_gas_price_strategy
 from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware
 
+import beamer.metrics
 from beamer.chain import ContractEventMonitor, EventProcessor
 from beamer.contracts import DeploymentInfo, make_contracts
 from beamer.state_machine import Context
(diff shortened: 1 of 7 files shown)


fredo (commented):
I don't get exactly why the whole `_DATA` object is needed or why this was implemented in this way, but it seems to look good to me.

Could you elaborate on this?

istankovic:
> I don't get exactly why the whole `_DATA` object is needed or why this was implemented in this way, but it seems to look good to me.

The way Prometheus client library works is that you create a number of counters, gauges etc. and then update them as needed. Our metrics configuration is anyway static so it makes sense to do it just once, at agent start, and then update. That's the reason for the `_DATA` singleton. The associated lock guards against simultaneous updates from multiple threads.

I presume one could also create the metrics object at update time (I haven't checked it, though), however, that will only increase the cost of updating metrics, which I tried to minimise. The idea is to be minimally intrusive and not affect performance too much.

fredo:
Okay so the lock is there for the future in case we update from multiple threads? Because from what I can see now we only update in one thread

istankovic:
> Okay so the lock is there for the future in case we update from multiple threads? Because from what I can see now we only update in one thread

Exactly -- it's about making it robust and separation of concerns. Users of that API don't need to care about locks.

fredo (approved):

PR #559, opened 2022-04-13
ci: Make agent tests conditional in pure frontend PRs

palango:
This will avoid running agents tests when jst frontend is touched.

I decided to run in on merges to `main` unconditionally, so we have data there.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -3,8 +3,11 @@ name: Agent CI
 on:
   push:
     branches: [ main ]
+
   pull_request:
     branches: [ main ]
+    paths-ignore:
+      -  'frontend/**'
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:


.github/workflows/backend.yml
```diff
@@ -3,8 +3,11 @@ name: Agent CI
 on:
   push:
     branches: [ main ]
+
   pull_request:
     branches: [ main ]
+    paths-ignore:
+      -  'frontend/**'
```
istankovic, line 10: Should we also do this for pushes to `main`?

istankovic (approved):
Thanks for doing this!

.github/workflows/backend.yml
palango: As I wrote, I decided against it. But if you convince me I can add it.

istankovic: > I decided to run in on merges to main unconditionally, so we have data there.

Yeah, I did not get this completely, not sure what data you were refering to.
But in any case, it's a minor point, feel free to ignore and merge.

PR #560, opened 2022-04-13
Fix linting issue

nekhaly:


weilbith (approved):
CI is green again. So it does the job. Sad that this is necessary but that's how it is.
Thanks for taking care of this!

PR #488, opened 2022-03-31
Docs fixes

istankovic:
This updates the instructions on how to run and test agents locally.
It also extends the deployment script to allow deployments to a local ganache instance.

Part of #411.

PR #574, opened 2022-04-14
Disallow self as target chain

### LINKED ISSUE #496: Implement check in contracts for same source and targets chain ids
The source and target chain ids should not be the same.

palango:
Resolves #496 

We don't need an explicit check, because we have a check here now: https://github.com/beamer-bridge/beamer/blob/082a23ad9a956f61e0861e132f6b26603a8f42ba/contracts/contracts/RequestManager.sol#L158

I also appended a little commit that re-orders the requires, so that we burn less gas if a require should be hit.

### DIFF
--- a/scripts/deployment/main.py
+++ b/scripts/deployment/main.py
@@ -14,6 +14,7 @@ from web3.contract import Contract
 from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware
 
 OPTIMISM_L2_MESSENGER_ADDRESS = to_checksum_address("0x4200000000000000000000000000000000000007")
+GANACHE_CHAIN_ID = 1337
 
 
 class DeployedContract(Contract):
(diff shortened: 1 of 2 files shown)


scripts/deployment/main.py
```diff
@@ -214,7 +215,11 @@ def main(
         chain_id = web3_l2.eth.chain_id
         assert chain_id == l2_config["chain_id"]
 
-        supported_target_chains = {l2["chain_id"]: l2["finalization_time"] for l2 in config["L2"]}
+        supported_target_chains = {
+            l2["chain_id"]: l2["finalization_time"]
+            for l2 in config["L2"]
+            if l2["chain_id"] != chain_id or chain_id == GANACHE_CHAIN_ID
+        }
```
istankovic, line 222: How has it been working so far, without explicitly mentioning ganache?

istankovic: I've been using
```
--config-file scripts/deployment/ganache-local.json --test-messenger
```
as options to the deployment script for some time now and it worked for me.

palango: You removed the if clause in your PR: https://github.com/beamer-bridge/beamer/pull/488/files#diff-c6d950f3bb92a40d8b76462f0c0f34448141fee95c49373f1e35d96960a48828R217

So I bring it back and make sure that it still works on local chains.

istankovic: Ah, I see, you needed to bring it in due to it being the actual topic of the PR -- and then you need to allow for ganache again. Understood. :+1:

istankovic (approved):

PR #579, opened 2022-04-19
Cleanup contract scripts

### LINKED ISSUE #443: clean up `contracts/scripts/` directory
some files might be outdated

istankovic:
Resolves #443.


.github/workflows/backend.yml
```diff
@@ -94,16 +94,7 @@ jobs:
         run: poetry run brownie test -v -G beamer/tests/contracts
 
       - name: Run agent tests
-        run: |
-          # Generate deployment artifacts under contracts/build/deployments/dev.
-          # Start ganache separately so that brownie does not need to start
-          # it. If brownie starts ganache, it will remove the generated
-          # deployment artifacts after it is done.
-          ganache-cli --accounts 10 --hardfork istanbul --gasLimit 12000000 --mnemonic brownie --port 8545 -q &
```
palango, line 102: Why isn't this needed any mroe?

docker/optimism-scripts/utils.py
```diff
@@ -22,20 +21,24 @@
 OPTIMISM_L2_MESSENGER_ADDRESS = "0x4200000000000000000000000000000000000007"
 
 
+def _address_file_path() -> str:
+    return os.environ["ADDRESS_FILE_PATH"]
```
palango, line 25: Can you add a default here?

palango (approved):

docker/optimism-scripts/utils.py
istankovic: I would actually prefer the script to fail here if the env variable is not set, so that we know about it and can fix whatever is invoking the script.

PR #593, opened 2022-04-20
Update withdraw claim strategy to only withdraw own winning claims

cducrest:
I believe that without this, the agent will attempt to withdraw any invalid claim that were contested by someone else:

1- Bob makes wrong claim
2- Charles challenges
3- Claim expires and Charles wins
4- Agent (that did not participate so far) tries to withdraw

Also, without the `continue` in the withdraw `if` clause, the agent will try to challenge a terminated claim.

The withdraw does not fail, and instead sends the funds to Charles because that's how it is in the contracts, but I don't believe the Agent should do that.

I attempted to write a test and could actually verify the behaviour with logs, but since the withdraw tx does not fail, it's hard to write an effective test.

Idk if we should transition the claim to `ignored` state, I don't fully understand this.

### DIFF
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -245,6 +245,9 @@ def process_claims(context: Context) -> None:
     for claim in context.claims:
         log.debug("Processing claim", claim=claim)
 
+        if claim.is_ignored:
+            continue
+
         if claim.is_withdrawn:
             log.debug("Removing withdrawn claim", claim=claim)
             to_remove.append(claim.id)
@@ -265,11 +268,8 @@ def process_claims(context: Context) -> None:
         if claim.transaction_pending:
             continue
 
-        block = context.latest_blocks[request.source_chain_id]
-        if block["timestamp"] >= claim.termination:
-            withdraw(claim, context)
-
         if claim.is_claimer_winning or claim.is_challenger_winning:
+            maybe_withdraw(claim, context)
             maybe_challenge(claim, context)
 
     for claim_id in to_remove:
@@ -359,6 +359,10 @@ def maybe_challenge(claim: Claim, context: Context) -> bool:
     # 1) the claim is dishonest AND nobody challenged it yet
     #
     # 2) we participate in the game AND it is our turn
+    request = context.requests.get(claim.request_id)
+    block = context.latest_blocks[request.source_chain_id]
+    if block["timestamp"] >= claim.termination:
+        return False
     if int(time.time()) < claim.challenge_back_off_timestamp:
         return False
 
@@ -387,7 +391,17 @@ def maybe_challenge(claim: Claim, context: Context) -> bool:
     return True
 
 
-def withdraw(claim: Claim, context: Context) -> None:
+def maybe_withdraw(claim: Claim, context: Context) -> None:
+    request = context.requests.get(claim.request_id)
+    block = context.latest_blocks[request.source_chain_id]
+    if block["timestamp"] < claim.termination:
+        # TODO: once L1 resolution is implemented, maybe withdraw before claim is terminated
+        return
+
+    agent_winning = claim.get_winning_address() == context.address
+    if not agent_winning:
+        return
+
     func = context.request_manager.functions.withdraw(claim.id)
     try:
         txn_hash = _transact(func)
--- a/beamer/tests/agent/test_challenge.py
+++ b/beamer/tests/agent/test_challenge.py
@@ -311,3 +311,43 @@ def test_challenge_5(request_manager, fill_manager, token, config, honest_claim)
     agent.stop()
     proxy_l2b.stop()
     agent.wait()
+
+
+# Scenario 6:
+#
+# Charlie          Dave
+# --------------------------
+# claim
+#                  challenge
+#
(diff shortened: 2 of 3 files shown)


beamer/chain.py
```diff
@@ -393,3 +399,4 @@ def withdraw(claim: Claim, context: Context) -> None:
 
     claim.transaction_pending = True
     log.debug("Withdrew", claim=claim.id, txn_hash=txn_hash.hex())
+    return
```
istankovic, line 402: Why this `return`?

fredo:
When a claim is challenged and the actor is **not** involved the claim is transitioned into the `ignored` state. See https://github.com/beamer-bridge/beamer/blob/6f22a95e398687ec2703ea4c27905df3136c0bc8/beamer/state_machine.py#L158.

We should rather check whether the claim is in the correct state to actually call withdraw which is either `ClaimerWinning` or `ChallengerWinning` dependend on whether the agent is claimer or challenger. I'd rather like to see something more similar to `process_requests()` where we also ask for states, i.e.

```py
if request.is_filled:
    do_something()
```

palango:
> I attempted to write a test and could actually verify the behaviour with logs, but since the withdraw tx does not fail, it's hard to write an effective test.

How did you try to write those tests? Something like
https://github.com/beamer-bridge/beamer/blob/88b46bdd42360a8e7ab997fcd166fa622c56d0df/beamer/tests/agent/test_request.py#L149-L165

If so, I hope testing this would be easier with an approach as in
https://github.com/beamer-bridge/beamer/blob/88b46bdd42360a8e7ab997fcd166fa622c56d0df/beamer/tests/agent/test_handlers.py#L120-L138

cducrest:
> > I attempted to write a test and could actually verify the behaviour with logs, but since the withdraw tx does not fail, it's hard to write an effective test.
> 
> How did you try to write those tests? Something like
> 
> https://github.com/beamer-bridge/beamer/blob/88b46bdd42360a8e7ab997fcd166fa622c56d0df/beamer/tests/agent/test_request.py#L149-L165
> 
> If so, I hope testing this would be easier with an approach as in
> 
> https://github.com/beamer-bridge/beamer/blob/88b46bdd42360a8e7ab997fcd166fa622c56d0df/beamer/tests/agent/test_handlers.py#L120-L138

I did something like 

```
# Scenario 4:
#
# Charlie          Dave
# --------------------------
# claim
#                  challenge
#
# Winner: Charlie
#
# Note: Bob is not participating in the challenge here. We test whether Bob
# will attempt to withdraw the stakes in place of Dave.
def test_withdraw_not_participant(request_manager, fill_manager, token, config):
    target_address = accounts[8]
    requester, charlie, dave = accounts[:3]

    agent = beamer.agent.Agent(config)

    # Submit a request that Bob cannot fill.
    amount = token.balanceOf(agent.address) + 1
    request_id = make_request(request_manager, token, requester, target_address, amount)

    stake = request_manager.claimStake()
    request_manager.claimRequest(request_id, 0, {"from": charlie, "value": stake})

    collector = EventCollector(request_manager, "ClaimMade")
    claim = collector.next_event()
    request_manager.challengeClaim(claim.id, {"from": dave, "value": stake+1})
    brownie.chain.mine(timestamp=claim.termination)

    agent.start()

    time.sleep(2)
    agent.stop()
    agent.wait()
```
But you gave me better ideas that I do not yet fully understand.

beamer/chain.py
```diff
@@ -359,6 +359,10 @@ def maybe_challenge(claim: Claim, context: Context) -> bool:
     # 1) the claim is dishonest AND nobody challenged it yet
     #
     # 2) we participate in the game AND it is our turn
+    request = context.requests.get(claim.request_id)
+    block = context.latest_blocks[request.source_chain_id]
+    if block["timestamp"] >= claim.termination:
+        return False
```
palango, line 365: The return value doesn't seem to be used here: https://github.com/beamer-bridge/beamer/pull/593/files#diff-59108d402087b9f4bfab36ec32d92fd39d9c477b17ac359eeaaece1104e0f699R273

Is that expected?

beamer/chain.py
```diff
@@ -387,7 +391,17 @@ def maybe_challenge(claim: Claim, context: Context) -> bool:
     return True
 
 
-def withdraw(claim: Claim, context: Context) -> None:
+def maybe_withdraw(claim: Claim, context: Context) -> None:
```
palango, line 394: nit: Can you move the function above `maybe_challenge` as it's used before.

beamer/tests/agent/test_challenge.py
```diff
@@ -311,3 +311,43 @@ def test_challenge_5(request_manager, fill_manager, token, config, honest_claim)
...
+# Charlie          Dave
+# --------------------------
+# claim
+#                  challenge
+#
+# Winner: Charlie
+#
+# Note: Bob is not participating in the challenge here. We test whether Bob
+# will attempt to withdraw the stakes in place of Dave.
+def test_withdraw_not_participant(request_manager, token, config):
```
palango, line 327: I'd rather write test this just using the state machine. Did you try this?

Issue #271, opened 2022-03-01
Allow the node to submit L1 transaction


istankovic:
After there are enough funds to go to L1, the node should submit an L1 transaction to resolve the challenge game.

See #185.

When to go to L1:
As soon as there are enough funds.

Mechanics for going to L1:
- [ ] Should trigger when the other party raised us
  - [ ] Should we still counter-challenge to increase the stakes?
- [ ] Have a new state, so not to send multiple transactions.
- [ ] Check if transaction arrived.
- [x] Filter for event from resolution registry.
- [x] Store fill tx hash

palango:
The interface to #185 will be an executable that takes the following parameters:
-    L2 RPC
 -   L1 RPC
  -  L1 wallet
   - Hash of the L2 transaction that includes the message(s) to forward


We still need to decide what should be returned

PR #596, opened 2022-04-21
agent: Query events from ResolveRegistry

palango:
Part of #271

### DIFF
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -11,7 +11,7 @@ from web3.gas_strategies.rpc import rpc_gas_price_strategy
 from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware
 
 import beamer.metrics
-from beamer.chain import ContractEventMonitor, EventProcessor
+from beamer.chain import EventMonitor, EventProcessor
 from beamer.contracts import DeploymentInfo, make_contracts
 from beamer.state_machine import Context
 from beamer.tracker import Tracker
@@ -79,45 +79,38 @@ class Agent:
         )
         self._event_processor = EventProcessor(self.context)
 
-        self._contract_monitor_l2a = ContractEventMonitor(
-            "RequestManager",
-            request_manager,
-            l2a_contracts_info["RequestManager"].deployment_block,
-            self._event_processor.add_events,
-            self._event_processor.mark_sync_done,
+        self._event_monitor_l2a = EventMonitor(
+            web3=w3_l2a,
+            contracts=(request_manager, resolution_registry),
+            deployment_block=min(
+                l2a_contracts_info["RequestManager"].deployment_block,
+                l2a_contracts_info["ResolutionRegistry"].deployment_block,
+            ),
+            on_new_events=self._event_processor.add_events,
+            on_sync_done=self._event_processor.mark_sync_done,
         )
 
-        self._contract_monitor_l2a_registry = ContractEventMonitor(
-            "ResolutionRegistry",
-            resolution_registry,
-            l2a_contracts_info["ResolutionRegistry"].deployment_block,
-            self._event_processor.add_events,
-            self._event_processor.mark_sync_done,
-        )
-
-        self._contract_monitor_l2b = ContractEventMonitor(
-            "FillManager",
-            fill_manager,
-            l2b_contracts_info["FillManager"].deployment_block,
-            self._event_processor.add_events,
-            self._event_processor.mark_sync_done,
+        self._event_monitor_l2b = EventMonitor(
+            web3=w3_l2b,
+            contracts=(fill_manager,),
+            deployment_block=l2b_contracts_info["FillManager"].deployment_block,
+            on_new_events=self._event_processor.add_events,
+            on_sync_done=self._event_processor.mark_sync_done,
         )
 
     def start(self) -> None:
         assert self._stopped.is_set()
         beamer.metrics.init(self._config)
         self._event_processor.start()
-        self._contract_monitor_l2a.start()
-        self._contract_monitor_l2a_registry.start()
-        self._contract_monitor_l2b.start()
+        self._event_monitor_l2a.start()
+        self._event_monitor_l2b.start()
         self._stopped.clear()
 
     def stop(self) -> None:
         assert not self._stopped.is_set()
         self._event_processor.stop()
-        self._contract_monitor_l2a.stop()
-        self._contract_monitor_l2a_registry.stop()
-        self._contract_monitor_l2b.stop()
+        self._event_monitor_l2a.stop()
+        self._event_monitor_l2b.stop()
         self._stopped.set()
 
     @property
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -58,6 +58,7 @@ class Agent:
         l2b_contracts = make_contracts(w3_l2b, l2b_contracts_info)
 
         request_manager = l2a_contracts["RequestManager"]
+        resolution_registry = l2a_contracts["ResolutionRegistry"]
         fill_manager = l2b_contracts["FillManager"]
 
         if not fill_manager.functions.allowedLPs(config.account.address).call():
@@ -86,6 +87,14 @@ class Agent:
             self._event_processor.mark_sync_done,
         )
 
+        self._contract_monitor_l2a_registry = ContractEventMonitor(
+            "ResolutionRegistry",
+            resolution_registry,
+            l2a_contracts_info["ResolutionRegistry"].deployment_block,
+            self._event_processor.add_events,
+            self._event_processor.mark_sync_done,
+        )
+
         self._contract_monitor_l2b = ContractEventMonitor(
             "FillManager",
             fill_manager,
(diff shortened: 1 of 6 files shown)


beamer/agent.py
```diff
@@ -99,13 +108,15 @@ def start(self) -> None:
         beamer.metrics.init(self._config)
         self._event_processor.start()
         self._contract_monitor_l2a.start()
+        self._contract_monitor_l2a_registry.start()
```
fredo, line 111: Maybe a renaming to `_request_manager_monitor...` , `_fill_manager_monitor` would make sense now for a better readability of the code

beamer/chain.py
```diff
@@ -115,11 +115,11 @@ def __init__(self, context: Context):
     @property
     def _synced(self) -> bool:
         with self._lock:
-            return self._num_syncs_done == 2
+            return self._num_syncs_done == 3
 
     def mark_sync_done(self) -> None:
         with self._lock:
-            assert self._num_syncs_done < 2
+            assert self._num_syncs_done < 3
```
fredo, line 122: I don't know how but is it possible to not have this hard coded value here? Subject to errors in the future if we add contract monitors. 

Maybe one solution could be capturing all contract monitors in a dict. 

```py
self._contract_monitors = {
    "request_manager": ContractMonitor(...),
    "fill_manager": ....
    ....
    }
```

and then at the respective comparisons, you could use `len(self._contract_monitors)`. Just an idea...

palango: ANother alternative is that we make each event monitor able to query multiple contracts. That would also help to keep the number of RPC requests down

istankovic: > ANother alternative is that we make each event monitor able to query multiple contracts. That would also help to keep the number of RPC requests down

Only if all the contracts are on the same chain, right?

palango: Of course, we need at least one for the source and one for the target chain

istankovic: Hm, that sounds like a nice optimisation. I like it. Even more so if one considers the plan to support multiple chain pairs with a single agent. In that case, tracking all event monitors would become a bit messy, but with your approach we would only ever have 1 event monitor per supported chain, which makes scaling as good as it can possibly get. That's a definite like. :+1:

beamer/state_machine.py
```diff
@@ -63,6 +65,9 @@ def process_event(event: Event, context: Context) -> bool:
     elif isinstance(event, ClaimWithdrawn):
         return _handle_claim_withdrawn(event, context)
 
+    elif isinstance(event, (RequestResolved, FillHashInvalidated)):
+        return False
+
```
istankovic, line 70: Do we have documentation for these events/transitions?

palango: What do you mean? This PR just adds the events and queries them. No state machine code is touched, but we need to ignore them here because we run into a panic otherwise.

istankovic: > What do you mean? This PR just adds the events and queries them. No state machine code is touched, but we need to ignore them here because we run into a panic otherwise.

Right, but that just means that the state machine remains in the same state upon receiving those events.
What I'm trying to say is, it would be nice to have that fact reflected in our state machine diagram.

A related question is: when can those events be emitted i.e. for which states could it happen that those events arrive?

palango: > Right, but that just means that the state machine remains in the same state upon receiving those events.
> What I'm trying to say is, it would be nice to have that fact reflected in our state machine diagram.

I would put it differently, we ignore them just as before. And as this is part of a larger effort to get L1 resolution working I don't see any point in adjusting the docs in this PR.

istankovic (approved):
Looks fine to me.
Possible improvements to event monitors can/should be done at a later point.

beamer/agent.py
palango: Replaced this by multi-contract monitors now.

beamer/chain.py
palango: Done.

beamer/chain.py
```diff
@@ -48,26 +51,26 @@ def wrapper(*args, **kwargs):  # type: ignore
...
 
 
-class ContractEventMonitor:
+class EventMonitor:
     def __init__(
         self,
-        name: str,
-        contract: web3.contract.Contract,
+        web3: Web3,
+        contracts: list[Contract],
```
istankovic, line 58: I'd prefer if we used `tuple[Contract]` here -- mutability is not really required and tuples would be more robust against accidental changes to the list outside of the event monitor.

beamer/chain.py
```diff
@@ -48,26 +51,26 @@ def wrapper(*args, **kwargs):  # type: ignore
...
         deployment_block: BlockNumber,
         on_new_events: Callable[[list[Event]], None],
         on_sync_done: Callable[[], None],
     ):
-        self._name = name
-        self._contract = contract
+        self._web3 = web3
+        self._contracts = contracts
```
istankovic, line 64: It would be nice to `assert` that all contracts share the same chain ID.

beamer/chain.py
```diff
@@ -48,26 +51,26 @@ def wrapper(*args, **kwargs):  # type: ignore
...
         self._deployment_block = deployment_block
         self._stop = False
         self._on_new_events = on_new_events
         self._on_sync_done = on_sync_done
-        self._log = structlog.get_logger(type(self).__name__).bind(contract=name)
+        self._log = structlog.get_logger(type(self).__name__)
```
istankovic, line 69: Could you please bind a `chain_id` here? It should make the log files clearer.

beamer/chain.py
```diff
@@ -48,26 +51,26 @@ def wrapper(*args, **kwargs):  # type: ignore
...
 
     def start(self) -> None:
         self._thread = threading.Thread(
-            name=self._name, target=_wrap_thread_func(self._thread_func)
+            name="EventMonitor", target=_wrap_thread_func(self._thread_func)
```
istankovic, line 73: I'd also include the chain ID in the name.

beamer/chain.py
palango: Ho9w do I tell python that a tuple of arbitrary length is fine?

beamer/chain.py
```diff
@@ -92,7 +97,7 @@ def _thread_func(self) -> None:
                 self._on_new_events(events)
             # TODO: wait for new block instead of the sleep here
             time.sleep(1)
-        self._log.info("ContractEventMonitor stopped", chain_id=chain_id)
+        self._log.info("EventMonitor stopped", chain_id=chain_id)
```
istankovic, line 100: With bound `chain_id`, we wouldn't need to provide it here or above.

beamer/events.py
```diff
@@ -108,7 +108,15 @@ def _camel_to_snake(s: str) -> str:
 )
 
 
-def _make_topics_to_abi(contract: web3.contract.Contract) -> dict[bytes, ABIEvent]:
+def _make_topics_abi_mapping_for_contracts(contracts: list[Contract]) -> dict[bytes, ABIEvent]:
```
istankovic, line 111: I'd also use tuples here since we don't need mutability.
If one is in "proper Python typing" mode, I'd say using `Iterable[Contract]` would be most appropriate.

beamer/events.py
```diff
@@ -147,13 +155,13 @@ class EventFetcher:
     _ETH_GET_LOGS_THRESHOLD_FAST = 2
     _ETH_GET_LOGS_THRESHOLD_SLOW = 5
 
-    def __init__(self, contract_name: str, contract: Contract, start_block: BlockNumber):
-        self._contract_name = contract_name
-        self._contract = contract
+    def __init__(self, web3: Web3, contracts: list[Contract], start_block: BlockNumber):
```
istankovic, line 158: Similar thing here: please consider using `tuple[Contract]` or `Iterable[Contract]`.

beamer/events.py
```diff
@@ -147,13 +155,13 @@ class EventFetcher:
...
+        self._web3 = web3
+        self._contract_addresses = [c.address for c in contracts]
```
istankovic, line 160: Would be nice to `assert` that `web3`'s chain ID matches those of contracts.

beamer/agent.py
```diff
@@ -79,44 +79,37 @@ def __init__(self, config: Config):
         )
         self._event_processor = EventProcessor(self.context)
 
-        self._contract_monitor_l2a = ContractEventMonitor(
-            "RequestManager",
-            request_manager,
-            l2a_contracts_info["RequestManager"].deployment_block,
-            self._event_processor.add_events,
-            self._event_processor.mark_sync_done,
+        self._contract_monitor_l2a = EventMonitor(
```
istankovic, line 82: Nitpick: consider renaming the instance variable to `_event_monitor_l2a` to be consistent with the new type name.

beamer/agent.py
```diff
@@ -79,44 +79,37 @@ def __init__(self, config: Config):
...
-            self._event_processor.mark_sync_done,
-        )
-
-        self._contract_monitor_l2b = ContractEventMonitor(
-            "FillManager",
-            fill_manager,
-            l2b_contracts_info["FillManager"].deployment_block,
-            self._event_processor.add_events,
-            self._event_processor.mark_sync_done,
+        self._contract_monitor_l2b = EventMonitor(
```
istankovic, line 93: Nitpick: consider renaming the instance variable to `_event_monitor_l2b` to be consistent with the new type name.

Issue #569, opened 2022-04-13
Non-Fill: Multiparty bidding in challenges


fredo:
In order to prevent self-challenges and thus the attacker controlling the state of the claim, we need to allow other participants to join an existing challenge. 
The challenge mechanism stays mostly the same, with the only difference that **IF** the claimer is winning anyone (not only the existing challenger) can outbid the claimer and thus being a challenger himself.

The important key factor here is that we now have multiple challengers for one claim. This being said, we need to store all stakes mapped to the respective challengers. This information is crucial for the later payout in case the challenger wins.

In case of  the challengers winning  the payout scheme is as followed: 

- Each challenger who is not the last bidder will get double the amount of their stake
- The last bidder will get the remaining stake

This can be more than the stake of the last bidder (in case the claimer was winning and it got resolver by L1) or less (in case the challengers are leading and the excess amount is not covered by the claimer's stake.

Note, that this payout scheme holds true in the current implementation with number of challengers = 1.

### Implementation requirements
- Upon resolving the claim, each challenger will have to withdraw its stake. We must not iterate over a list of challengers, paying out all of them in a single transaction as this would limit the number of challengers. In theory we must allow an infinite number of challengers.
- We should store all necessary information so that the last bidder does not have to wait until all other challengers have withdrawn their stake. IOW, the order of withdraw must not matter.
- Take care that this will most likely also break the state machine of the agent. Even if the agent decides not to engage in multiparty challenges, there will be most likely (depending on the implementation) multiple `ClaimWithdrawn` events for one Claim. The agent must be able to consume multiple ClaimWithdrawn events for one claim. A potential error I can think of is that if the agent is one of the challengers, we must make sure that definetly calls withdraw on the claim and is able to differentiate other withdraw events from other challengers.
- The underlying Claim class must be extende that it can hold a set of challengers. I assume that the ClaimMade event will submit the corresponding challenger address and the agent must collect these data and collect all challengers

The logic that the agent will engage in ongoing challenges I will outsource to another issue. This should be independent from the contract's implementation of the mechanism.

PR #600, opened 2022-04-25
contracts: allow muliparty challenges

fredo:
fixes: #569 

Currently the agent will try to challenge any claim where the claimer is leading and not honest. I haven't found a useful if condition to check the non-fill condition we discussed. 

### Still Todo
- Check docs for update 

Note, that it is not possible anymore, that any address can call withdraw in the case that the challengers win. This is due to that the challengers will withdraw their own partial stake.
We should think about exposing a functionality in the contract where anybody can call withdraw on behalf of an address. But I leave that for another PR.


istankovic:
Gas profile difference:
```
                                    /tmp/before │ /tmp/after   │ difference │
FillManager
   ├─ constructor            - avg:     813,250      813,247           -3  
   ├─ fillRequest            - avg:      93,686       93,683           -3  
   ├─ invalidateFillHash     - avg:      48,684       48,684            0  
   ├─ addAllowedLP           - avg:      44,221       44,236          +15  
   └─ removeAllowedLP        - avg:      20,119       20,119            0  
MintableToken
   ├─ constructor            - avg:     694,990      694,990            0  
   ├─ mint                   - avg:      50,688       50,685           -3  
   └─ approve                - avg:      43,661       43,688          +27  
OptimismProofSubmitter
   ├─ constructor            - avg:     550,890      550,895           +5  
   ├─ addCaller              - avg:      44,798       44,795           -3  
   └─ submitProof            - avg:      24,020       24,020            0  
RequestManager
   ├─ constructor            - avg:   1,988,923    2,032,740      +43,817  
   ├─ createRequest          - avg:     212,149      212,232          +83  
   ├─ claimRequest           - avg:     137,768      139,570       +1,802  
   ├─ withdraw               - avg:      69,493       63,759       -5,734  
   ├─ withdrawExpiredRequest - avg:      57,663       57,663            0  
   ├─ challengeClaim         - avg:      57,405       61,512       +4,107  
   ├─ setFinalizationTime    - avg:      42,602       42,602            0  
   ├─ withdrawbeamerFees     - avg:      22,403       22,403            0  
   └─ updateFeeData          - avg:      17,063       16,992          -71  
ResolutionRegistry
   ├─ constructor            - avg:     546,165      546,165            0  
   ├─ addCaller              - avg:      65,111       65,114           +3  
   ├─ resolveRequest         - avg:      37,053       37,059           +6  
   └─ invalidateFillHash     - avg:      22,839       22,839            0  
Resolver
   ├─ constructor            - avg:     698,931      698,931            0  
   ├─ addCaller              - avg:      65,134       65,136           +2  
   ├─ addRegistry            - avg:      65,124       65,128           +4  
   ├─ resolve                - avg:      23,468       23,468            0  
   └─ resolveNonFill         - avg:      22,964       22,964            0  
TestCrossDomainMessenger
   ├─ constructor            - avg:     237,338      237,338            0  
   ├─ setLastSender          - avg:      42,635       42,635            0  
   └─ setForwardState        - avg:      25,705       25,569         -136  
```

beamer/models/claim.py
```diff
@@ -17,6 +18,7 @@ def __init__(
     ) -> None:
         super().__init__()
         self._latest_claim_made = claim_made
+        self.challengers: defaultdict[Address, int] = defaultdict(int)
```
palango, line 21: Do we need the `defaultdict` here?

contracts/contracts/RequestManager.sol
```diff
@@ -306,50 +312,84 @@ contract RequestManager is Ownable {
...
+            else {
+                ethToTransfer -= (challengerStakeTotal - claimerStake);
+            }
+        }
+
+        // First time withdraw is called, remove it from active claim
+        if (claim.withdrawnAmount == 0) {
+            request.activeClaims -= 1;
+        }
+        claim.withdrawnAmount += ethToTransfer;
```
istankovic, line 377: I would suggest something like
```
require(claim.withdrawnAmount <= claimerStake + challengerStakeTotal, "Amount to withdraw too large")
```
after this line, just to be safe.

contracts/contracts/RequestManager.sol
```diff
@@ -58,7 +59,7 @@ contract RequestManager is Ownable {
     );
 
     event ClaimMade(
-        uint256 requestId,
+        uint256 indexed requestId,
```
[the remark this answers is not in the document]
fredo: This is a general optimization, the frontend team asked for it

contracts/contracts/RequestManager.sol
```diff
@@ -70,7 +71,7 @@ contract RequestManager is Ownable {
 
     event ClaimWithdrawn(
         uint256 claimId,
-        uint256 requestId,
+        uint256 indexed requestId,
```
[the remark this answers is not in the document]
fredo: The frontend needs to query for claim events on requests

beamer/models/claim.py
fredo: since we add new stakes on top of challengers once a new `ClaimMade` event arrives, I used this to safe me the 

```
old_stake = 0
if challenger in self.challengers:
    old_stake = self.challengers[challenger]
self.challengers[challenger] = old_stake + extra_stake
```

contracts/contracts/RequestManager.sol
```diff
@@ -252,20 +253,21 @@ contract RequestManager is Ownable {
...
                 periodExtension = finalizationTimes[request.targetChainId] + challengePeriodExtension;
             } else {
                 periodExtension = challengePeriodExtension;
             }
-            nextActor = claim.challenger;
-            minValue = claim.claimerStake - claim.challengerStake + 1;
+            require(claim.claimer != msg.sender, "Cannot challenge own claim");
+            nextActor = msg.sender;
+            minValue = claimerStake - challengerStakeTotal + 1;
         } else {
```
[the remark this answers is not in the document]
fredo: totally! nice catch. Will fix this

beamer/tests/contracts/test_request_manager.py
```diff
@@ -676,6 +720,41 @@ def test_withdraw_without_challenge_with_resolution(
...
+        request_manager.withdraw(claim_id, {"from": claimer})
+
+    # First challenger withdraws his partial stake
+    request_manager.withdraw(claim_id, {"from": first_challenger})
+
+    # All stakes are withdrawn already
+    with brownie.reverts("Claim already withdrawn"):
+        request_manager.withdraw(claim_id, {"from": claimer})
+
+
```
istankovic, line 757: Do we have a test for 
```
// Invalid claim but no one challenged, stake goes to protocol
// Invalid claim but no one challenged, stake goes to protocol 
ethToTransfer = claimerStake;                                  
claimReceiver = address(this);                                 
```
? If not, we should.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -676,6 +720,41 @@ def test_withdraw_without_challenge_with_resolution(
...
 def test_withdraw_expired(token, request_manager):
     """Test that a request can be withdrawn once it is expired"""
     validity_period = 60 * 5
```
istankovic, line 760: We should also have a test with 2 or 3 challengers, with different stake amounts, to make sure that the amounts of ETH they withdraw at the end is correct. I haven't seen this covered.

contracts/contracts/RequestManager.sol
```diff
@@ -34,9 +34,10 @@ contract RequestManager is Ownable {
         uint256 requestId;
         address claimer;
         uint256 claimerStake;
-        address challenger;
-        uint256 challengerStake;
-        bool withdrawn;
+        mapping (address => uint256) challengers;
```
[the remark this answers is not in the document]
istankovic: Maybe `challengeStakes`?

beamer/models/claim.py
```diff
@@ -71,10 +73,10 @@ def valid_claim_for_request(self, request: Request) -> bool:
             return False
         return True
 
-    def get_winning_address(self) -> Address:
+    def get_winning_addresses(self) -> frozenset[Address]:
```
istankovic, line 76: Nice to see `frozenset` being used! ;-)

beamer/models/claim.py
```diff
@@ -17,6 +18,7 @@ def __init__(
     ) -> None:
         super().__init__()
         self._latest_claim_made = claim_made
+        self.challengers = defaultdict(int)
```
istankovic: We could make this private, no?

beamer/models/claim.py
```diff
@@ -99,6 +101,13 @@ def on_ignore(self, new_claim_made: Optional[ClaimMade] = None) -> None:
             self._on_new_claim_made(self._latest_claim_made)
 
     def _on_new_claim_made(self, new_claim_made: ClaimMade) -> None:
+
```
istankovic, line 101: Nitpick: unnecessary empty line.

beamer/tests/agent/test_challenge.py
```diff
@@ -344,3 +344,72 @@ def test_withdraw_not_participant(request_manager, token, config):
...
+#                  claim
+#                                       challenge
+#                  challenge
+# challenge
+# Winner: Bob + Dave
+#
+# Note: testing that the agent can handle multiparty bidding
+def test_challenge_7(request_manager, fill_manager, token, config):
+    target_address = accounts[8]
+    requester, charlie, dave = accounts[:3]
```
istankovic, line 362: We should use `alloc_accounts` now that it is available also to agent tests.

beamer/models/claim.py
palango: I see. I'm mostly scared that accessing this somewhere will add an incorrect challenger.

beamer/tests/agent/test_challenge.py
```diff
@@ -344,3 +344,72 @@ def test_withdraw_not_participant(request_manager, token, config):
...
+# Bob              Charlie              Dave
+# -----------------------------------------------
+#                  claim
+#                                       challenge
+#                  challenge
+# challenge
+# Winner: Bob + Dave
+#
+# Note: testing that the agent can handle multiparty bidding
+def test_challenge_7(request_manager, fill_manager, token, config):
```
istankovic, line 360: According to the commit messages in this PR, this part (the agent test) should be in the next commit, not this one.

beamer/tests/agent/test_challenge.py
```diff
@@ -344,3 +344,72 @@ def test_withdraw_not_participant(request_manager, token, config):
...
+    target_address = accounts[8]
+    requester, charlie, dave = accounts[:3]
+
+    agent = beamer.agent.Agent(config)
+
+    w3 = brownie.web3
+    with earnings(w3, agent, num_fills=0) as agent_earnings, earnings(
+        w3, dave, num_fills=0
```
istankovic: Didn't we get rid of `num_fills`?

beamer/tests/agent/test_challenge.py
```diff
@@ -344,3 +344,72 @@ def test_withdraw_not_participant(request_manager, token, config):
...
+        brownie.chain.mine(timestamp=claim.termination)
+
+        request_manager.withdraw(claim.claimId, {"from": agent.address})
+
+        agent.stop()
+        agent.wait()
+
+    assert agent_earnings() == stake
+
+    request_manager.withdraw(claim.claimId, {"from": dave})
```
istankovic, line 412: I would move this `withdraw` into the `with` block above, after the agent withdraws.

contracts/contracts/RequestManager.sol
```diff
@@ -252,15 +252,13 @@ contract RequestManager is Ownable {
...
 
         if (claimerStake > challengerStakeTotal) {
             if (challengerStakeTotal == 0) {
-                periodExtension = finalizationTimes[request.targetChainId] + challengePeriodExtension;
-            } else {
-                periodExtension = challengePeriodExtension;
+                periodExtension += finalizationTimes[request.targetChainId];
             }
             require(claim.claimer != msg.sender, "Cannot challenge own claim");
             nextActor = msg.sender;
```
istankovic, line 264: Nitpick: would be nice to squash this with the first commit that touches this part.

istankovic (changes requested):
Looks quite good overall, with minor nitpicks. We do need some more tests, though.

beamer/models/claim.py
istankovic: Yes, this is one of those things where Python really is missing something.
I've commented that maybe this could be made private, which could help with this sort of thing.

beamer/tests/agent/test_challenge.py
fredo: let's do this in a different PR, as it counts for all tests

beamer/tests/contracts/test_request_manager.py
```diff
@@ -517,7 +561,7 @@ def test_withdraw_with_two_claims_first_unsuccessful_then_successful(
 
     # The first claim gets withdrawn first
     # As the challenger wins, no requests funds must be paid out
-    withdraw1_tx = request_manager.withdraw(claim1_id, {"from": requester})
+    withdraw1_tx = request_manager.withdraw(claim1_id, {"from": challenger})
```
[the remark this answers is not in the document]
fredo: This is a different unsolved problem. WIth multiple challengers there is currently no option to do so (for claimers it is). See https://github.com/beamer-bridge/beamer/issues/613.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -676,6 +720,41 @@ def test_withdraw_without_challenge_with_resolution(
...
+    # Claimer outbids again
+    request_manager.challengeClaim(claim_id, {"from": claimer, "value": claim_stake + 1})
+    # Second challenger challenges
+    request_manager.challengeClaim(claim_id, {"from": second_challenger, "value": claim_stake + 1})
+
+    # Timetravel after claim period
+    chain.mine(timedelta=finalization_time + challenge_period_extension)
+
+    # Show that last challenger can withdraw first
+    request_manager.withdraw(claim_id, {"from": second_challenger})
```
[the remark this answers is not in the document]
fredo: This is excessively tested in other tests, but will do

contracts/contracts/RequestManager.sol
fredo: claim.withdrawnAmount does not store the funds withdrawn. It is the claim stakes. funds can only ever be withdrawn once. So the require does not protect the funds of the users

fredo: the deposits are protected by `request.depositReceiver`

beamer/tests/contracts/test_request_manager.py
fredo: added the balance check to the test and changed stake amounts by the challengers to be different

contracts/contracts/RequestManager.sol
istankovic: > claim.withdrawnAmount does not store the funds withdrawn. It is the claim stakes. funds can only ever be withdrawn once. So the require does not protect the funds of the users

Right, but I was not talking about that, I was talking about the funds of liquidity providers. Those are real and we should also take care to protect them too. Without them, there would be no protocol.

beamer/tests/agent/test_challenge.py
fredo: changed it

contracts/contracts/RequestManager.sol
fredo: right. But the question is what do you really protect with this require?

I would argue that it is not necessarily protecting the protocol's funds. If funds are being stolen due to a security leak,, it might not happen on this very occasion where the require is hit but actually sooner. 
So it could be an eligible owner of the funds and he is not able to withdraw it. If this require is hit it means that the protocol is in debt but with putting the `require` in place the code simply decides to block the current execution.

And the address of the current execution might be a rightful owner of the stakes. This will prevent him from withdrawing his funds.

One should be cautious if this is really helping. And I want to raise awareness to such things. Because it is not simply a "just to be safe" condition.
Please note, that I'm not saying that it isn't a good idea to do so. But it requires a bit more thinking of all the potential cases when this require can be hit.

beamer/tests/contracts/test_request_manager.py
fredo: good thinking, I will check

fredo:
still to be changed 

- the challenger dictionary
- the require withdrawnAmount

contracts/contracts/RequestManager.sol
```diff
@@ -252,15 +252,13 @@ contract RequestManager is Ownable {
 
         address nextActor;
         uint256 minValue;
-        uint256 periodExtension;
+        uint256 periodExtension = challengePeriodExtension;
```
palango, line 255: Can you add an assertion somewhere that catches this?

fredo: What exctly do you have in mind to be asserted?

palango: Basically make sure we test this, so an assertion for the extension of the challenge period in that case.

contracts/contracts/RequestManager.sol
istankovic: > And the address of the current execution might be a rightful owner of the stakes. This will prevent him from withdrawing his funds.

You're right, I see your point.

However, for that to happen, there needs to exist another way to drain contract funds -- which we could/should consider and review.

Also, part of my motivation here is simply to prevent accidental programming mistakes in the calculation of the withdraw amount above -- some mistakes would be caught early enough then, with such a `require` statement.

In any case, feel free to ignore my comment, I don't have any strong opinions on it, it just seemed to me like a useful addition.

contracts/contracts/RequestManager.sol
istankovic: Yes, I agree with @palango here, it'd be nice to say "new termination time is _at least this much_ later than old termination time".

contracts/contracts/RequestManager.sol
fredo: I thought about it the last days. And decided to add the require. The main reason is that there is exactly one chance to exploit this (in the next call there is already an require at the beginning of the function).
If we add the require the maximum amount to be exploited is `claimer_stake + challenger_stakes`. If we do not add it the maximum amount is `balance of contract`. Even though the address might be the rightful owner, this is a trade off worth to take in respect to the best of the protocol. Anyway thanks for the discussion.

contracts/contracts/RequestManager.sol
fredo: I can also do it in the contract instead of a test.

fredo: There is `test_claim_period_extension()` which simply didn't cover that path. I extended the test.

palango: Perfect, thanks!

beamer/tests/contracts/test_request_manager.py
fredo: Paul will add it in his invalidateFillHash PR

beamer/models/claim.py
```diff
@@ -53,14 +57,19 @@ def request_id(self) -> RequestId:
...
-        return self._latest_claim_made.challenger
-
     @property
     def termination(self) -> Termination:
         return self._latest_claim_made.termination
 
+    def get_challenger_stake(self, challenger: Address) -> int:
+        if challenger in self._challenger_stakes:
+            return self._challenger_stakes[challenger]
+        return 0
```
istankovic, line 67:
```suggestion
        return self._challenger_stakes.get(challenger, 0)
```

beamer/models/claim.py
```diff
@@ -53,14 +57,19 @@ def request_id(self) -> RequestId:
...
+
+    def add_challenger_stake(self, challenger: Address, amount: int) -> None:
```
istankovic, line 69: This could be private, no?

fredo: it could be private because we don't use it right now outside. But it could be the case in the future and this would be totally fine, and I'm afraid that we forget why we chose it to be private.
I mean it is simply not used outside of the class but it's totally fine to do so.

beamer/models/claim.py
fredo: good catch :facepalm:

beamer/tests/contracts/test_request_manager.py
```diff
@@ -121,28 +123,72 @@ def test_claim_counter_challenge(request_manager, token, claim_stake):
...
+    outbid = request_manager.challengeClaim(
+        claim_id, {"from": challenger, "value": claim_stake + 1}
+    )
+    assert "ClaimMade" in outbid.events
+
+    # Claimer outbids again
+    outbid = request_manager.challengeClaim(claim_id, {"from": claimer, "value": claim_stake + 1})
+    assert "ClaimMade" in outbid.events
+
+    # The second challenger must be able to re-challenge, but must increase the stake
```
istankovic, line 156: Why is this the "second" challenger?  Above you say "first challenger", but you're using the same `challenger` address as here.

istankovic: Ah, you're referring to the role, not the address... ignore my comment.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -160,31 +206,61 @@ def test_claim_period_extension(
...
+        challenge.timestamp + challenge_period
+        == request_manager.claims(claim_id)[RM_C_FIELD_TERMINATION]
+    )
 
     # Timetravel close to end of challenge period
-    chain.mine(timedelta=challenge_period * 8 / 10)
+    chain.mine(timestamp=request_manager.claims(claim_id)[RM_C_FIELD_TERMINATION] - 10)
+
+    # Claimer challenges close to the end of challenge
+    # Should increase the challenge termination
```
istankovic, line 242: Very nice to see this test!

beamer/tests/contracts/test_request_manager.py
```diff
@@ -160,31 +206,61 @@ def test_claim_period_extension(
...
+    challenge = request_manager.challengeClaim(
+        claim_id, {"from": claimer, "value": claim_stake + 1}
+    )
+    assert (
+        challenge.timestamp + challenge_period_extension
+        == request_manager.claims(claim_id)[RM_C_FIELD_TERMINATION]
+    )
```
istankovic, line 249: Can we also check here that the new termination is strictly greater than the old one?

beamer/tests/contracts/test_request_manager.py
fredo: Oh I think the comment is wrong. I created its own test for two challengers. This one only tests counter-challenge

beamer/tests/contracts/test_request_manager.py
```diff
@@ -675,6 +751,64 @@ def test_withdraw_without_challenge_with_resolution(
...
+    request_manager.withdraw(claim_id, {"from": first_challenger})
+
+    assert (
+        web3.eth.get_balance(first_challenger.address)
+        == first_challenger_eth_balance + claim_stake + 1
+    )
+    assert (
+        web3.eth.get_balance(second_challenger.address)
+        == second_challenger_eth_balance + claim_stake + 9
+    )
```
istankovic, line 805: I am not sure whether this is already covered, but it would be nice to 
- test that the withdraw order does not matter, i.e. if the fist challenger withdraws first, the amounts withdrawn are still the same
- test that if the claimer is last and wins, all stakes go to the claimer (we have a single-challenger test for this, but I'm not sure for multi-party bidding)

contracts/contracts/RequestManager.sol
```diff
@@ -267,18 +267,24 @@ contract RequestManager is Ownable {
...
+            claim.lastChallenger = msg.sender;
+            claim.challengersStakes[msg.sender] += msg.value;
+            claim.challengerStakeTotal += msg.value;
         }
 
+
+
         claim.termination = Math.max(claim.termination, block.timestamp + periodExtension);
+        uint256 minimumTermination = block.timestamp + challengePeriodExtension;
+        require(claim.termination >= minimumTermination, "Challenge termination did not increase enough");
```
istankovic, line 279: Nitpick:
```suggestion
        require(claim.termination >= minimumTermination, "Claim termination did not increase enough");
```

contracts/contracts/RequestManager.sol
```diff
@@ -299,46 +307,74 @@ contract RequestManager is Ownable {
...
+        // If the challenger wins and is the last challenger
+        // he gets either the excess stakes if claimer was winning
+        // or twice his stake minus the difference of the claimer and challenger stakes
+        if (msg.sender == claim.lastChallenger) {
+            if (claimerStake > challengerStakeTotal) {
+                ethToTransfer += (claimerStake - challengerStakeTotal);
+            }
+            else {
+                ethToTransfer -= (challengerStakeTotal - claimerStake);
+            }
```
istankovic, line 366: The comment above this block is slightly wrong: in one case the challenger gets "the excess", but in the other he gets the full amount (2*stake minus the difference).
In order to be consistent, I'd suggest something like
```
        // If the challenger wins and is the last challenger, he gets either
        // twice his stake plus the excess stake if claimer was winning, or
        // twice his stake minus the difference of the claimer and challenger stakes
```

beamer/tests/contracts/test_request_manager.py
fredo: the second bullet point is covered

istankovic:
I would really prefer if commit 8b201f83bb2537303add8a73a791f55909cd8f8a (`contracts: allow muliparty challenges`) were split up so that
- the contract changes are in their own separate commit, as the commit message already indicates 
- the agent changes are in their own separate commit
- the tests are in their own separate commit

Not only would that have made the review a bit easier, but also it would be helpful for future debugging and history inspection.

beamer/tests/contracts/test_request_manager.py
fredo: can you elaborate what exactly you want to test ? What do you mean by the amounts withdrawn are the same?

istankovic: > can you elaborate what exactly you want to test ? What do you mean by the amounts withdrawn are the same?

I would like to test that for
```
challenger1 withdraws first, gets A ETH
challenger2 withdraws second, gets B ETH
```
and
```
challenger2 withdraws first, gets C ETH
challenger1 withdraws second, gets D ETH
```
the following holds:
```
A == D
B == C
```

fredo:
> I would really prefer if commit [8b201f8](https://github.com/beamer-bridge/beamer/commit/8b201f83bb2537303add8a73a791f55909cd8f8a) (`contracts: allow muliparty challenges`) were split up so that
> 
> * the contract changes are in their own separate commit, as the commit message already indicates
> * the agent changes are in their own separate commit ([ecdb9da](https://github.com/beamer-bridge/beamer/commit/ecdb9dab74c28f174a9364aa30560ef697864e29) looks like it would be perfect for squashing into that commit)
> * the tests are in their own separate commit
> 
> Not only would that have made the review a bit easier, but also it would be helpful for future debugging and history inspection.

I'm sorry but I really see that commits are split exactly how you describe it. The only changes in tests or in the agent are adaptions "the interface of the contract" namely the event variable names. I really doubt that future debugging would be improved by separating that into different commits. But we can disagree on this.

fredo:
> * the agent changes are in their own separate commit ([ecdb9da](https://github.com/beamer-bridge/beamer/commit/ecdb9dab74c28f174a9364aa30560ef697864e29) looks like it would be perfect for squashing into that commit)

Also this is really not perfect to squash it together. Adapting the interface and adaptaing to a new feature is somthing completely different.

beamer/tests/contracts/test_request_manager.py
fredo: added test for this

Issue #566, opened 2022-04-13
Non-Fill: #3 Non-Fill proofs from FillManager


fredo:
The FillManager.sol has to expose a new function to create non-fill proofs which invalidates a fill hash. A non fill proof means that a given fillHash is not the correct fillHash which filled the corresponding request. The fillManager contract can verify that.

- Change `mapping(bytes32 => bool) public fills;` to `mapping(bytes32 => bytes32) public fills;` where the value of the mapping is the fillHash (requestHash => fillHash).
- Add a new function to the IProofSubmitter `submitNonFillProof(address l1Resolver, bytes32 fillHash, uint256 sourceChainId)`
- Add a new function to FillManager `invalidateFillHash(bytes32 requestHash, bytes32 fillID)` which will recompute the fillHash and check whether this fill hash filled the request defined by the request hash. If not it will submit a non fill proof.

PSEUDO CODE

```sol
function invalidateFillHash(requestHash, fillId) {
    fillHash = _createFillHash(requestHash, fillId);

    if(fills[requestHash] != fillHash){
        IProofSubmitter.submitNonFillProof(fillHash);
    }
}
```

PR #601, opened 2022-04-25
Fill manager/non fill proof

cducrest:
Closes https://github.com/beamer-bridge/beamer/issues/566


contracts/contracts/FillManager.sol
```diff
@@ -55,21 +56,29 @@ contract FillManager is Ownable {
...
-        return fillId;
+        emit RequestFilled(requestId, proofReceipt.fillId, sourceChainId, targetTokenAddress, msg.sender, amount);
+
+        return proofReceipt.fillId;
+    }
+
+    function invalidateFillHash(bytes32 requestHash, bytes32 fillId, uint256 sourceChainId) external {
+        bytes32 fillHash = BeamerUtils.createFillHash(requestHash, fillId);
+        if (fills[requestHash] != fillHash) {
+            proofSubmitter.submitNonFillProof(l1Resolver, sourceChainId, fillHash);
```
palango, line 80: Should we emit an event here?

palango (approved):
Looks good, I don't have a strong opinion on the event side.

contracts/contracts/FillManager.sol
cducrest: You're right, it's a good idea.

fredo:
> Looks good, I don't have a strong opinion on the event side.

We need this event for other agents not to send another invalidation.

PR #604, opened 2022-04-26
agent: Store fill tx hash

palango:
This will be required to call the message relayer.

Part of #271


istankovic (approved):
Nice and minimalistic. :+1:

PR #612, opened 2022-04-28
agent: Add `InitiateL1ResolutionEvent`

palango:
Part of #271 

Also adds a way of the state machine adding events returned by the
individual handlers.

### DIFF
--- a/beamer/events.py
+++ b/beamer/events.py
@@ -33,6 +33,11 @@ class LatestBlockUpdatedEvent(Event):
 
 
 @dataclass(frozen=True)
+class InitiateL1ResolutionEvent(Event):
+    request_id: RequestId
+
+
+@dataclass(frozen=True)
 class TxEvent(Event):
     tx_hash: HexBytes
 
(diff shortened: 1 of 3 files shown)


beamer/events.py
```diff
@@ -32,6 +32,11 @@ class LatestBlockUpdatedEvent(Event):
     block_data: BlockData
 
 
+@dataclass(frozen=True)
+class InitiateL1ResolutionEvent(Event):
```
fredo, line 36: When is this event going to be created? Will this event unconditionally execute the L1 tx?

beamer/state_machine.py
```diff
@@ -187,11 +209,19 @@ def _handle_claim_made(event: ClaimMade, context: Context) -> bool:
...
             claim.ignore(event)
         claim.challenge(event)
+
+        if _l1_resolution_criteria_fulfilled(claim, context):
+            events = [
+                InitiateL1ResolutionEvent(
+                    chain_id=request.target_chain_id,
+                    request_id=request.id,
+                )
+            ]
```
istankovic, line 221: Why is this inside the `try...except` block?

PR #630, opened 2022-05-03
agent: Add config to Context

palango:
This will be required to call the message relayer with the correct data.

Related to #271


istankovic (approved):
Looks good. :+1:

PR #632, opened 2022-05-04
Use tx_receipt.gasUsed instead of tx.gas for gas spending calculation in tests

fredo:
a quick fix to calculate the correct gas spending by using tx_receipt.gasUsed


istankovic (approved):
Thanks a bunch for this, man! :1st_place_medal:

Issue #605, opened 2022-04-27
frontend - proper multi step actions with data persistence


weilbith:
While working on tasks of the state management with stores, testing, refactoring and also having #367 in mind, I realized that our current data representation of requests and their state is not optimal. Especially having extensibility, testability and code quality in mind.

Currently we have two different interfaces (`Request` and `RequestMetadata`) to represent a request. Additionally there is the `RequestState` which is (now, after refactoring) decoupled form both.
When the user initiates a request, multiple functions are involved into this process. These functions are defined within composables. The Vue component that includes the trigger has an internal state reference of the requests its state that and provides it to all these functions. Each function then sets this reference to new values depending on what they did.
The type of the state is an enumeration which a strict sorting. In fact this means strings mapped to numbers in a sorted array. Each update means to set the number reference to a new number. It is assumed that the number gets always increased, usually by one for each step. Visual components make use of this assumption and render each number with some text and colors. Comparing the current number to the list of numbers, the component knows which steps have been completed and what's next.
Somehow it has now been tried to add another state to represent a failing state. The state it's under-the-hood number is bigger than all other states. In turn all states are "completed" once an error occurs. If it would now be desirable to fail an any intermediate state/step it would be necessary to add new enumeration entries in the correct order.

If the above summarized concept should be unclear, please ask or look at the code from commit 79d971d08935b679aeef6997dcdf69f7693afbb8 (current HEAD, where this is the de-facto state).

This approach comes with a bunch of flaws. First of all, the concept of the sorted enumeration that maps to numbers is very implicit and fragile. The code is not very expressive and it is incredibly easy to make mistakes and mess it up. This also goes for the integration of failure states which are horrible to add into this structure. Finally the actual data representation is not stable for updates. That means when we develop #367 and safe past requests to local storage, this data must be read and parsed on each new load of the frontend. Due do the approach with a numeric enumeration, the stored data will be just pure numbers. In result the meaning of this number depends on the mapping in the enumeration. Should the enumeration ever change in a non-appending manner (means injecting or removing), the parsed data meaning changes from what it originally was. So for example this could lead to the case that a user made a request that failed at step number 3. Later, with a new version of the frontend, he loads his history again and now the step has completed or failed in a different step because number 3 maps to a different enumeration entry now.
[...]

PR #633, opened 2022-05-04
Frontend/chore/unit tests for simple vue components

### LINKED ISSUE #587: Frontend: Add unit tests for simple Vue components
Follow up of #384 and #584

Sibling issue to #586 and #589, focusing on simple Vue components. The point is that testing Vue components works slightly different. The components to focus on here are the simple and re-usable components. Those who don't include any logic about the core business logic. The goal is to identify these components and if they do not yet fulfill these requirements, refactor them. In the end it should be possible to test them all with a shallow mount (feature of the Vue test utils). The latter explicitly ensures that this are unit tests and are not affected by other components. In worst case the "unit" tests of one component need to satisfy the requirements of a child-component. This should **not** be the case for these kind of tests.

The team should argue and exchange experience if snapshot tests are a good addition here or not. The upside is that it saves the developers a lot of manual test code to ensure the component renders as expected. The downside is that the handling of failing snapshot tests needs care and should **not** be simply accepted without further investigation (simply too easy to do in comparison to adapt the tests if necessary).

We should agree to use a special selectors to all DOM elements that are in interest of a test case.

weilbith:
Closes #587

Not covered components are:
  - `RequestDialog` (too complicated right now)
  - `RequestFormInputs` (too complicated right now)
  - `Feedback` (nothing to test, inactive anyways)

This includes a few refactoring for the request processing stuff. Though I did this before coming up with #605. But that isn't bad because the code has become leaner and testable.

Nothing crazy else.


frontend/tests/unit/components/layout/Card.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+describe('Card.vue', () => {
+  it('renders given default slot', () => {
+    const wrapper = createWrapper({ slot: '<span>test</span>' });
+
+    expect(wrapper.html()).toContain('<span>test</span>');
+  });
+
+  it('has round corders', () => {
```
manuelwedler, line 21: Nitpick: I don't like too much testing the design. If something purely visual is changed in the design, we should not have to adapt the tests imo.

manuelwedler (approved):
Great work here, thanks! I like how you separated the RequestProcessing in smaller units. 
I'm approving early. Just address my minor comments before merging.

frontend/src/components/ImprintModal.vue
```diff
@@ -1,7 +1,14 @@
 <template>
-  <input id="imprint-modal" type="checkbox" class="modal-toggle" />
-  <div class="modal">
-    <div class="modal-box bg-dark w-8/12 max-w-5xl p-14">
+  <span class="btn-link" data-test="open-button" @click="openModal">Imprint</span>
+
+  <div
+    v-if="modalIsVisible"
+    class="fixed top-0 left-0 w-full h-full z-10 flex justify-center items-center bg-modal-background"
```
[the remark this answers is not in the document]
weilbith: Ah, nice. That is much better! Haven't found this syntax in the docs.

frontend/tests/unit/components/layout/Card.spec.ts
weilbith: Okay. I'll remove it. 👍🏾

weilbith: ✔️

weilbith:
Alright, I'll rebase on `main` now and thereby resolve the fixup commits. Let's see how much will break. 🙈

PR #634, opened 2022-05-04
L1 resolution

### LINKED ISSUE #271: Allow the node to submit L1 transaction

palango:
Resolves #271 

The main work for initiating and handling L1 resolution. Some follow-ups will be needed one we have the new challenge and non-fill proofs integrated.

This is hopefully readable on a per-commit basis. Some important things that are added:
- Add the L1 RPC URL to the CLI and config
- Add the `InitiateL1ResolutionEvent`. It triggers a check if the L1 resolution is viable (financially) and triggers it in this case.
  - Important caveat: This currently only takes one claim into account. Do we want to fix this?
- Resolution is triggered by adding a new task to the `resolution_pool`. This is a `ThreadPoolExecutor` with one worker, which takes care of serializing multiple parallel resolutions.
- The `RequestResolved` event is handled in the state machine and writes the correct filler into the `Request` object.

### DIFF
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -1,4 +1,5 @@
 import threading
+from concurrent.futures import ThreadPoolExecutor
 
 import structlog
 import web3
@@ -34,6 +35,10 @@ class Agent:
         self._config = config
         self._stopped = threading.Event()
         self._stopped.set()
+        # Just add one worker, as this effectively serializes the work
+        # This is necessary as we use the account for all resolutions and
+        # would otherwise run into nonce problems
+        self._resolution_pool = ThreadPoolExecutor(max_workers=1)
 
         w3_l1 = _make_web3(config.l1_rpc_url, config.account)
         w3_l2a = _make_web3(config.l2a_rpc_url, config.account)
@@ -65,6 +70,7 @@ class Agent:
             latest_blocks={},
             config=config,
             web3_l1=w3_l1,
+            resolution_pool=self._resolution_pool,
         )
         self._event_processor = EventProcessor(self.context)
 
@@ -100,6 +106,7 @@ class Agent:
         self._event_processor.stop()
         self._event_monitor_l2a.stop()
         self._event_monitor_l2b.stop()
+        self._resolution_pool.shutdown(wait=True, cancel_futures=False)
         self._stopped.set()
 
     @property
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -35,6 +35,7 @@ class Agent:
         self._stopped = threading.Event()
         self._stopped.set()
 
+        w3_l1 = _make_web3(config.l1_rpc_url, config.account)
         w3_l2a = _make_web3(config.l2a_rpc_url, config.account)
         w3_l2b = _make_web3(config.l2b_rpc_url, config.account)
 
@@ -63,6 +64,7 @@ class Agent:
             address=config.account.address,
             latest_blocks={},
             config=config,
+            web3_l1=w3_l1,
         )
         self._event_processor = EventProcessor(self.context)
 
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -71,6 +71,7 @@ class Agent:
             config=config,
             web3_l1=w3_l1,
             resolution_pool=self._resolution_pool,
+            l1_resolutions={},
         )
         self._event_processor = EventProcessor(self.context)
 
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -426,28 +426,29 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
     assert request is not None, "Active claim for non-existent request"
 
     block = context.latest_blocks[request.source_chain_id]
-    if block["timestamp"] < claim.termination:
-        # If the agent is the L1 filler, we can withdraw
-        # Condition is inverted, as the withdraw happens below
-        has_l1_resolution = (
-            request.l1_resolution_filler == context.address
-            and context.address
-            in {
-                claim.claimer,
-                claim.challenger,
-            }
-        )
-        if not has_l1_resolution:
-            return
+    has_l1_resolution = request.l1_resolution_filler is not None
+    agent_is_claimer = claim.claimer == context.address
+    agent_is_challenger = claim.challenger == context.address
 
-        # TODO: what if the none of the participants is the l1 filler
+    # When request is L1 resolved, the termination isn't important
+    if has_l1_resolution:
+        # We claimed the request
+        if agent_is_claimer:
+            assert request.l1_resolution_filler == context.address, "We cheated"
+            _withdraw(claim, context)
 
-    # The L1 resolution has precedence over the challenge game
-    claim_winner = request.l1_resolution_filler or claim.get_winning_address()
-    agent_winning = claim_winner == context.address
-    if not agent_winning:
-        return
+        # Claimer cheated
+        if claim.claimer != request.l1_resolution_filler and agent_is_challenger:
+            _withdraw(claim, context)
 
+    # Otherwise check that the challenge period is over
+    elif block["timestamp"] >= claim.termination:
+        claim_winner = claim.get_winning_address()
+        if claim_winner == context.address:
+            _withdraw(claim, context)
+
+
+def _withdraw(claim: Claim, context: Context) -> None:
     func = context.request_manager.functions.withdraw(claim.id)
     try:
         txn_hash = _transact(func)
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -5,6 +5,7 @@ import sys
 import threading
 import time
 import traceback
+from concurrent.futures import Future
 from typing import Any, Callable, Optional, cast
 
 import requests.exceptions
@@ -233,6 +234,15 @@ def process_requests(context: Context) -> None:
     to_remove = []
     for request in context.requests:
         log.debug("Processing request", request=request)
+
+        if request.l1_resolution_tx_handle is not None:
+            handle: Future = request.l1_resolution_tx_handle
+
+            if handle.done():
+                try:
+                    handle.result()
+                except Exception as ex:
+                    log.warning("L1 Resolution failed", ex=ex)
 
         if request.is_pending:
             fill_request(request, context)
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -157,7 +157,6 @@ class EventProcessor:
             if self._synced:
                 process_requests(self._context)
                 process_claims(self._context)
-                process_l1_resolutions(self._context)
 
         self._log.info("EventProcessor stopped")
 
@@ -226,22 +225,6 @@ def _transact(func: ContractFunction, **kwargs: Any) -> HexBytes:
 
     func.web3.eth.wait_for_transaction_receipt(txn_hash)
     return txn_hash
-
-
-def process_l1_resolutions(context: Context) -> None:
-    log.debug("Processing L1 resolutions", num_tasks=len(context.l1_resolutions))
-
-    to_remove = []
-    for request_id, future in context.l1_resolutions.items():
-        if future.done():
-            try:
-                future.result()
-                to_remove.append(request_id)
-            except Exception as ex:
-                log.error("L1 Resolution failed", ex=ex)
-
-    for request_id in to_remove:
-        del context.l1_resolutions[request_id]
 
 
 def process_requests(context: Context) -> None:
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -5,7 +5,6 @@ import sys
 import threading
 import time
 import traceback
-from concurrent.futures import Future
 from typing import Any, Callable, Optional, cast
 
 import requests.exceptions
@@ -158,6 +157,7 @@ class EventProcessor:
             if self._synced:
                 process_requests(self._context)
                 process_claims(self._context)
+                process_l1_resolutions(self._context)
 
         self._log.info("EventProcessor stopped")
 
@@ -228,21 +228,28 @@ def _transact(func: ContractFunction, **kwargs: Any) -> HexBytes:
     return txn_hash
 
 
+def process_l1_resolutions(context: Context) -> None:
+    log.info("Processing L1 resolutions", num_tasks=len(context.l1_resolutions))
+
+    to_remove = []
+    for request_id, future in context.l1_resolutions.items():
+        if future.done():
+            try:
+                future.result()
+                to_remove.append(request_id)
+            except Exception as ex:
+                log.error("L1 Resolution failed", ex=ex)
+
+    for request_id in to_remove:
+        del context.l1_resolutions[request_id]
+
+
 def process_requests(context: Context) -> None:
     log.info("Processing requests", num_requests=len(context.requests))
 
     to_remove = []
     for request in context.requests:
         log.debug("Processing request", request=request)
-
-        if request.l1_resolution_tx_handle is not None:
-            handle: Future = request.l1_resolution_tx_handle
-
-            if handle.done():
-                try:
-                    handle.result()
-                except Exception as ex:
-                    log.warning("L1 Resolution failed", ex=ex)
 
         if request.is_pending:
             fill_request(request, context)
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -162,15 +162,19 @@ class EventProcessor:
 
     def _process_events(self) -> None:
         iteration = 0
+        created_events: list[Event] = []
         while True:
             with self._lock:
                 events = self._events[:]
 
-            unprocessed = []
+            unprocessed: list[Event] = []
             any_state_changed = False
             for event in events:
-                state_changed = process_event(event, self._context)
+                state_changed, new_events = process_event(event, self._context)
                 any_state_changed |= state_changed
+
+                if new_events:
+                    created_events.extend(new_events)
                 if not state_changed:
                     unprocessed.append(event)
 
@@ -192,6 +196,14 @@ class EventProcessor:
             iteration += 1
             if not any_state_changed:
                 break
+
+        # New events might be created by event handlers. If they would be
+        # processed directly, there would be a chance of skipping certain states
+        # in the request/claim state machines accidentally. So instead they are
+        # collected and attached to the events list after the current batch of
+        # events has been processed.
+        with self._lock:
+            self._events.extend(created_events)
 
 
 class _TransactionFailed(Exception):
@@ -355,10 +367,9 @@ def claim_request(request: Request, context: Context) -> None:
 
 def maybe_challenge(claim: Claim, context: Context) -> bool:
     # We need to challenge if either of the following is true:
-    #
-    # 1) the claim is dishonest AND nobody challenged it yet
-    #
-    # 2) we participate in the game AND it is our turn
+    #   1) the claim is dishonest AND nobody challenged it yet
+    #   2) we participate in the game AND it is our turn
+
     request = context.requests.get(claim.request_id)
     # As per definition an invalid or expired request cannot be claimed
     # This gives us a chronological order. The agent should never garbage collect
@@ -368,6 +379,7 @@ def maybe_challenge(claim: Claim, context: Context) -> bool:
     block = context.latest_blocks[request.source_chain_id]
     if block["timestamp"] >= claim.termination:
         return False
+
     if int(time.time()) < claim.challenge_back_off_timestamp:
         return False
 
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -417,10 +417,24 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
 
     block = context.latest_blocks[request.source_chain_id]
     if block["timestamp"] < claim.termination:
-        # TODO: once L1 resolution is implemented, maybe withdraw before claim is terminated
-        return
+        # If the agent is the L1 filler, we can withdraw
+        # Condition is inverted, as the withdraw happens below
+        has_l1_resolution = (
+            request.l1_resolution_filler == context.address
+            and context.address
+            in {
+                claim.claimer,
+                claim.challenger,
+            }
+        )
+        if not has_l1_resolution:
+            return
 
-    agent_winning = claim.get_winning_address() == context.address
+        # TODO: what if the none of the participants is the l1 filler
+
+    # The L1 resolution has precedence over the challenge game
+    claim_winner = request.l1_resolution_filler or claim.get_winning_address()
+    agent_winning = claim_winner == context.address
     if not agent_winning:
         return
 
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -433,8 +433,7 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
     # When request is L1 resolved, the termination isn't important
     if has_l1_resolution:
         # We claimed the request
-        if agent_is_claimer:
-            assert request.l1_resolution_filler == context.address, "We cheated"
+        if agent_is_claimer and request.l1_resolution_filler == context.address:
             _withdraw(claim, context)
 
         # Claimer cheated
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -229,7 +229,7 @@ def _transact(func: ContractFunction, **kwargs: Any) -> HexBytes:
 
 
 def process_l1_resolutions(context: Context) -> None:
-    log.info("Processing L1 resolutions", num_tasks=len(context.l1_resolutions))
+    log.debug("Processing L1 resolutions", num_tasks=len(context.l1_resolutions))
 
     to_remove = []
     for request_id, future in context.l1_resolutions.items():
@@ -245,7 +245,7 @@ def process_l1_resolutions(context: Context) -> None:
 
 
 def process_requests(context: Context) -> None:
-    log.info("Processing requests", num_requests=len(context.requests))
+    log.debug("Processing requests", num_requests=len(context.requests))
 
     to_remove = []
     for request in context.requests:
@@ -268,7 +268,7 @@ def process_requests(context: Context) -> None:
 
 
 def process_claims(context: Context) -> None:
-    log.info("Processing claims", num_claims=len(context.claims))
+    log.debug("Processing claims", num_claims=len(context.claims))
 
     to_remove = []
     for claim in context.claims:
--- a/beamer/l1_resolution.py
+++ b/beamer/l1_resolution.py
@@ -1,10 +1,27 @@
+import shutil
 import subprocess
 
+import structlog
 from hexbytes import HexBytes
 
 from beamer.typing import URL
 
-RELAYER_EXECUTABLE = "beamer-relayer"
+
+log = structlog.get_logger(__name__)
+
+_RELAYER_EXECUTABLE = "beamer-relayer"
+
+
+def relayer_executable_exists() -> bool:
+    """Checks if the relayer executable is in PATH, return `False` if not"""
+
+    maybe_relayer = shutil.which(_RELAYER_EXECUTABLE)
+    if maybe_relayer is not None:
+        log.debug("Relayer found in PATH", relayer=maybe_relayer)
+        return True
+
+    log.warning("Relayer executable not found in PATH")
+    return False
 
 
 def run_relayer(l1_rpc: URL, l2_rpc: URL, privkey: str, tx_hash: HexBytes) -> None:
--- a/beamer/l1_resolution.py
+++ b/beamer/l1_resolution.py
@@ -1,0 +1,26 @@
+import subprocess
+
+from hexbytes import HexBytes
+
+from beamer.typing import URL
+
+RELAYER_EXECUTABLE = "beamer-relayer"
+
+
+def run_relayer(l1_rpc: URL, l2_rpc: URL, privkey: str, tx_hash: HexBytes) -> None:
+    subprocess.run(
+        [
+            _RELAYER_EXECUTABLE,
+            "--l1rpcprovider",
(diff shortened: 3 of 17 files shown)


beamer/state_machine.py
```diff
@@ -45,29 +46,32 @@ class Context:
     config: Config
 
 
-def process_event(event: Event, context: Context) -> bool:
+HandlerResult = Tuple[bool, Optional[List[Event]]]
```
fredo, line 54: Regarding returning a list of events during process events, I wonder if this is the right way to go. This means that the state output of the state in the middle will not be processed.
This might work for the L1 resolution case (although I have not thought it through) but could be problematic for others in the future.

Have you thought about generating new events in `process_requests()` ?

beamer/chain.py
```diff
@@ -405,13 +405,20 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
...
+        # If the agent is the L1 filler, we can withdraw
+        # Condition is inverted, as the withdraw happens below
+        if request.l1_resolution_filler != context.address:
+            return
 
-    agent_winning = claim.get_winning_address() == context.address
+        # TODO: what if the non of the participants the the l1 filler
+
+    # The L1 resolution has precedence over the challenge game
+    claim_winner = request.l1_resolution_filler or claim.get_winning_address()
```
fredo, line 437: the agent could also be the winner if he is not the `l1_resolution_filler`. 
If the agent is the challenger the winning condition for him is `claim.claimer != l1_resolution_filler`
If he is the claimer the winning condition is `claim.claimer == context.address`

beamer/chain.py
```diff
@@ -405,13 +405,20 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
     # This gives us a chronological order. The agent should never garbage collect
     # a request which has active claims
     assert request is not None, "Active claim for non-existent request"
+    assert context.address in {claim.claimer, claim.challenger}, "Agent is not part of claim"
```
fredo: I think we have to remove that line with the multi party bidding scheme anyway.

beamer/chain.py
```diff
@@ -408,10 +408,23 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
 
     block = context.latest_blocks[request.source_chain_id]
     if block["timestamp"] < claim.termination:
-        # TODO: once L1 resolution is implemented, maybe withdraw before claim is terminated
-        return
+        # If the agent is the L1 filler, we can withdraw
+        # Condition is inverted, as the withdraw happens below
+        has_l1_resolution = (
+            request.l1_resolution_filler == context.address and
```
fredo: please revise due to my comment below

beamer/state_machine.py
```diff
@@ -179,20 +196,30 @@ def _handle_claim_made(event: ClaimMade, context: Context) -> bool:
...
         # Agent is not part of ongoing challenge
         if context.address not in {event.claimer, event.challenger}:
             claim.ignore(event)
+
         claim.challenge(event)
     except TransitionNotAllowed:
-        return False
+        return False, events
+
+    if request.l1_resolution_tx_handle is None and request.fill_tx is not None:
```
fredo, line 212: If I understand this correctly, whenever a claim is challenged this event is emitted everytime a new claimMade event arrives.

`_handle_initiate_l1_resolution()` never returns True, so each event will stay in the event loop forever. See the other comment at `_handle_initiate_l1_resolution()`

beamer/state_machine.py
```diff
@@ -205,3 +232,57 @@ def _handle_claim_withdrawn(event: ClaimWithdrawn, context: Context) -> bool:
...
+        if claim._latest_claim_made.challenger_stake > limit:
+            return True
+    elif claim.challenger == context.address:
+        if claim._latest_claim_made.claimer_stake > limit:
+            return True
+
+    return False
+
+
+def _handle_initiate_l1_resolution(event: InitiateL1ResolutionEvent, context: Context) -> bool:
```
fredo, line 268: the function will never remove the event, and there are potentially multiple events for the same request, which could trigger the assertion error

palango: Good point, this should return `True`

beamer/chain.py
palango: Already removed this as it caused problems in the tests.

beamer/state_machine.py
palango: Another option that comes to mind is to just queue the returned events for the next run of the event loop, so that the requests and claim state machines would always be run before. What do you think about this?

palango: > I'm not quite sure whether this has been changed in the meantime, but previously the order and nature of events was irrelevant by design, i.e. we iterated multiple times over events until we could no longer advance. Has this been changed?

I don't think this changed, I just added a way for handlers to return new events.

fredo: It's not about the order. As a new feature, the state machine itself can create new events which is looped back into the current list of events. As a consequence, those events might alternate the state of  a request in one go. 
For events coming from the blockchain this is totally fine, because we don't want to process a request on an old state. I'm just wondering if we might want to process a request based on an event coming from the chain, and as an output generating an event which again will be looped back into the event processor. 

Can you see the difference in the approaches?

palango: Yes, that why I proposed this:

> Another option that comes to mind is to just queue the returned events for the next run of the event loop, so that the requests and claim state machines would always be run before. What do you think about this?

palango: currently it `[E2, E3, E4]`, but another proposal above is to not process E4 in that round of events.

fredo: As you described @istankovic order should not matter. The design still holds. But What Paul is suggesting is to not even put it in the current list, rather than postponing its processing to the next `cycle` (Maybe the term iteration is misleading here)

Terminology: 

iteration = one iteration over the list of events per cycle
cycle = every execution of `process_events()`

istankovic: I guess I was confused by this comment:

> Another option that comes to mind is to just queue the returned events for the next run of the event loop, so that the requests and claim state machines would always be run before.

My interpretation was that the ordering would matter, for some reason.
So if my new understating is correct, the only difference is, if we postpone the newly generated event for the next _cycle_, it would delay processing of that event for up to 1 second?
Thay may or may not be fine, depending on the nature of the generated event (can that event cause new events to be generated? etc.).

fredo: I think your new understanding is correct. The main difference here is that `process_requests()` and `process_claims()` runs on the "intermediary" state of the object once, and then in the next cycle the new event is applied.

We may or may not want that. It was just a concers that we kind of leapfrog a certain state if we directly apply the new event.

palango: I see Freds point and implemented the system so that newly created events will be processed in the next cycle.

palango: > Also, while I'm here: do we need a list (mutability) here?

New events will usually be attached to some list inside the handler, so I don't think it would be worthwhile to convert them to a tuple in the end.

beamer/chain.py
```diff
@@ -408,10 +408,24 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
...
+            }
+        )
+        if not has_l1_resolution:
+            return
 
-    agent_winning = claim.get_winning_address() == context.address
+        # TODO: what if the none of the participants is the l1 filler
+
+    # The L1 resolution has precedence over the challenge game
+    claim_winner = request.l1_resolution_filler or claim.get_winning_address()
```
istankovic: I'm wondering whether we can simplify the above conditions, e.g.
```python
     claim_winner = request.l1_resolution_filler
     if claim_winner is None and block["timestamp"] >= claim.termination:
        claim_winner = claim.get_winning_address()
```
And then we exit if we're not `claim_winner`.
This notably ignores multi-party bidding, much like the current code, but that work is in progress anyway.

beamer/state_machine.py
```diff
@@ -237,3 +237,18 @@ def _handle_claim_withdrawn(event: ClaimWithdrawn, context: Context) -> bool:
...
     return True
+
+
+def _handle_request_resolved(event: RequestResolved, context: Context) -> bool:
+    request = context.requests.get(event.request_id)
+    if request is None:
+        return False
+
+    request.l1_resolution_filler = event.filler
+
```
istankovic, line 242: Nitpick: unnecessary empty line.

beamer/agent.py
```diff
@@ -49,6 +50,11 @@ def __init__(self, config: Config):
         resolution_registry = l2a_contracts["ResolutionRegistry"]
         fill_manager = l2b_contracts["FillManager"]
 
+        # Just add one worker, as this effectivly serializes the work
```
istankovic, line 53: Typo: effectivly -> effectively

beamer/agent.py
```diff
@@ -49,6 +50,11 @@ def __init__(self, config: Config):
...
+        # This is necessary as we use the account for all resolutions and
+        # would otherwise run into nonce problems
+        self.resolution_pool = ThreadPoolExecutor(max_workers=1)
```
istankovic, line 56: Is there a good reason this is public?

beamer/agent.py
```diff
@@ -100,6 +107,7 @@ def stop(self) -> None:
         self._event_processor.stop()
         self._event_monitor_l2a.stop()
         self._event_monitor_l2b.stop()
+        self.resolution_pool.shutdown(wait=True, cancel_futures=False)
```
istankovic, line 110: Why do set `cancel_futures` to `False` here?

beamer/chain.py
```diff
@@ -225,6 +226,15 @@ def process_requests(context: Context) -> None:
...
 
+        if request.l1_resolution_tx_handle is not None:
+            handle: Future = request.l1_resolution_tx_handle
+
+            if handle.done():
+                try:
+                    handle.result()
+                except Exception as ex:
+                    log.warning("L1 Resolution failed", ex=ex)
+
```
istankovic, line 246: There are a couple of things here:
- I think this should be moved outside of this function, preferably into its own function which would then be called from the caller of this one (this block does not really relate to request processing)
- why do we catch all exceptions here?
- why do we log exceptions via `warning` instead of `error`?

beamer/l1_resolution.py
```diff
@@ -0,0 +1,26 @@
+import subprocess
+
+from hexbytes import HexBytes
+
+from beamer.typing import URL
+
+RELAYER_EXECUTABLE = "beamer-relayer"
```
istankovic, line 12: Does this need to be public?

beamer/l1_resolution.py
```diff
@@ -0,0 +1,26 @@
...
+def run_relayer(l1_rpc: URL, l2_rpc: URL, privkey: str, tx_hash: HexBytes) -> None:
+    subprocess.run(
+        [
+            RELAYER_EXECUTABLE,
+            "--l1rpcprovider",
+            l1_rpc,
+            "--l2rpcprovider",
+            l2_rpc,
+            "--l1wallet",
+            privkey,
```
istankovic, line 36: Is `privkey` really the private key itself? If so, we should reconsider other ways of providing that data to the relayer, for example via environment variables. Command line arguments are trivially visible.

beamer/l1_resolution.py
```diff
@@ -0,0 +1,26 @@
...
+            "--froml2transactionhash",
```
istankovic, line 37: Nitpick: joining words makes for a harder-to-read option. Usually the convention is to use `-`, e.g. `--from-l2-transaction-hash`, `--l1-wallet`, `--l2-rpc-provider`...

beamer/models/request.py
```diff
@@ -31,7 +32,8 @@ def __init__(
         self.filler: Optional[ChecksumAddress] = None
         self.fill_tx: Optional[HexBytes] = None
         self.fill_id: Optional[int] = None
-        self.l1_resolution_filler = Optional[ChecksumAddress]
+        self.l1_resolution_filler: Optional[ChecksumAddress] = None
+        self.l1_resolution_tx_handle: Optional[Future] = None
```
istankovic, line 36: Do we really need the future handle on `Request`?

beamer/l1_resolution.py
```diff
@@ -1,12 +1,29 @@
...
+
+log = structlog.get_logger(__name__)
+
 RELAYER_EXECUTABLE = "beamer-relayer"
 
 
+def healthcheck() -> bool:
+    """Checks if the relayer executable is in PATH, return `False` if not"""
+
+    maybe_relayer = shutil.which(RELAYER_EXECUTABLE)
```
istankovic, line 18: This is asking for trouble, doubly so given the relayer executable is being given the private key. It would be better if the relayer executable were specified explicitly, either via a command-line argument or the sorely-missing beamer agent configuration file.

beamer/l1_resolution.py
```diff
@@ -1,12 +1,29 @@
...
 from beamer.typing import URL
 
+
+log = structlog.get_logger(__name__)
+
 RELAYER_EXECUTABLE = "beamer-relayer"
 
 
+def healthcheck() -> bool:
+    """Checks if the relayer executable is in PATH, return `False` if not"""
```
istankovic, line 16: This is a weird name for a function that checks the existence of a executable.

beamer/state_machine.py
```diff
@@ -219,17 +209,14 @@ def _handle_claim_made(event: ClaimMade, context: Context) -> HandlerResult:
...
-        )
-        log.info("Initiated L1 resolution", request=request, claim_id=event.claim_id)
+    if request.l1_resolution_tx_handle is None and request.fill_tx is not None:
+        events = [
+            InitiateL1ResolutionEvent(
+                chain_id=request.target_chain_id,  # Resolution happens on the target chain
+                request_id=request.id,
+                claim_id=claim.id,
+            )
+        ]
```
istankovic, line 218: I'm wondering, what will happen if we ignore this claim and then encounter this even in the next iteration?

beamer/tests/agent/test_handlers.py
```diff
@@ -169,3 +175,32 @@ def test_request_garbage_collection_with_claim(token, config: Config):
...
+
+    claim = make_claim(request, claimer=config.account.address)
+    context.claims.add(claim.id, claim)
+
+    # Must only be called if request is filled
+    with pytest.raises(AssertionError, match="Request not yet filled"):
+        process_event(event, context)
+
+    # Check that task is added to resolution pool
+    context.web3_l1.eth.gas_price = 1
```
istankovic: Why do we change gas price here?
Would this change persist across all later tests?

beamer/tests/agent/test_handlers.py
```diff
@@ -228,3 +229,26 @@ def test_handle_request_resolved(token: Contract, config: Config):
...
+    context = make_context(config)
+
+    request = make_request(token)
+    request.fill(config.account.address, b"", b"")
+    context.requests.add(request.id, request)
+
+    claim = make_claim(request, claimer=config.account.address)
+    context.claims.add(claim.id, claim)
+
+    event = deepcopy(claim._latest_claim_made)
```
istankovic, line 263: :warning: private attribute access. Please use public API (if there is not one, consider adding it).

beamer/chain.py
```diff
@@ -197,6 +198,14 @@ def _process_events(self) -> None:
             if not any_state_changed:
                 break
 
+        # New event might be created by event handlers. If they would be
```
istankovic, line 201: Typo: event -> events

istankovic (changes requested):
I'm going to stop short of completing the review as there is already a lot and some the things I commented on are already changed.

beamer/l1_resolution.py
palango: Any recommendation?

beamer/l1_resolution.py
palango: We can always do that later, for now we have full control about the name of the binary, so I don't see why we shouldn't hardcode it.

istankovic: The problem is, one may not be aware of the current PATH settings, which is very much exploitable. For our own test cases, for sure, that does not matter, but we're not designing systems for our test cases.
In any case, if we are to leave this as-is, I'd very much prefer to switch to another scheme before advertising that other people should run their agents.

palango: Does that threat model make sense? The whole agent is a hot wallet, so if anybody can put programs into the path, there will be bigger problems anyways.

istankovic: > Does that threat model make sense? The whole agent is a hot wallet, so if anybody can put programs into the path, there will be bigger problems anyways.

True to a very large extent, but the point is not about completely eliminating all possible attack vector, it's about reducing the number of them.

beamer/l1_resolution.py
istankovic: `relayer_executable_exists` maybe? Something along those lines, if we're going to retain the current `which` approach.

beamer/state_machine.py
```diff
@@ -45,29 +46,32 @@ class Context:
...
-        return _handle_request_created(event, context)
+        return _handle_request_created(event, context), None
 
     elif isinstance(event, RequestFilled):
-        return _handle_request_filled(event, context)
+        return _handle_request_filled(event, context), None
 
     elif isinstance(event, DepositWithdrawn):
-        return _handle_deposit_withdrawn(event, context)
+        return _handle_deposit_withdrawn(event, context), None
```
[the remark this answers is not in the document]
palango: ✔️

beamer/l1_resolution.py
palango: Yes, we need to pass this information to the relayer. What's the recommended approach for this?

beamer/chain.py
palango: This has already been reworked in a later commit.

beamer/chain.py
palango: 1) An alternative I see is putting all the futures in a separate queue and poll theme there. This would also remove them from the `Request` objects for a simple boolean flag instead. Opinions on that?
2) `result` forwards all exceptions the internal call might yield. So this is the safe thing to do here.

beamer/agent.py
istankovic: What troubles me here is the fact that that can potentially take a long time. The idea behind `stop` was to stop regardless of the state we're in.

beamer/l1_resolution.py
istankovic: An environment variable, for example, would be an easy way.

beamer/models/request.py
istankovic: > Yes, we need a way top track the tasks progress for this request.

Isn't the state machine supposed to be doing that?
This feels like it should rather be a new state, e.g. `submitted_l1_proof`, or something like that.

beamer/chain.py
istankovic: > 1. An alternative I see is putting all the futures in a separate queue and poll theme there. This would also remove them from the `Request` objects for a simple boolean flag instead. Opinions on that?

Hmm... not a bad idea. But then I would go a step further and introduce a new thing that manages L1 resolutions,
e.g. `L1ResolutionManager` or something like that, which would be common to all (future) chains.
The thing would run in its own thread, like the event monitors and the even processor we already have.
Then it would take care of all things related to L1 resolution and you would merely need to track the state on request,
which would be in line with our state machine vision.

What do you think?

>     2. `result` forwards all exceptions the internal call might yield. So this is the safe thing to do here.

Yes, but I'm wondering why do we catch _all_ exceptions here?
If an exception happens, we would like to be notified about it by shutting down the agent. This way, it will be buried in the logs.

palango: > Hmm... not a bad idea. But then I would go a step further and introduce a new thing that manages L1 resolutions,
> e.g. `L1ResolutionManager` or something like that, which would be common to all (future) chains.
> The thing would run in its own thread, like the event monitors and the even processor we already have.
> Then it would take care of all things related to L1 resolution and you would merely need to track the state on request,
> which would be in line with our state machine vision.

I see the idea, but think it would be mostly a simple wrapper around the `ThreadPoolExecutor` anyways, so I'd rather avoid it. The whole L1 resolution is nothing more than a list of futures, and my expectation is that the futures will be short-lived as well.

beamer/tests/agent/test_handlers.py
```diff
@@ -267,3 +278,62 @@ def test_handle_claim_made():
             claim_id=CLAIM_ID,
         )
     ]
+
+
+@patch("beamer.chain._withdraw")
+def test_maybe_claim_no_l1(patched_withdraw: MagicMock):
```
istankovic, line 278: I really prefer not to this. Is there not another way that you can check what you need, without resorting to patching private functions or making the function public?

beamer/chain.py
```diff
@@ -228,22 +228,29 @@ def _transact(func: ContractFunction, **kwargs: Any) -> HexBytes:
     return txn_hash
 
 
+def process_l1_resolutions(context: Context) -> None:
+    log.info("Processing L1 resolutions", num_tasks=len(context.l1_resolutions))
```
istankovic, line 232: We should use `debug` here, both because in other similar places we use `debug` and also because it's going to get called every second or so.

beamer/state_machine.py
```diff
@@ -278,13 +279,16 @@ def _handle_initiate_l1_resolution(
...
-        request.l1_resolution_tx_handle = context.resolution_pool.submit(
+        future = context.resolution_pool.submit(
             run_relayer,
             context.config.l1_rpc_url,
             context.config.l2b_rpc_url,
             context.config.account.key,
             request.fill_tx,
         )
+        context.l1_resolutions[request.id] = future
+        request.l1_resolution_started = True
```
istankovic, line 290: It seems we could use the `request.id in context.l1_resolutions` to find out whether L1 resolution has been started, thereby removing the need for `l1_resolution_started`.
However, having said that, I still think we should re-evaluate the request state machine and make this explicit in the state, because that is what our state machine serves for.

istankovic, line 290: We could consider using the future's `add_done_callback` here so that we don't need a separate `process_l1_resolutions` function -- the futures can remove themselves from the dict once they are done.

palango: > It seems we could use the `request.id in context.l1_resolutions` to find out whether L1 resolution has been started, thereby removing the need for `l1_resolution_started`.

That doesn't work, because we remove the future once it's finished.

> However, having said that, I still think we should re-evaluate the request state machine and make this explicit in the state, because that is what our state machine serves for.

This would duplicate most states, so I don't really like it.

istankovic: > That doesn't work, because we remove the future once it's finished.

Ah, right.
 
> > However, having said that, I still think we should re-evaluate the request state machine and make this explicit in the state, because that is what our state machine serves for.
> 
> This would duplicate most states, so I don't really like it.

We just had a call about this.
The idea here is not to duplicate any of the states, but just add a new one, e.g. `L1ResolutionInitiated`.
The new state could be entered only from `Claimed` and exited only to `Withdrawn`.
We would enter this new state upon seeing the relevant blockchain event, regardless of who actually initiated
L1 resolution.
Once the request is resolved, we simply do a withdraw, as usual.

palango: Done

beamer/tests/agent/test_handlers.py
```diff
@@ -167,3 +196,194 @@ def test_request_garbage_collection_with_claim(token, config: Config):
...
+    )
+    context.claims.add(claim.id, claim)
+
+    # Make sure we're outside the challenge period
+    block = context.latest_blocks[request.source_chain_id]
+    assert block["timestamp"] >= claim.termination
+
+    assert claim.transaction_pending is False
+    process_claims(context)
+    assert claim.transaction_pending is False
```
istankovic, line 302:
```suggestion
    assert not claim.transaction_pending
```
PEP8.

beamer/tests/agent/test_handlers.py
```diff
@@ -167,3 +196,194 @@ def test_request_garbage_collection_with_claim(token, config: Config):
...
+        claimer_stake=Wei(100),
+        challenger_stake=Wei(50),
+        termination=Termination(TIMESTAMP + 1),
+    )
+    context.claims.add(claim.id, claim)
+
+    # Make sure we're inside the challenge period
+    assert block["timestamp"] < claim.termination
+
+    assert claim.transaction_pending is False
```
istankovic, line 335:
```suggestion
    assert not claim.transaction_pending
```

beamer/tests/agent/test_handlers.py
```diff
@@ -167,3 +196,194 @@ def test_request_garbage_collection_with_claim(token, config: Config):
...
+    process_claims(context)
+    assert claim.transaction_pending is False
```
istankovic, line 337:
```suggestion
    assert not claim.transaction_pending
```

beamer/tests/agent/test_handlers.py
```diff
@@ -167,3 +196,194 @@ def test_request_garbage_collection_with_claim(token, config: Config):
...
+
+    request = make_request()
+    request.fill(config.account.address, b"", b"")
+    request.l1_resolution_filler = l1_filler
+    context.requests.add(request.id, request)
+
+    claim = make_claim(request=request, claimer=config.account.address, termination=termination)
+    context.claims.add(claim.id, claim)
+
+    assert claim.transaction_pending is False
```
istankovic, line 353:
```suggestion
    assert not claim.transaction_pending
```

beamer/tests/agent/test_handlers.py
```diff
@@ -167,3 +196,194 @@ def test_request_garbage_collection_with_claim(token, config: Config):
...
+    process_claims(context)
+
+    # If agent is the filler, `withdraw` should be called, otherwise not
+    if l1_filler == context.address:
+        assert claim.transaction_pending is True
```
istankovic, line 358:
```suggestion
        assert claim.transaction_pending
```

beamer/tests/agent/test_handlers.py
```diff
@@ -167,3 +196,194 @@ def test_request_garbage_collection_with_claim(token, config: Config):
...
+    else:
+        assert claim.transaction_pending is False
```
istankovic, line 360:
```suggestion
        assert not claim.transaction_pending
```

beamer/tests/agent/test_handlers.py
```diff
@@ -167,3 +196,194 @@ def test_request_garbage_collection_with_claim(token, config: Config):
...
+
+    claim = make_claim(
+        request=request,
+        claimer=ADDRESS1,
+        challenger=config.account.address,
+        termination=termination,
+    )
+    context.claims.add(claim.id, claim)
+
+    assert claim.transaction_pending is False
```
istankovic, line 381:
```suggestion
    assert not claim.transaction_pending
```

beamer/tests/agent/test_handlers.py
```diff
@@ -167,3 +196,194 @@ def test_request_garbage_collection_with_claim(token, config: Config):
...
+    process_claims(context)
+
+    # If agent is the challenger and the claimer cheated,
+    # `withdraw` should be called, otherwise not
+    if l1_filler != ADDRESS1:
+        assert claim.transaction_pending is True
```
istankovic, line 387:
```suggestion
        assert claim.transaction_pending
```

beamer/l1_resolution.py
palango: Pushed into https://github.com/beamer-bridge/beamer/issues/649

beamer/l1_resolution.py
palango: I created an issue to discuss and decide this. https://github.com/beamer-bridge/beamer/issues/650

beamer/state_machine.py
palango: That version is implemented in bc111f054b70014fdc8b4a4f33a84e434d3cfcf7

But I'm not sure if I prefer it over the other, more explicit one. You're choice.

beamer/l1_resolution.py
istankovic: Sure!

beamer/state_machine.py
istankovic: The one difference that comes to my mind is that the latter might result in removal of the future from the dict somewhat earlier than the former, due to the former being tied to the loop processing. But I don't think we should care too much about that, so I'm also fine with your original approach.

beamer/models/request.py
```diff
@@ -31,24 +31,35 @@ def __init__(
         self.filler: Optional[ChecksumAddress] = None
         self.fill_tx: Optional[HexBytes] = None
         self.fill_id: Optional[int] = None
+        self.l1_resolution_filler: Optional[ChecksumAddress] = None
 
     pending = State("Pending", initial=True)
     filled = State("Filled")
     claimed = State("Claimed")
+    l1_resolved = State("L1Resolved")
```
istankovic, line 39: Could you please also update the state machine diagram in the docs to reflect this new state?

istankovic (approved):
Looks good to me. It's been quite some effort. :1st_place_medal:

istankovic:
Minor nitpick: it'd be be nice if some of the fixup commits were squashed.
Apart from that, this is a nice commit series!

beamer/models/request.py
palango: Sure

PR #637, opened 2022-05-05
frontend: Refactor transfer code into single composable

### LINKED ISSUE #534: frontend: Make proper use of composition API (refactoring)
There are a couple places where we didn't fully follow the ideas of the vue 3 composition API. See https://vuejs.org/guide/reusability/composables.html for reference.

Components like `RequestFormInputs` grew quite big. It would be better to refactor logical units into composables. This will provide better testability, readability and reusability. For example, requesting the faucet could be one composable.

 The `useRequestTransaction` file provides several composables. By convention, every composable should be in its own file.

manuelwedler:
Closes #534 

Before we had two composables, one for the fulfillment and one for the request.
They were not adding much value, because their error variables were not used.
This change also removes the request construction code from the RequestDialog to make it cleaner.

According to #605, we want to have methods for running our actions attached to a Transfer object in the future (which is kind of the Request object at the moment).
This should then be created in the new `useTransfer` composable. Having one composable facilitates these changes.

Also, now the errors of a sent request and the fulfillment are handled properly, in one variable of the `useTransfer` variable.


weilbith (approved):
👍🏾

Issue #472, opened 2022-03-30
Add WalletConnect


czepluch:
# Rationale
Add support for [WalletConnect](https://walletconnect.com/). This will allow users to easily use the app without having to rely only on MetaMask. 

# Implementation
See https://docs.walletconnect.com/

czepluch:
#456 related

PR #647, opened 2022-05-09
Integrate WalletConnect to Dapp

nekhaly:
To support non Metmask users, WalletConnect support has been provided  
through this PR. This still lacks the proper UI and some minor issues  
are still around. The core idea here is two things, first is to  
implement `EthereumProvider` interface just like `MetamaskProvider` and  
then adapt the different functions with the slight differnces between  
Metamask and WalletConnect. The second thing, is proposing the store  
persistor as it was manadatory for keeping the state of the connected  
wallet (mainly for WalletConnect), so the user is kept connected even  
after reloading the screen. `pinia-plugin-persist` has been used for  
this goal, and in configuration, `localStorage` has been chosen as the  
storage technology (API) for browser. A new store `settings` has been  
added to keep the connection state and a unit test has been created  
for it.  
Testing should be added for the provider classes and the proper UI  
should be applied before merging this, including the disconnect from  
wallet feature. We have also and issue with showing the source chain  
value in select box after changing because window reloads and the value  
resets and is no longer available after that! A fix will be pushed  
later for this.

Resolves:  
#472


frontend/src/stores/settings.ts
```diff
@@ -0,0 +1,25 @@
+import { defineStore } from 'pinia';
+
+import { Settings, WalletType } from '@/types/settings';
```
weilbith, line 3: I don't think need to define `Settings` externally here. There is only this store who cares about the type anyway. All sub-types that are used within the store might be interesting. But be aware that type inheritance also works out of the box.

frontend/src/components/RequestDialog.vue
```diff
@@ -117,14 +134,22 @@ const {
...
   if (provider.value) {
     requestSigner(provider.value);
+    settings.setConnectedWallet(WalletType.Metamask);
   }
 };
 
+const connectWalletConnect = async () => {
+  const rpcUrlsList = useRpcUrls(configuration.chains);
+  provider.value = await createWalletConnectProvider(rpcUrlsList);
+  settings.setConnectedWallet(WalletType.WalletConnect);
```
manuelwedler, line 150: createWalletConnectProvider may return undefined. I think in this case we should not call this method.

frontend/src/stores/settings.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const useSettings = defineStore('settings', {
+  state: () => ({
+    settings: undefined as Settings,
+  }),
+  actions: {
+    setConnectedWallet(connectedWallet: WalletType) {
+      this.settings = { ...this.settings, connectedWallet };
```
weilbith, line 11: If I'm not mistaken this is not necessary. Pinia is reactive by default and it doesn't require to replace the whole property. In turn, I think you can get rid of the whole action and just set the property directly where needed. That is how Pinia is meant to work.

frontend/src/stores/settings.ts
```diff
@@ -0,0 +1,25 @@
...
+    },
+  },
+  getters: {
+    connectedWallet: (state): WalletType | undefined => state.settings?.connectedWallet,
```
weilbith, line 15: Will become "useless" as soon as you flatten the state.

frontend/src/types/settings.ts
```diff
@@ -0,0 +1,11 @@
+export type Settings =
+  | {
+      connectedWallet: WalletType | undefined;
+    }
+  | undefined;
+
+// Values must not be changed for backward compatibility
+export enum WalletType {
+  Metamask = 'metamask',
```
weilbith, line 9: Nitpick^10:

```suggestion
  MetaMask = 'metamask',
```

frontend/src/composables/useRpcUrls.ts
```diff
@@ -0,0 +1,8 @@
+import { ChainConfigMapping } from '@/types/config';
+
+export function useRpcUrls(chains: ChainConfigMapping) {
```
manuelwedler, line 3: This is not a composable. Composables encapsulate stateful logic and return some functions and/or references.
This looks like a helper function to me.

frontend/src/composables/useRpcUrls.ts
```diff
@@ -0,0 +1,8 @@
...
+  const rpcUrlsList = Object.keys(chains).reduce((prev, curr) => {
```
manuelwedler, line 4: Also the naming "list" seems not appropriate to me, because we are returning an object.

frontend/tests/unit/stores/settings.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    const pinia = createPinia();
+    setActivePinia(pinia);
+  });
+
+  describe('setConnectedWallet', () => {
+    it('can set a connected wallet', () => {
+      const settings = useSettings();
+      settings.setConnectedWallet(WalletType.Metamask);
+      expect(settings.$state.settings?.connectedWallet).toBe(WalletType.Metamask);
+    });
```
weilbith, line 19: I expect the action to become removed. So in this case this comment becomes useless. But anyway: you should check first if `connectedWallet` is undefined. Else the execution of `setConnectedWallet` is not correctly tested as this test does not implicitly set or verify that it was a different value before calling the function.

So either add a `expect(settings.settings?.connectedWallet).toBeUndefined()` or you do `settings.settings = { connectedWallet = undefined }`. The second has the advantage to make the test work even we should decide to (for what ever reason) have a different default state.

frontend/src/services/web3-provider/helper.ts
```diff
@@ -0,0 +1,31 @@
...
+import { useSettings } from '@/stores/settings';
+import { WalletType } from '@/types/settings';
+
+import { createMetaMaskProvider, createWalletConnectProvider } from './';
+import { EthereumProvider } from './types';
+
+export async function getConnectedWalletProvider(): Promise<EthereumProvider | undefined> {
+  const configuration = useConfiguration();
+  const settings = useSettings();
+  const rpcUrlsList = useRpcUrls(configuration.chains);
```
manuelwedler, line 12: same as the other comment: stores and composables should not be accessed in helper functions.

frontend/tests/unit/stores/settings.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  describe('connectedWallet getter', () => {
+    it('can get connected wallet value', () => {
+      const settings = useSettings();
+      settings.setConnectedWallet(WalletType.Metamask);
```
weilbith, line 25: Please try to avoid using other functions that are under test within the test of another one. This means if one is broken also the other test fails. Rather set the expected state manually via `settings.settings.connectedWallet = WalletType.MetaMask;`. Following this concept doesn't always work, but here it does and should be followed.

weilbith (changes requested):
Thanks for all the time you put into this feature. It will be great to have WalletConnect in our application! 🙃 
There are a couple of comments, but many of them relate to each other. So it isn't as much as it might seam. After all I think we have a couple of options to improve our `EthereumProvider`. But that should not be done within this PR.

frontend/src/components/RequestDialog.vue
```diff
@@ -117,14 +134,22 @@ const {
   error: requestSignerError,
 } = useRequestSigner();
 
-const runRequestSigner = () => {
+const connectMetamask = async () => {
   // TOOD: In future we will not separate getting provider and signer which
   // resolve the undefined provider case.
+  provider.value = await createMetaMaskProvider();
```
manuelwedler, line 140: I think we should show some feedback to the user if there is no MetaMask installed. But maybe that's already going to be addressed with the proper UI.

Also the button will not block / show the spinner until the requestSigner method is called. I think for async calls, it should block from the beginning.

manuelwedler (changes requested):
Thank you for the work on this important feature! Quite nice to have this. There are couple of comments, but a lot of nitpicks. Many of them might already be raised by Thore.

frontend/src/components/RequestProcessing.vue
```diff
@@ -24,7 +24,7 @@ const props = defineProps<Props>();
 
 const progressSteps = computed(() => [
   {
-    label: 'Please confirm your request on Metamask',
+    label: 'Please confirm your request on wallet',
```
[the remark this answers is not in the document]
nekhaly: I now see that I was in a rush enough to use the slang version of the language which ruins the grammar very well :D! Will go over this kind of stuff and fix it!

frontend/src/components/RequestDialog.vue
```diff
@@ -3,7 +3,7 @@
     <div class="h-14">
       <div v-if="ethereumProvider.signer" class="flex flex-row gap-4 justify-center items-center">
         <div class="h-7 w-7 rounded-50 border-4 border-solid border-teal-light bg-green"></div>
-        <span class="text-lg">You are currently connected via Metamask</span>
```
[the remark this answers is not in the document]
nekhaly: Yes, you are right... Actually, I was hoping for more than that so I didn't waste my time making it! I need a full wallet detailed box like that for Gnosis Safe app :)

frontend/src/components/RequestDialog.vue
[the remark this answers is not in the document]
nekhaly: I think in this certain case, yes... The assumption is the scenario of the user asks to connect through Metamask (user is not logged in), then when the box appears, the user doesn't login on Metamask, and then the user returns to the screen without assigning a provider, so he chooses again between WalletConnect and Metmask again!  
The problem I just realized is that the user has logged in but didn't connect to one of the accounts! Then your assumption is true about this case I think!

nekhaly: > I think we should show some feedback to the user if there is no MetaMask installed. But maybe that's already going to be addressed with the proper UI.
> 
I think this should be handled like Gnosis Safe app as well. We don't show the user anything if no Metamask on his/her browser, and if the user has decided to connect, we redirect him/her to the download page of Metamask!

> Also the button will not block / show the spinner until the requestSigner method is called. I think for async calls, it should block from the beginning.
Yup, nice caught.. I've seen it but totally forgot to fix it.. Thank you!

frontend/src/components/RequestDialog.vue
nekhaly: yes, right.. will do!

frontend/src/components/RequestDialog.vue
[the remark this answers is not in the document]
weilbith: Okay... So does that mean you will do it like I asked? 😬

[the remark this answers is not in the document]
nekhaly: The role applies to both approaches. I should think about a 3rd way :grimacing:!

frontend/tests/unit/stores/settings.spec.ts
nekhaly: It was already there and has been removed an idiot move! will add it again! Thank you!

frontend/tests/unit/stores/settings.spec.ts
nekhaly: That was a stupid assumption as well.. It seems I was asleep while implementing this because I assumed it will not reach this test case except passing the previous one!

frontend/src/types/settings.ts
nekhaly: Not a nitpick! I checked MetaMask lib and found that they use the camel-case version. I thought at the beginning that they deal with it as one word!

frontend/src/services/web3-provider/metamask-provider.ts
```diff
@@ -102,7 +102,7 @@ export class MetaMaskProvider implements EthereumProvider {
         },
       });
       if (!wasAdded) {
-        throw new Error("Couldn't add token to metamask");
+        throw new Error("Couldn't add token to Metamask");
```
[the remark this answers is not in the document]
nekhaly: Would use the camel case in variable names but not for user interfacing text! but I agree with capitalizing the first letter!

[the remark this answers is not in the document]
nekhaly: OK.. Looks like I'm wrong again in my assumption.. even user interfacing text for MetaMask is camel-cased! will unify it over the app!

frontend/src/services/web3-provider/walletconnect-provider.ts
```diff
@@ -0,0 +1,132 @@
...
+          },
+        },
+      });
+      if (!wasAdded) {
+        throw new Error("Couldn't add token to WalletConnect");
+      }
+    } catch (error) {
+      console.log(error);
+    }
+  }
```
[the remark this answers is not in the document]
nekhaly: Yup! That's a good option! I was looking for a better design pattern to give more flexibility, but I think the one you suggest is enough so it's not complicated!

frontend/src/composables/useRequestFee.ts
```diff
@@ -12,7 +12,12 @@ export function useRequestFee(
   const error = ref<string | undefined>(undefined);
   const amount = ref<number>(0);
 
-  const show = computed(() => !!provider.value && !!requestManagerAddress);
+  const show = computed(
+    () =>
+      !!provider.value &&
+      isSupportedChain(provider.value.chainId.value) &&
```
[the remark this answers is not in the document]
nekhaly: Yes, you are right! The issue was is that we were checking using the reference of requestManagerAddress which was always true.. We should have used requestManageAddress.value instead! Anyway, in my opinion, I'll remove requestManagerAddress check and use isSupportedChain because it gives a more meaning for the check is about IMO!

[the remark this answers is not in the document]
nekhaly: And this would include for sure enhancing the helper functions themeselves!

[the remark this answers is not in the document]
nekhaly: A generic fix for the whole composables issues is gonna be provided!

frontend/src/services/web3-provider/helper.ts
```diff
@@ -0,0 +1,31 @@
+import { useRpcUrls } from '@/composables/useRpcUrls';
```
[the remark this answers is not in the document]
nekhaly: totally removed!

frontend/src/composables/useRequestFee.ts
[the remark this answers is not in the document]
nekhaly: helper has been removed and code has been transferred into proper places (getters in stores or within `useWallet` composable)

frontend/src/services/web3-provider/walletconnect-provider.ts
```diff
@@ -0,0 +1,132 @@
...
+      return null;
+    }
+  }
+
+  async addChain(chainData: ChainData): Promise<boolean> {
+    try {
+      const { chainId, name, rpcUrl } = chainData;
+      const chainIdHexValue = hexValue(chainId);
+      const walletConnectNetworkData = getNetwork(chainId);
+      walletConnectNetworkData?.name !== 'unknown' ? walletConnectNetworkData?.name : name;
```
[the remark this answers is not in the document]
nekhaly: I think is means if `walletConnectNetworkData === undefined || walletConnectNetworkData.name === 'unknown` then use the `name` from configuration!

nekhaly:
Few comments that are unresolved, but it's ready to review right now IMO!

frontend/src/components/RequestDialog.vue
nekhaly: The spinner issue has been fixed!

nekhaly:
Tested after rebase and works well! To the merge!

weilbith:
The change requests by Manuel are still blocking the merge. As Manuel only works 1.5 days a week, we can't wait for that. So I'll use my permissions to force merge.

istankovic:
:rocket: :fireworks: :wine_glass:

PR #655, opened 2022-05-11
Contract formatting

palango:
We need some standard here.


istankovic (approved):
Thank you for doing this.

Issue #645, opened 2022-05-09
Resolution registry does not invalidate all claims of request


fredo:
This is a bug. The resolution registry maps from fill hash to filler. That implies that an attacker who presents a false claim with a random fill id will not be invalidated by the fill proof of the real filler. This attack only happens if the real filler hasn't withdrawn yet. But it is possible that this happens.

# Decision
Instead of sending the fill hash we will send the request hash and fill id separately. This counts both for Non fill as well as fill proof.
Functions must be renamed accordingly in the resolution registry. In the Fill manager the fill hash can remain.
The `fillers` mapping changes to `mapping request hash => filler, fill id`. 
For the invalid fill hashes the fill hash could remain as is. There is no need to separate the two in the mapping. But they must arrive separated upon invalidation in order to check against the fillers registry.

fredo:
decided on b)

we need to separate `fillhash to tuple(request_hash, fill id)` both in `fillers` and `invalidFillHashes`

PR #666, opened 2022-05-12
Withdraw invalid hash

fredo:
Check upon withdraw if the fillhash is invalidated

### DIFF
--- a/beamer/tests/contracts/test_request_manager.py
+++ b/beamer/tests/contracts/test_request_manager.py
@@ -1,8 +1,10 @@
 import brownie
+import pytest
 from brownie import chain, web3
 from brownie.convert import to_bytes
 from eth_utils import to_hex
 
+from beamer.tests.agent.utils import make_address
 from beamer.tests.constants import RM_C_FIELD_TERMINATION
 from beamer.tests.util import alloc_accounts, create_fill_hash, earnings, make_request
 from beamer.typing import ClaimId, Termination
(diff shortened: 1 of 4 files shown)


contracts/contracts/ResolutionRegistry.sol
```diff
@@ -30,10 +32,15 @@ contract ResolutionRegistry is CrossDomainRestrictedCalls {
         bytes32 fillHash,
         uint256 resolutionChainId
     ) external restricted(resolutionChainId, msg.sender) {
+        require(
+            fillers[fillHash] == address(0),
+            "Cannot invalidate resolved fillHashes"
```
istankovic, line 37: I'm thinking, would it make sense to unify the two mappings?

```
    // mapping from fillHash to filler
    mapping(bytes32 => address) public fillers;
    // mapping from fillHash to validity flag
    mapping(bytes32 => bool) public invalidFillHashes;
```

Why not have just one `mapping(bytes32 => address)` where 
```
address == 0  => unresolved
address == 1  => invalidated
address == x  => resolved x as the filler
```
?

fredo: Additionally with the upcoming changes from #645 this will split up things even more because the invalidation happens on a fillHash basis but the filler mapping is on a request hash basis.

contracts/contracts/RequestManager.sol
```diff
@@ -331,6 +331,22 @@ contract RequestManager is Ownable {
...
+            request.depositReceiver == address(0) &&
+            claimReceiver == claim.claimer
+        ) {
+            withdrawDeposit(request, claim, claimReceiver);
+        }
+
+        return claimReceiver;
+    }
+
+    function resolveClaim(uint256 claimId) private returns (address) {
```
istankovic, line 347: I like the way you went with this. Can I suggest that we go one step further and make this function a `view` function?     
```
// Returns (deposit receiver, eth to transfer)
function resolveClaim(uint256 claimId) private view returns (address, uint256) {
```
That way, all the claim resolution logic would be contained in a single side-effect-free function, while the boring mechanics of sending ETH, emitting events etc. would be outside.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -714,22 +742,41 @@ def test_withdraw_without_challenge_with_resolution(
...
+
+    # Register a L1 resolution
     resolution_registry.resolveRequest(
-        request_id, fill_hash, web3.eth.chain_id, claimer.address, {"from": contracts.messenger2}
+        request_id, fill_hash, web3.eth.chain_id, l1_filler, {"from": contracts.messenger2}
     )
 
+    # Assert that correct filler is resolved, it reverts the false invalidation
+    if invalidate and l1_filler == claimer:
+        assert not resolution_registry.invalidFillHashes(fill_hash)
```
istankovic, line 761: It would be nice to see the entire boolean table covered, e.g.
```
assert resolution_registry.invalidFillHashes(fill_hash) == f(invalidate, claimer, l1_filler)
assert resolution_registry.filllers(fill_hash) == g(invalidate, claimer, l1_filler)
```

beamer/tests/contracts/test_request_manager.py
```diff
@@ -714,22 +742,41 @@ def test_withdraw_without_challenge_with_resolution(
...
+
     # The claim period is not over, but the resolution must allow withdrawal now
     withdraw_tx = request_manager.withdraw(claim_id, {"from": claimer})
-    assert "DepositWithdrawn" in withdraw_tx.events
-    assert "ClaimWithdrawn" in withdraw_tx.events
 
-    assert token.balanceOf(requester) == 0
-    assert token.balanceOf(claimer) == transfer_amount
+    if claimer == l1_filler:
+        assert "DepositWithdrawn" in withdraw_tx.events
```
istankovic, line 767: It would also be nice to test that this event is not emitted if the condition is not true.
In general, with parameterized tests such as this, I think it is a good idea to cover all branches.
To illustrate:
```
if condition:
    assert something
...
```
could be extended to cover also the case where `condition` is false:
```
assert (condition and something) or (not condition and not something)
...
```

istankovic (approved):
Looks good overall. We should consider
- `resolveClaim` conversion to a `view` function, per my comment
- better case coverage in tests
but those are nice-to-haves (and I'd really love to see them, either here or in another PR).

contracts/contracts/ResolutionRegistry.sol
```diff
@@ -21,6 +21,8 @@ contract ResolutionRegistry is CrossDomainRestrictedCalls {
     ) external restricted(resolutionChainId, msg.sender) {
         require(fillers[fillHash] == address(0), "Resolution already recorded");
         fillers[fillHash] = filler;
+        // Revert fill hash invalidation, fill proofs outweigh an invalidation
+        invalidFillHashes[fillHash] = false;
```
palango, line 25: Why do we need this? Resolution should work either way.

If we leave this in, this won't we idempotent any more, e.g. receiving the fill proof after the non-fill proof will look different than receiving the non-fill proof after the fill proof.

palango:
Needs some style fixer, but then LGTM.

PR #668, opened 2022-05-13
Frontend/feature/tab design

### LINKED ISSUE #657: frontend - implement tabs design
Checkout the [mock-ups](https://www.figma.com/file/Tx5V8SUrD1HrvDVqX7iEjn/Beamer-App?node-id=0%3A1) for more details.

Tasks for a single little PR:
  - [x] view component that defines "hole" for the content in the center of the screen
  - [x] view component defines a portal targets for action buttons
  - [x] view component is (for now) responsible for the wallet connection button (extract from dialog component)
  - [x] add new tabs component that takes full "hole" (parent size 100%)
  - [x] tab component has a left and right label property plus a left and right content
  - [x] tab content uses `<KeepAlive>` to don't unload
  - [x] `RequestDialog` gets put into the left tab slot
  - [x] `RequestDialog` teleports its `Transfer Funds` button to the action button target
  
This should be fairly small and is a good preparation for the next upcoming design changes (will be linked).

weilbith:
Closes #657

### DIFF
--- a/frontend/src/components/RequestDialog.vue
+++ b/frontend/src/components/RequestDialog.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="request-dialog px-20 pt-18 pb-16">
+  <div class="request-dialog">
     <FormKit
       ref="requestForm"
       v-slot="{ state: { valid } }"
@@ -55,12 +55,6 @@ import { useEthereumProvider } from '@/stores/ethereum-provider';
 import { RequestState } from '@/types/data';
 import type { RequestFormResult } from '@/types/form';
 
-interface Emits {
-  (e: 'reload'): void;
-}
-
-const emit = defineEmits<Emits>();
-
 const configuration = useConfiguration();
 const ethereumProvider = useEthereumProvider();
 const { provider, signer, chainId } = storeToRefs(ethereumProvider);
@@ -100,7 +94,11 @@ const submitRequestTransaction = async (formResult: RequestFormResult) => {
   );
 };
 
-watch(chainId, () => location.reload());
+watch(chainId, (_, oldChainId) => {
+  if (oldChainId !== -1) {
+    location.reload();
+  }
+});
 
 watch(transferError, async () => {
   if (transferError.value && requestState.value !== RequestState.RequestFailed) {
@@ -109,6 +107,6 @@ watch(transferError, async () => {
 });
 
 const newTransfer = () => {
-  emit('reload');
+  requestState.value = RequestState.Init;
 };
 </script>
--- a/frontend/src/components/RequestDialog.vue
+++ b/frontend/src/components/RequestDialog.vue
@@ -1,11 +1,5 @@
 <template>
-  <div class="request-dialog">
-    <div class="h-14">
-      <div v-if="ethereumProvider.signer" class="flex flex-row gap-4 justify-center items-center">
-        <div class="h-7 w-7 rounded-50 border-4 border-solid border-teal-light bg-green"></div>
-        <span class="text-lg">You are currently connected via Metamask</span>
-      </div>
-    </div>
+  <div class="request-dialog px-20 pt-18 pb-16">
     <FormKit
       ref="requestForm"
       v-slot="{ state: { valid } }"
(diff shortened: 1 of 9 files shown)


frontend/src/components/layout/Tabs.vue
```diff
@@ -0,0 +1,103 @@
...
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import type { Component } from 'vue';
+import { computed, onMounted, ref, shallowRef, watch } from 'vue';
+
+interface Props {
+  leftLabel: string;
```
nekhaly, line 36: While this fits exactly the current case we have, but in future if we decided to have more tabs in our app, this will be a very annoying part to update! Can we make this component more flexible to accept more tabs with a class definition for first and last tabs (or even detect them programatically if you need more fun :D)?! That we will more flexible IMO!

frontend/src/components/RequestDialog.vue
```diff
@@ -14,47 +8,36 @@
...
-      </div>
+      <RequestFormInputs v-if="requestState === RequestState.Init" />
+      <RequestStatus v-else :metadata="requestMetadata!" :state="requestState" />
+      <Transition name="expand">
+        <div v-if="transferError" class="mt-7 text-right text-lg text-orange-dark">
+          {{ transferError }}
+        </div>
+      </Transition>
+
+      <Teleport to="#action-button-portal">
```
nekhaly, line 19: I have red flags in my IDE for using Teleport and FormKit components without importing them in `script` part!

frontend/src/views/Home.vue
```diff
@@ -1,43 +1,85 @@
...
+
+      <div id="action-button-portal">
+        <FormKit
+          v-if="!signer"
+          input-class="w-112 bg-orange flex flex-row justify-center"
+          type="button"
+          @click="runRequestSigner"
+        >
+          <div v-if="requestSignerActive" class="h-8 w-8">
+            <spinner></spinner>
```
nekhaly, line 34: Same comment for `FormKit`, but here I think because we import it as `Spinner` with capital S! I also suggest to use a self-closing tag like this `<Spinner />`

nekhaly (changes requested):
Overall, it does the job perfectly, and my comment is for future mintainability! If we can do that from beginning it will be nice IMO to avoid hassle on `CHANGE` requests :D

frontend/src/views/Home.vue
weilbith: Did I said we need better linting? 😬

frontend/src/components/RequestDialog.vue
weilbith: If your IDE complains I'll fix that ofc. 😉

frontend/src/components/layout/Tabs.vue
weilbith: Yeaaah. Manuel and I thought about that and actively decided against it. But to live up to my own values, I probably should just do it.

frontend/src/components/RequestDialog.vue
weilbith: Okay, so `FormKit` is actually already imported... And if I import `Teleport` from `vue`, linting actually complains that it is unused and makes the CI fail.
Can you confirm that?

nekhaly: mmm.. I have tried to import it and the IDE linting had the same complaint! :thinking:  .. Not sure what to do actually so satisfy "Master Linting"! :)

frontend/src/components/layout/Tabs.vue
weilbith: Done. Please check if you are happy with the result and close the conversation if so.

frontend/src/components/RequestDialog.vue
nekhaly: OK.. all is good.. It's this stupid IDE which became insane recently!

frontend/src/components/RequestDialog.vue
```diff
@@ -132,19 +96,19 @@ const submitRequestTransaction = async (formResult: RequestFormResult) => {
   );
 };
 
-watch(chainId, () => location.reload());
-
-const shownError = computed(() => {
-  return requestSignerError.value || transferError.value;
+watch(chainId, (oldChainId) => {
+  if (oldChainId === -1) {
```
nekhaly, line 100: This causes the same issue I'm having in my WalletConnect PR at the moment and couldn't find a way to resolve yet! If you are trying to change between Boba and Metis after initial screen loading, page then is not reloading and no Token appears in select box!

weilbith: Good catch. The reason is very simple and stupid. The signature of a `WatchCallback` is `(newValue, oldValue) => void`. So I need to put it `(_, oldChainId)` and then `if (oldChainId !== -1)`. Pushing a fixup commit.

weilbith: Please confirm that it works for you.

nekhaly: Yup, it fixes a part of the issue, but not the issue itself. So as per our chat, we can figure out a solution for these doomed watchers!

weilbith: Okay, the problem is a bit more complex. On it...

weilbith: Alright. That should be fixed now by [f8b9eff](https://github.com/beamer-bridge/beamer/pull/668/commits/f8b9effa6002500173bcb79c640fe49db680f14a). Please check it out. 

I guess due to that it makes sense to merge this PR soon, so you can rebase your #647 on top of it to benefit from this fix too.

nekhaly: Amazing.. Works perfectly! Thank you!

Issue #570, opened 2022-04-13
Non-Fill: Agent submits Non-Fill when unfilled claim occured


fredo:
The agent should detect invalid fill hashes of unfilled requests. This means there are false claims to unfilled requests. If so the agent takes the request hash and fill id and invalidates it on the FillManager side.

It is essential that this happens before `claimPeriodExtension` minutes have passed after the claim started.

The agent should still respect `fill_wait_time` as the target chain might be slow.

Once the Non-Fill Proof was submitted by anybody the agent should acknowledge that and not send a Non-Fill Proof anymore. We probably have to reflect that in the claim state machine.

The agent should still call the Non-fill proof **EVEN** the claim was already challenged.

PR #672, opened 2022-05-16
Add handler for FillHashInvalidated events

palango:
Ref #570 

Broke this out of the issue as it it self contained and nicer to review like this.


beamer/models/claim.py
```diff
@@ -37,6 +38,7 @@ def __init__(
         | challenger_winning.to(claimer_winning)
         | ignored.to(ignored)
     )
+    invalidate = claimer_winning.to(invalidated) | challenger_winning.to(invalidated)
```
istankovic, line 41: Could you please also update the docs?

beamer/state_machine.py
```diff
@@ -241,9 +242,18 @@ def _handle_request_resolved(event: RequestResolved, context: Context) -> Handle
...
-    return False, None
+def _handle_fill_hash_invalidated(event: FillHashInvalidated, context: Context) -> HandlerResult:
+    request = context.requests.get(event.request_id)
+    if request is None:
+        return False, None
+
+    # Mark the claims with that fill_id as invalidated
+    for claim in context.claims:
+        fill_hash_of_claim = request.fill_hash_with_fill_id(claim.latest_claim_made.fill_id)
+        if claim.request_id == request.id and encode_hex(fill_hash_of_claim) == event.fill_hash:
```
istankovic, line 253: Nitpick: if the claim does not belong to this request, we could avoid even computing the fill hash (not that I expect it to be too expensive).

beamer/state_machine.py
```diff
@@ -241,9 +242,18 @@ def _handle_request_resolved(event: RequestResolved, context: Context) -> Handle
...
+            claim.invalidate()
+
+    return True, None
```
istankovic, line 256: We should return `True` here only if we invalidated any claims, right?

istankovic:
> Broke this out of the issue as it it self contained and nicer to review like this.

Thank you, that's much appreciated!

beamer/models/claim.py
palango: Yup.

beamer/tests/agent/unit/test_handlers.py
```diff
@@ -1,131 +1,25 @@
 from copy import deepcopy
-from pathlib import Path
```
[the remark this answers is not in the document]
palango: I don't mind, do you?

My hope is that we eventually don't even need to run ganache for those and they become even quicker.

beamer/state_machine.py
palango: Fixed

beamer/state_machine.py
fredo: It sounds that @istankovic 's question relates more to the fact if the agent or any other agent invalidated the claim. In that case, I'd say it shouldn't matter who invalidated.

istankovic: Hm, I see your concern, but on the other hand, my concern is that if we somehow get events out of order, we will drop the invalidate event and thus never be able to have the correct state for that claim.
So I would rather prefer not to drop such events.

Also, we probably have additional places/events where this might be of concern and I'd rather that we have a general solution, e.g. an expiration time for such events in the sense "drop the event if we are unable to proceed with it for more than N hours/minutes/etc.".

beamer/tests/agent/unit/test_handlers.py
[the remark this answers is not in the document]
istankovic: > I don't mind, do you?

Nah, I guess it just feels an extra level for no real benefit, at least right now. But maybe it will grow.

> My hope is that we eventually don't even need to run ganache for those and they become even quicker.

That would be nice, indeed.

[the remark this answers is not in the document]
palango: > Nah, I guess it just feels an extra level for no real benefit, at least right now. But maybe it will grow.

In that case I'd rather keep it and rather split the existing `test_handlers.py` so that it's more clear.

beamer/state_machine.py
fredo: Another question which comes to my mind in this discussion. If `FillHashInvalidated` is emitted by the `ResolutionRegistry`, are we also handling the `HashInvalidated` Event emitted by `FillManager`?

beamer/state_machine.py
```diff
@@ -241,9 +242,21 @@ def _handle_request_resolved(event: RequestResolved, context: Context) -> Handle
...
-    # TODO: implement once contract is finished
-    return False, None
+def _handle_fill_hash_invalidated(event: FillHashInvalidated, context: Context) -> HandlerResult:
+    request = context.requests.get(event.request_id)
+    if request is None:
+        return False, None
+
+    # Mark the claims with that fill_id as invalidated
+    for claim in context.claims:
+        if not claim.request_id == request.id:
```
istankovic, line 252: This should be:
```
         if claim.request_id != request.id:
```

beamer/state_machine.py
palango: Good point, for this case we don't need to L1 relay.

fredo: These two events should somehow be reflected in the state machine I guess. Because with `HashInvalidated` Event an agent would know that he doesn't need to call `invalidateFillHash` on the `FillManager` anymore.

beamer/models/claim.py
```diff
@@ -29,6 +29,8 @@ def __init__(
     claimer_winning = State("ClaimerWinning", initial=True)
     # Challenger is winning
     challenger_winning = State("ChallengerWinning")
+    invalidated = State("Invalidated")
+    invalidated_l1_resolved = State("L1Invalidated")
```
istankovic, line 33: Nitpick: so far we had the state identifiers consistent with state names, but here we're diverging from that practice.
I'd suggest
```suggestion
    invalidated_l1_resolved = State("InvalidatedL1Resolved")
```

beamer/tests/agent/unit/test_invalidate.py
```diff
@@ -0,0 +1,75 @@
...
+
+    request = make_request()
+    request.fill(config.account.address, b"", FILL_ID)
+    context.requests.add(request.id, request)
+    assert request.fill_id is not None
+
+    claim_1 = make_claim(request=request, claimer=config.account.address, fill_id=request.fill_id)
+    claim_2 = make_claim(
+        request=request,
+        claim_id=ClaimId(CLAIM_ID + 1),
```
istankovic, line 20: Nitpick: I'm unsure of the relation of `CLAIM_ID` and `claim_1.id` here. If they are supposed to be the same or related, then I'd suggest deriving the second claim ID from the first one, e.g. `ClaimId(claim_1.id + 1)` or similar.

beamer/tests/agent/unit/test_invalidate.py
```diff
@@ -0,0 +1,75 @@
...
+        tx_hash=HexBytes(""),
+        request_id=request.id,
+        fill_hash=fill_hash,
+    )
+    assert process_event(event, context) == (True, None)
+
+    # Check that a matching event invalidates the claim
+    assert claim_1.is_invalidated_l1_resolved  # pylint:disable=no-member
+
+    # Check that a unmatching event does not invalidate the claim
```
istankovic, line 39: Nitpick:
```suggestion
    # Check that a non-matching event does not invalidate the claim
```

beamer/tests/agent/unit/test_invalidate.py
```diff
@@ -0,0 +1,75 @@
...
+        request_hash=request.request_hash,
+        fill_id=request.fill_id,
+        fill_hash=fill_hash,
+    )
+    assert process_event(event, context) == (True, None)
+
+    # Check that a matching event invalidates the claim
+    assert claim_1.is_invalidated  # pylint:disable=no-member
+
+    # Check that a unmatching event does not invalidate the claim
```
istankovic, line 74: Nitpick:
```suggestion
    # Check that a non-matching event does not invalidate the claim
```

istankovic (approved):
Few nitpicks, looks good!

PR #675, opened 2022-05-17
agent: Invalidate invalid claims

### LINKED ISSUE #570: Non-Fill: Agent submits Non-Fill when unfilled claim occured

palango:
Resolves #570

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -47,7 +47,7 @@ jobs:
           path: |
             ~/.cache/pypoetry/virtualenvs
             ~/.solcx
-          key: poetry-${{ runner.os }}-${{ hashFiles('.github/workflows/ci.yml', 'poetry.lock') }}
+          key: poetry-1-${{ runner.os }}-${{ hashFiles('.github/workflows/ci.yml', 'poetry.lock') }}
 
       - name: Install ganache
         run: npm install ganache --global


beamer/chain.py
```diff
@@ -265,15 +265,29 @@ def process_claims(context: Context) -> None:
             to_remove.append(claim.id)
             continue
 
+        if claim.is_invalidated_l1_resolved:
+            # TODO: See https://github.com/beamer-bridge/beamer/issues/674
+            continue
+
+        if claim.is_invalidated:
```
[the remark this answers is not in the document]
istankovic: :woozy_face:

beamer/tests/agent/unit/test_invalidate.py
```diff
@@ -73,3 +76,72 @@ def test_handle_hash_invalidated():
 
     # Check that a non-matching event does not invalidate the claim
     assert claim_2.is_claimer_winning  # pylint:disable=no-member
+
+
+def test_maybe_invalidate_claim_wrong_fill_id():
+    context, config = make_context()
+
+    request = make_request(valid_until=int(time.time() * 2))
```
fredo, line 84: Is it necessary to make the request valid here? Would it change anything if we leave it with the default?

beamer/models/claim.py
```diff
@@ -39,7 +39,11 @@ def __init__(
         | challenger_winning.to(claimer_winning)
         | ignored.to(ignored)
     )
-    invalidate = claimer_winning.to(invalidated) | challenger_winning.to(invalidated)
+    invalidate = (
+        claimer_winning.to(invalidated)
+        | challenger_winning.to(invalidated)
+        | invalidated.to(invalidated)
```
fredo, line 45: when is this transition triggered?

palango: For example when we trigger the invalidation and then receive the event later.

beamer/chain.py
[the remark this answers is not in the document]
palango: Just discusses this with Fred. We think it makes most sense to prepend a new state before the challenge game and do the invalidation there if necessary.

fredo:
One important note, I just realized. When you change the state machine, please make sure that the following case can occur. 

We have an unfilled request, but it got challenged by someone. Still, nobody made the invalidation. This happens exactly in the `self-challenge`  case. One solution could be to not process the event if we are still in the `fill_wait_time`. And once the `fill_wait_time` is passed, we process the claim once more in the `started` state (to trigger the invalidation) and then process the challenge state. In this solution we would need to make sure to not challenge it in the same cycle. 
Let's discuss in private if you have further questions

beamer/tests/agent/unit/test_invalidate.py
palango: The default is currently some fixed number that's in the past, so we need this here.

beamer/chain.py
```diff
@@ -414,6 +428,24 @@ def maybe_challenge(claim: Claim, context: Context) -> bool:
...
+    # As per definition an invalid or expired request cannot be claimed
+    # This gives us a chronological order. The agent should never garbage collect
+    # a request which has active claims
+    assert request is not None, "Active claim for non-existent request"
+
+    timestamp = int(time.time())
+    if timestamp < claim.challenge_back_off_timestamp:
+        return
+
+    if timestamp > request.valid_until:
```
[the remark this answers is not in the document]
palango: Good point, I guess we can completely remove this.

beamer/chain.py
```diff
@@ -150,7 +150,7 @@ def add_events(self, events: list[Event]) -> None:
     def _thread_func(self) -> None:
         self._log.info("EventProcessor started")
         while not self._stop:
-            if self._have_new_events.wait(1):
+            if self._have_new_events.wait(1) or len(self._events) > 0:
```
fredo, line 153: Is this correct? Couldn't it be the case that there are still events in the queue which cannot be processed? Wouldn't we in that case never wait anymore?

palango: In my understanding this will first wait, and then, if there are still events in the queue, process them, even if no new events came in.

beamer/chain.py
```diff
@@ -473,3 +504,16 @@ def _withdraw(claim: Claim, context: Context) -> None:
 
     claim.transaction_pending = True
     log.debug("Withdrew", claim=claim.id, txn_hash=txn_hash.hex())
+
+
+def _invalidate(request: Request, claim: Claim, context: Context) -> None:
+    func = context.fill_manager.functions.invalidateFillHash(
+        request.id, request.request_hash, claim.latest_claim_made.fill_id, request.source_chain_id
```
fredo, line 511: building on the latest main, request id is not a parameter of `invalidateFillHash`anymore

beamer/models/claim.py
```diff
@@ -25,25 +25,23 @@ def __init__(
...
-    claimer_winning = State("ClaimerWinning", initial=True)
+    claimer_winning = State("ClaimerWinning")
     # Challenger is winning
     challenger_winning = State("ChallengerWinning")
-    invalidated = State("Invalidated")
     invalidated_l1_resolved = State("InvalidatedL1Resolved")
     ignored = State("Ignored")
     withdrawn = State("Withdrawn")
 
+    start_challenge = started.to(claimer_winning) | claimer_winning.to(claimer_winning)
```
fredo, line 37: out of curiosity, in which case do we call `start_challenge()` on a claimer_winning state?

fredo: Ah, when we received a fillHashInvalid event from the fill manager, right?

beamer/state_machine.py
```diff
@@ -262,8 +265,13 @@ def _handle_request_resolved(event: RequestResolved, context: Context) -> Handle
 def _handle_hash_invalidated(event: HashInvalidated, context: Context) -> HandlerResult:
     claim = _find_claim_by_fill_hash(context, event.fill_hash)
     if claim is not None:
-        claim.invalidate()
-        return True, None
+        request = context.requests.get(claim.request_id)
+        assert request is not None, "Request missing for claim"
+
+        fill_hash_of_claim = request.fill_hash_with_fill_id(claim.latest_claim_made.fill_id)
+        if fill_hash_of_claim == event.fill_hash and claim.is_started:
```
fredo, line 272: this is already covered in `_find_claim_by_fill_hash`

beamer/chain.py
palango: This would look like this:
```diff
diff --git a/beamer/chain.py b/beamer/chain.py
index 46fc26d..cd9d43a 100644
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -150,7 +150,7 @@ class EventProcessor:
     def _thread_func(self) -> None:
         self._log.info("EventProcessor started")
         while not self._stop:
-            if self._have_new_events.wait(1) or len(self._events) > 0:
+            if self._have_new_events.wait(1):
                 self._have_new_events.clear()
                 self._process_events()

@@ -195,6 +195,8 @@ class EventProcessor:
             )
             iteration += 1
             if not any_state_changed:
+                if len(unprocessed) > 0:
+                    self._have_new_events.set()
                 break

         # New events might be created by event handlers. If they would be

File: beamer/tests/agent/unit/test_invalidate.py [hunk 9]
```diff
@@ -77,7 +79,76 @@ def test_handle_hash_invalidated():
     assert process_event(event, context) == (True, None)
 
     # Check that a matching event invalidates the claim
```
fredo, line 81: a different comment would be nicer. Semantically we do not see that the claim is invalidated anymore as no invalidated state exists anymore.

fredo: btw I realise that we need to store some transaction hash same as goes for the fill in order to call l1 eventually. But we can add that in the respective issue.

fredo: But with that tx_hash we could do the assert correctly, checking that it was invalidated.

beamer/state_machine.py
palango: Thanks, overlooked this.

beamer/state_machine.py
```diff
@@ -262,8 +265,13 @@ def _handle_request_resolved(event: RequestResolved, context: Context) -> Handle
...
+            claim.start_challenge()
+            return True, None
```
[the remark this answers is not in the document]
palango: We also need to handle the case where the claim is in challenged state already, then this can be removed.

beamer/chain.py
palango: @fredo What's your favorite?

beamer/tests/agent/unit/test_invalidate.py
```diff
@@ -77,7 +79,76 @@ def test_handle_hash_invalidated():
...
-    assert claim_1.is_invalidated  # pylint:disable=no-member
+    assert claim.is_claimer_winning  # pylint:disable=no-member
 
-    # Check that a non-matching event does not invalidate the claim
-    assert claim_2.is_claimer_winning  # pylint:disable=no-member
+
+@patch("beamer.chain._invalidate")
+def test_maybe_invalidate_claim_wrong_fill_id(mocked_invalidate):
```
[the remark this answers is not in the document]
palango: Added

docs/source/claim_state_machine.dot
```diff
@@ -1,16 +1,15 @@
 digraph claim_state_machine {
-    node [] claimer_winning, challenger_winning, invalidated, invalidated_l1_resolved, ignored, withdrawn;
+    node [] started, claimer_winning, challenger_winning, invalidated_l1_resolved, ignored, withdrawn;
+
+    started -> claimer_winning [label = start_challenge];
+    started -> challenger_winning [label = start_challenge];
```
fredo, line 5: I think this is not correct. `start_challenge()` always transitions from started to claimer_winning

fredo (commented):
One last comment, otherwise looks good! thanks for the effort!

docs/source/claim_state_machine.dot
palango: Totally.

palango: Fixed.

fredo (approved):

PR #678, opened 2022-05-18
Add an account generation script

istankovic:
This is useful to have and will be used in the upcoming e2e scripts, so to get this out of the way...

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -32,10 +32,8 @@ Start ganache::
 
 Create a JSON keyfile corresponding to one of the accounts pre-funded by ganache::
 
-    python -c "import eth_account, json;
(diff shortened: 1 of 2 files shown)


palango:
When would we use this? I wonder because this account will not be funded unless the account was known beforehand, and then most likely there's an account file as well.

istankovic:
> 



> When would we use this? I wonder because this account will not be funded unless the account was known beforehand, and then most likely there's an account file as well.

There is a "funnel" account included in the Arbitrum setup, however, I would rather avoid using that since I'm not 100% sure about everything it is currently being used for. So I would prefer to create a new account each time we run an E2E test, fund that, and proceed with it.

Also, once we consolidate the Optimism and Arbitrum setups to use the same chain, I think we may want to use "our" account for testing purposes, rather than relying on whatever account both of them provide. So it's mainly about decoupling and clarity.

palango (approved):

PR #679, opened 2022-05-18
tests: Use `alloc_accounts` everywhere

### LINKED ISSUE #646: Use alloc_accounts in all tests
Test refactoring to use alloc_accounts in all tests

palango:
Needed a little filler issue without the need to think.

Resolves #646

### DIFF
--- a/beamer/tests/contracts/test_request_manager.py
+++ b/beamer/tests/contracts/test_request_manager.py
@@ -79,12 +79,11 @@ def test_claim_with_different_stakes(token, request_manager, claim_stake):
 
 def test_claim_challenge(request_manager, token, claim_stake):
     """Test challenging a claim"""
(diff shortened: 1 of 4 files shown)


beamer/tests/contracts/test_request_manager.py
```diff
@@ -79,12 +79,11 @@ def test_claim_with_different_stakes(token, request_manager, claim_stake):
 
 def test_claim_challenge(request_manager, token, claim_stake):
     """Test challenging a claim"""
-    (requester,) = alloc_accounts(1)
+    (requester, challenger) = alloc_accounts(2)
```
istankovic, line 82: Nitpick: I believe mypy (or whatever was previously complaining) will not complain if we omit parentheses here.

istankovic (approved):
Nice! :heart:

PR #682, opened 2022-05-19
frontend: Improve types of the properties of Transfer

manuelwedler:
This adds a sourceAmount and targetAmount (type: TokenAmount) to Transfer and removes the sourceToken, the targetToken and the amount.

The tokens can now be accessed from the amounts. This also enables later having a different target amount from the source one.
Also it makes the fees a TokenAmount because the fee model changes and we do not want to break backwards compatibility with encoding in the local storage.

For convenience an EthereumAmount class is created which inherits from the TokenAmount and sets the right symbol when created.  This type is also now used in the useRequestFee composable. Later, the composable will need to be adapted to the new fee model and use a TokenAmount.

### DIFF
--- a/frontend/src/types/data.ts
+++ b/frontend/src/types/data.ts
@@ -16,3 +16,9 @@ export type Token = {
   symbol: string;
   decimals: number;
 };
+
+export const ETH: Token = {
+  address: '',
+  symbol: 'ETH',
+  decimals: 18,
+};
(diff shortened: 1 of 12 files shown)


frontend/src/types/data.ts
```diff
@@ -16,3 +16,9 @@ export type Token = {
   symbol: string;
   decimals: number;
 };
+
+export const ETH: Token = {
```
weilbith, line 20: I doubt this a good place for this, but I don't care too much right now. Just a nitpick...

weilbith (approved):
Great work. Really good. Exactly how we both planned it. 👍🏾 👌🏾

weilbith:
As Manuel is working next time in 5 days, we need this PR, and I approved it, I go ahead and merge it.

Issue #571, opened 2022-04-14
Non-Fill: Agent engages in ongoing challenges


fredo:
#569 introduces the possibility to join an ongoing challenge at any time if the claimer is winning. 

While this opens a whole new set of possible strategies for an agent, I suggest to make only necessary changes to the current implementation of the agent. The most important point and one of the bigger reasons why we implemented multiparty bidding is to prevent an attack on unfilled requests.
That's why I suggest that the agent only tries to challenge an ongoing challenge if the underlying request is not filled.

So what changes in the agent with this issue is: 

1) check whether underlying request is filled
2) check if claimer is winning
3) try to challenge with at least cost of L1

This implicitly means that a claim should only go to the ignored state if the underlying request is filled. Even if the claim is challenged and not by the agent, we should not transition to ignored as long as the request is not filled.

fredo:
Outcome of the discussion with @palango @istankovic :

- Multiparty bidding is essential in self-challenges of unfilled requests
- Multiparty bidding is optional in all other ongoing challenges

We will keep it simple and only implement the strategy to step in ongoing challenges to secure the protocol.

### Conditions for stepping in

- A claim exists for an unfilled request
- fill_wait_time is passed and still unfilled
- The claim is already challenged
- The claimer is leading

Independently of this, the fill hash should be invalidated on target rollup.

The agent will challenge with enough stake to cover L1 cost.

PR #686, opened 2022-05-24
Agent: engage in ongoing challenges only at unfilled requests

fredo:
fixes: #571 

tests are missing yet

### DIFF
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -398,11 +398,13 @@ def maybe_challenge(claim: Claim, context: Context) -> bool:
             return False
 
     initial_claim_stake = context.request_manager.functions.claimStake().call()
-    minimum_stake = claim.get_minimum_challenge_stake(initial_claim_stake)
+    stake = claim.get_minimum_challenge_stake(initial_claim_stake)
 
     # TODO: have a central variable and proper L1 cost calculations
     l1_cost = Wei(initial_claim_stake + 10**15)
-    stake = max(minimum_stake, Wei(l1_cost - own_challenge_stake))
+
+    if claim.latest_claim_made.challenger_stake_total > 0:
+        stake = max(stake, Wei(l1_cost - own_challenge_stake))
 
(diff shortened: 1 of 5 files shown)


beamer/chain.py
```diff
@@ -382,19 +382,29 @@ def maybe_challenge(claim: Claim, context: Context) -> bool:
...
 
-    if claim.is_challenger_winning and claim.claimer != context.address:
-        return False
-
     agent_winning = context.address in claim.get_winning_addresses()
     if agent_winning:
         return False
 
+    own_challenge_stake = Wei(claim.get_challenger_stake(context.address))
+    agent_losing = claim.claimer == context.address or own_challenge_stake > 0
```
palango, line 390: maybe we can find a better name for the variable

palango: I think `agent_participating` fits better

palango: Yes, I just feel this makes it harder to read the condition below. But it's just a nitpick, both are fine.

beamer/tests/agent/unit/test_challenge.py
```diff
@@ -0,0 +1,83 @@
...
+def test_challenge_as_claimer():
+    context, _ = make_context()
+
+    request = make_request()
+    context.requests.add(request.id, request)
+    request.filler = make_address()
+    claim = make_claim(
+        request=request,
+        claimer=context.address,
+        challenger=make_address(),
```
palango, line 48: you don't strictly need this, it will be set automatically.

palango (approved):
Looks good, and nice tests!

beamer/tests/agent/unit/test_challenge.py
fredo: for claimer you are right, but I think for challenger it's not the case (can be None)

fredo: yeah just came across that .. I think, I will change that in order to be able to create an unchallenged claim.

PR #688, opened 2022-05-25
Reorganize contracts

### LINKED ISSUE #662: Create messenger proxy contract for sending L1 -> L2 messages
In order to adapt the resolver to work with arbitrum we will introduce a messenger proxy contract for each roll up architecture. 

Currently the message flow is `....sender L2 outbox -> resolver -> receiver L2 inbox`. This will change to `....sender L2 outbox -> resolver -> messenger proxy -> receiver L2 inbox`.

- Create an Interface
- Create a proxy contract to work for optimism contracts
- adapt the resolver to this
- Adapt deployment scripts for messenger contracts

istankovic:
Part of the Arbitrum integration work, this PR prepares the way for Arbitrum-specific contracts by introducing a new contract organization. The new directory structure:
```
contracts/
├── BeamerUtils.sol
├── chains
│   ├── optimism
│   │   ├── ICrossDomainMessenger.sol
│   │   ├── Lib_PredeployAddresses.sol
│   │   ├── OptimismMessengers.sol
│   │   └── OptimismProofSubmitter.sol
│   └── test
│       ├── TestMessengers.sol
│       └── TestProofSubmitter.sol
├── CrossDomainRestrictedCalls.sol
├── FillManager.sol
├── MintableToken.sol
├── RequestManager.sol
├── ResolutionRegistry.sol
├── Resolver.sol
├── RestrictedCalls.sol
└── StandardToken.sol
```

image

Red lines indicate restricted calls, red dashed lines indicate restricted calls that cross the L1/L2 boundary.
Black lines represent ordinary, unrestricted calls.
With the new scheme we would have 3 rollup-specific contracts: `L1Messenger` (deployed on L1), `L2Messenger` (deployed on L2) and `ProofSubmitter` (deployed on L2).
Everything else would be rollup-independent.

Closes #662.


contracts/contracts/chains/optimism/OptimismMessengers.sol
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
+
+contract OptimismL1Messenger is OptimismMessengerBase {
+    constructor(address messenger_) {
+        messenger = ICrossDomainMessenger(messenger_);
+    }
+}
+
+contract OptimismL2Messenger is OptimismMessengerBase {
```
fredo, line 32: Why do we need the distinction between L1 and L2 messenger? I thought you wrote it in a way that it is equal. Is it just to have the predeployed address?

scripts/deployment/rinkeby_testnets.json
```diff
@@ -9,14 +9,16 @@
...
+            "l2_to_l1_messenger": "0xF10EEfC14eB5b7885Ea9F7A631a21c7a82cf5D76",
             "finalization_time": 604800
         },
         {
             "name": "metis testnet",
             "rpc": "https://stardust.metis.io/?owner=588",
             "chain_id": 588,
-            "messenger_contract_address": "0xBa6f1aAF3F909F208b224F3985eCda745FA9101A",
+            "proof_submitter": "OptimismProofSubmitter",
+            "l2_to_l1_messenger": "0xBa6f1aAF3F909F208b224F3985eCda745FA9101A",
```
fredo, line 21: here is some inconsistent naming. Anywhere else the term `l2_messenger` or `l1_messenger` is used. I think consistency would be nicer.

.gitmodules
```diff
@@ -1,5 +1,5 @@
 [submodule "docker/optimism"]
-	path = docker/optimism
+	path = docker/optimism/optimism
```
palango, line 2: why change this?

.solhintignore
```diff
@@ -0,0 +1,2 @@
+contracts/contracts/chains/optimism/ICrossDomainMessenger.sol
+contracts/contracts/chains/optimism/Lib_PredeployAddresses.sol
```
palango, line 2: Why do we ignore those?

contracts/contracts/chains/optimism/Lib_PredeployAddresses.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+/**
+ * @title Lib_PredeployAddresses
+ */
+library Lib_PredeployAddresses {
```
palango, line 7: Do we expect this to change (often) in the future?

docs/source/contracts.dot
```diff
@@ -0,0 +1,50 @@
+digraph "Beamer contracts" {
```
fredo, line 1: Can you maybe also mark chain-independent contracts and chain dependent contracts with different colors?

In general it would be nice to have clearly explained somewhere (in one or two sentences) that there are chain dependent contracts to work with their specific messengers (or more abstract contract interfaces) and chain independent ones which are purely dependent to our protocol. Basically a short description why we need to distinguish.

docker/optimism/optimism.sh
```diff
@@ -1,61 +1,70 @@
 #!/usr/bin/env bash
-ROOT=$(dirname $(dirname "$(realpath "$0")"))
-DOCKER_COMPOSE_FILES="-f ${ROOT}/beamer/docker/optimism/ops/docker-compose.yml -f ${ROOT}/beamer/docker/optimism/ops/docker-compose-nobuild.yml"
+. "$(realpath $(dirname $0))/../scripts/common.sh"
```
fredo, line 2: very nice composition of tasks :)

.gitmodules
istankovic: > why change this?

This makes it consistent with the arbitrum stuff and also leaves the room for optimism-specific scripts in `docker/optimism`, like the one we already have, `optimism.sh`.

I also thought about moving all that stuff into e.g. `${ROOT}/integrations/` or `${ROOT}/rollups/` or something similar, because it does not have anything to do with our docker setup, as the toplevel dir name may suggest.

contracts/contracts/chains/optimism/OptimismMessengers.sol
istankovic: For the Optimism case, both are almost identicall, but as we go on adding support for other rollups I suspect we might run into more differences. So I thought it was good to leave some space for that possibility and to clearly indicate which messengers belong to which chain.

scripts/deployment/rinkeby_testnets.json
istankovic: Good catch! That's a leftover from a previous iteration of the patches and is clearly wrong.

contracts/contracts/chains/optimism/Lib_PredeployAddresses.sol
istankovic: I guess the values here will not change anytime soon, but I nevertheless wanted to have meaningful, searchable names instead of simply a long constant. I think this way makes it more clear where the value came from and would make it easier to search for in the future.

docker/optimism/optimism.sh
istankovic: Eh... bash... :-)

docs/source/contracts.dot
istankovic: I totally agree and I planned to write a bit more, but my time was eaten away by... more pressing concerns. :sweat_smile: 
Now that the Optimism E2E test is working, I can write some more.

istankovic: Done. Chain-dependent contracts are now colored differently. Also added tables listing all the contracts and the chains they are deployed to. Finally, some wording has been improved and a few sentences added to make things more clear.

scripts/deployment/rinkeby_testnets.json
istankovic: Fixed.

contracts/contracts/chains/test/TestMessengers.sol
```diff
@@ -0,0 +1,43 @@
...
+    function setForwardState(bool forward) external {
+        forwardMessages = forward;
+    }
+
+    function nativeMessenger() external view returns (address) {
+        return address(this);
+    }
+}
+
+contract TestL1Messenger is TestMessengerBase {}
```
palango, line 41: Can you suppress those warning below_

docker/optimism/optimism.sh
```diff
@@ -0,0 +1,101 @@
...
+
+down() {
+    echo "Shutting down the end-to-end environment"
+    make -C ${OPTIMISM}/ops down
+}
+
+up() {
+    echo "Starting the end-to-end environment"
+    docker_compose_file="-f ${OPTIMISM}/ops/docker-compose.yml"
+    docker compose ${docker_compose_file} up --scale relayer=1 -d
```
palango, line 27: If this doesn't work, why have `--scale relayer=1` at all?

scripts/deployment/main.py
```diff
@@ -67,7 +68,18 @@ def contract_to_json(contracts: dict[str, DeployedContract]) -> dict:
     }
 
 
-def deploy_contract(web3: Web3, name: str, *args: Any) -> DeployedContract:
+def deploy_contract(web3: Web3, constructor_spec: Union[str, Sequence]) -> DeployedContract:
+    # constructor_spec is either
```
palango, line 72: If this is so, then the type annotation is wrong.

scripts/deployment/main.py
```diff
@@ -196,56 +209,39 @@ def main(
     password: str,
     output_dir: Path,
     config_file: Path,
-    test_messenger: bool,
     allow_same_chain: bool,
 ) -> None:
-    commit_id = get_commit_id()
     account = account_from_keyfile(keystore_file, password)
+    print("Deployer:", account.address)
```
palango, line 215: This should probably go away

contracts/contracts/chains/test/TestMessengers.sol
istankovic: Fixed.

docker/optimism/optimism.sh
istankovic: Good point, removed.

scripts/deployment/main.py
istankovic: > If this is so, then the type annotation can be improved

I'm not sure what you mean, could you please expand?
If you meant something like

```python
constructor_spec: Union[str, Sequence[str]]
```

that would unfortunately not work because there are types other than `str` that may appear there.
Maybe something like `Union[str, Sequence[Union[str, int, ChecksumAddress]]]` might work, but I haven't tried...

scripts/deployment/main.py
istankovic: Actually, I added this intentionally since it proved useful when I was inspecting the output of the deployment script and correlating that with the E2E script. It may come in handy so I'd prefer to leave it in, but if you feel strongly about it, I can remove it.

palango (approved):

istankovic:
@palango could you please take a look at the 3 newly added commits? I've added these because I noticed the deployment script was flaky on CI -- we need to wait for the tx receipt. For that I reused the existing code in the agent, just moved it to `beamer.util` and made it public.

PR #694, opened 2022-05-29
frontend: enforce type imports

weilbith:
Since TypeScript 3.8 it is a good practice to use type imports (`import type ...`) where possible. Type imports allow for certain compiler and tooling optimizations.

This rule is auto-fixable. So it doesn't add extra work to the developers. Except the ESLint rule adaption, all changes in this commit are caused by automated fixes.

Note that this got applied to the `relayer` too.


weilbith:
Just a nice to have. Nothing critical with priority to become reviewed. Critics are welcome.

palango (approved):
Makes sense to me.

PR #697, opened 2022-05-31
Remove request ID in L1 contract events

fredo:
The agent must find the corresponding request by the fill hash.

For simplicity I chose iterating over claims and reconstructing the fill hash to find the corresponding requests / claims.

This is only a temporary change and will be changed soon by #645 changes in the resolution registry. In that PR we can optionally introduce optimizations on the agent to find the corresponding objects more efficiently.

### DIFF
--- a/beamer/state_machine.py
+++ b/beamer/state_machine.py
@@ -32,7 +32,7 @@ from beamer.l1_resolution import run_relayer
 from beamer.models.claim import Claim
 from beamer.models.request import Request
 from beamer.tracker import Tracker
-from beamer.typing import ChainId, ClaimId, RequestId
+from beamer.typing import ChainId, ClaimId, FillHash, RequestId
 from beamer.util import TokenMatchChecker
 
 log = structlog.get_logger(__name__)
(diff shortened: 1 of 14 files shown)


beamer/state_machine.py
```diff
@@ -93,6 +93,17 @@ def process_event(event: Event, context: Context) -> HandlerResult:
         raise RuntimeError("Unrecognized event type")
 
 
+def _find_claim_by_fill_hash(context: Context, fill_hash: FillHash) -> Optional[Claim]:
+    for claim in context.claims:
+        request = context.requests.get(claim.request_id)
+        assert request is not None
```
istankovic, line 99: Can we encounter claims for non-existent requests?
Or do we discard those immediately?

istankovic (approved):
Looks fine.

beamer/state_machine.py
fredo: I think this must not happen. As claims and requests come from the same chain they arrive chronologically. The contracts forbid creating a claim for an invalid, expired or withdrawn request (id).
The request object in the agent should never be discarded before expiration or withdrawal. Thus, if a claim event arrives there must be a corresponding request object (otherwise it's a bug)

PR #699, opened 2022-05-31
frontend: display error messages in steps

weilbith:
Currently we still use `daisyUI` for the progress steps. Having this, it wasn't easy to have multiline content in the step label with different colors. The workaround with an absolute `<div>`, conditional padding etc. is not optimal. As it is only a visual component, this is okay. In the long term we should replace this with a custom implementation that can be tailed 100% to our needs.


manuelwedler (approved):
Looks good to me! Thanks! I also hope that we can get rid of daisyui soon.

PR #703, opened 2022-06-02
frontend: Add typescript compiler check to linting

manuelwedler:
I was a bit annoyed that my IDE showed typescript errors which were not caught by our linting and the CI, so I added tsc to the linting script.


weilbith (approved):
Thank you! 🙏🏾 
This was annoying me too and I think somewhere I also have a branch adding this. 😆 
Next we need a PR to add these checks for Vue SFC and especially the template section too. Still have some errors there in my IDE.

PR #706, opened 2022-06-02
Trigger L1 invalidation

### LINKED ISSUE #669: Non-Fill: Trigger L1 relaying
Currently the agent triggers L1 resolution only for fill proofs, but not the non-fill proof.

Depends on sending the non-fill proof in #570

## Tasks
- [ ] Decide when to trigger non-fill L1 relaying
- [ ] Add new `InitiateL1InvalidationEvent` or generalize current event
- [ ] Trigger event when non-fill proff should be relayed using L1
- [ ] Add unit tests validating the above

palango:
Resolves #669 

- First commit fixes an oversight on the return type of the `_l1_resolution_criteria_fulfilled` function
- Second commit add the `InitiateL1InvalidationEvent`, which is triggered when a claim that has been invalidated can be L1 resolved.

### DIFF
--- a/beamer/state_machine.py
+++ b/beamer/state_machine.py
@@ -22,13 +22,14 @@ from beamer.events import (
     Event,
     FillHashInvalidated,
     HashInvalidated,
+    InitiateL1InvalidationEvent,
     InitiateL1ResolutionEvent,
     LatestBlockUpdatedEvent,
     RequestCreated,
     RequestFilled,
     RequestResolved,
 )
-from beamer.l1_resolution import run_relayer
+from beamer.l1_resolution import run_relayer_for_tx
 from beamer.models.claim import Claim
 from beamer.models.request import Request
 from beamer.tracker import Tracker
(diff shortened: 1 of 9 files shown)


beamer/state_machine.py
```diff
@@ -217,22 +221,30 @@ def _handle_claim_made(event: ClaimMade, context: Context) -> HandlerResult:
     if claim.is_started:
         return False, None
 
+    events: list[Event] = []
+    if not claim.is_invalidated_l1_resolved and claim.invalidation_tx is not None:
+        events.append(
+            InitiateL1InvalidationEvent(
```
fredo, line 227: Wouldn't it be possible to check here for L1 criteria fulfilled? In that way we wouldn't create an Event every iteration and then drop it again. 
This also counts for L1 resolution

fredo: @palango What are your thoughts on that? We can also outsource it into its own PR as it was like that already for L1 resolution

palango: Yeah, let's do that in another PR. The problem is that we need to builds the infrastructure to emit events from the requests and claim handlers for this.

fredo (approved):

PR #707, opened 2022-06-02
Also store fill id in request when withdrawn to correctly resolve other active claims

fredo:
There was a bug in the contracts code when a claimer had multiple claims on the same request.

If one of them was successfully withdrawn, all other claims would be resolved in favor of the claimer which could be wrong if the fill id was different.

The rule is: 

All identical claims should be resolved equally. A correct claim is only correct if the claimer is the filler and the fill id matches. If the correct filler creates a claim with a wrong fill id, the claim becomes invalid.

This must be reflected in the withdraw function: In order to achieve that we need to store all relevant information with the request upon deposit withdraw, so that future claims can be resolved correctly. 
That means if the future claim is from the same address who withdrew the deposit AND the fill id is the same as from the claim which was used when withdrawing the deposit, then the claimer is entitled to receive the claim stake. In all other cases the challenger wins.

was found and is partially needed for #645 

Please note that the `FillerInfo` object will be used in an upcoming PR in the resolution registry as well

### DIFF
--- a/contracts/contracts/BeamerUtils.sol
+++ b/contracts/contracts/BeamerUtils.sol
@@ -2,6 +2,11 @@
 pragma solidity ^0.8.12;
 
 library BeamerUtils {
+    struct FillInfo {
+        address filler;
+        bytes32 fillId;
+    }
+
     function createRequestHash(
         uint256 requestId,
         uint256 sourceChainId,
--- a/contracts/contracts/RequestManager.sol
+++ b/contracts/contracts/RequestManager.sol
@@ -23,7 +23,7 @@ contract RequestManager is Ownable {
         address targetTokenAddress;
         address targetAddress;
         uint256 amount;
-        address depositReceiver;
+        BeamerUtils.FillInfo withdrawInfo;
         uint192 activeClaims;
         uint256 validUntil;
         uint256 lpFee;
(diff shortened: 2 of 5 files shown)


contracts/contracts/BeamerUtils.sol
```diff
@@ -2,6 +2,11 @@
 pragma solidity ^0.8.12;
 
 library BeamerUtils {
+    struct FillerInfo {
```
palango, line 5: Should we maybe call this `FillInfo`?

contracts/contracts/RequestManager.sol
```diff
@@ -175,7 +175,7 @@ contract RequestManager is Ownable {
         newRequest.targetTokenAddress = targetTokenAddress;
         newRequest.targetAddress = targetAddress;
         newRequest.amount = amount;
-        newRequest.depositReceiver = address(0);
+        newRequest.withdrawInfo = BeamerUtils.FillerInfo(address(0), "0x0");
```
palango, line 178: I think `bytes32(0)` is clearer.

fredo: Okay, I wasn't sure what'S the most obvious way.

contracts/contracts/BeamerUtils.sol
fredo: agreed

contracts/contracts/RequestManager.sol
```diff
@@ -211,7 +211,7 @@ contract RequestManager is Ownable {
         );
         require(request.activeClaims == 0, "Active claims running");
 
-        request.depositReceiver = request.sender;
+        request.withdrawInfo = BeamerUtils.FillInfo(request.sender, bytes32(0));
```
palango, line 214: Wouldn't it be enough to just set the filler here?

palango (approved):
LGTM

contracts/contracts/RequestManager.sol
fredo: totally right

PR #710, opened 2022-06-03
frontend: fix issues on wallet connect import

weilbith:
When we worked on the WalletConnect integration, we had some weird issues with importing the dependency. While TypeScript is fine, the application does no more work at all in the browser. Therefore we uses a more specific import of a specific file of the package. That import causes issues with TypeScript due to a missing declaration file.
This issue was solved in the past and working, because we had no TypeScript linting enabled. Recently we enabled the latter and this import got simply fixed, without checking the functionality of the application.
This commit is now recovering the ugly fix of the past and makes the linting pass.


weilbith:
Due to the current personnel, I'm allowed to merge such blocking, but feasibly safe,l PRs on my own. Therefore I'm using my administration permissions and merge this PR without a review.

PR #713, opened 2022-06-04
frontend: add waiting dots for active steps

weilbith:
Little weekend PR to implement a little UI feature for our designer. 😉 

A new component called `WaitingDots` got added. This component does an animation of three dots that appear after each other ('' -> '.' -> '..' -> '...' -> '' -> '.' -> ...).
It gets used as an indicator that a `ProgressStep` is currently active (and waiting or doing something).

This includes a missing test case for error messages in progress steps.

PR #714, opened 2022-06-04
Frontend/feature/custom progress dots

weilbith:
Just another weekend PR to improve design and UI handling.
Based on top of #713 to reduce merge conflicts and potential mistakes.

The DaisyUI dependency can be a pita and we want to get rid of it. Therefore we need to add custom implementations for the occurrences where we use it. One of these places were the `Progress` and `ProgressStep` components. They used the classes of `steps*` to get the list with circles and lines between. Recently, when we added error message to steps, it was not possible to create dynamic sizing of the steps. Furthermore we had to trick around with absolute placement, which lead to accidentally cut text etc. pp. Everything but not optimal. At this point a dependency is no more helpful but painful.
This commit provides a custom implementation of these steps. The advantage: it is much less code than before and 100% customizable to our needs. So got the previously mentioned problems already solved. Any future request by the designer
can be easily addressed too.

Note that this commit merges two commits into one. First of all, it could be argued that the former split was simply overload. Second it is for the reason that it makes no sense to have all this special styling of the list items in an individual list item component. This only makes list for a whole list and therefore binding this together makes sense. Less code, less tests, better
readability, 100% customizability.


weilbith:
#713 got merged. Resolving conflicts now.

weilbith:
**Disclaimer:**
In result of the current personnel shortage, management has decided to empower me to merge less critical pull requests on my own without further reviews. This can be seen especially critical since we still have not enough test coverage. Especially a safety net with end-to-end tests is missing. I'll do my best job to find a balance between making progress and working secure as a single developer for the majority of the upcoming pull requests that are frontend related. Note that I'll have to use my permissions as repository administrator to circumvent the blocked merge button.

weilbith:
There was an issue with other listings in the UI, because the `Progress.vue` component did not set its styling to `scoped`. This got fixed in #717 and it fine now. Though not good. 😑

PR #724, opened 2022-06-07
frontend: allow token selection without wallet

weilbith:
When the user has not connected with his wallet yet, the list of selectable tokens is empty even if he has chosen a source chain. The reason was that the tokens get filtered by the connected chain of the wallet, not the selected chain. But the user tends to first put in some values before connecting. Seeing that no token is available is a bad experience.

This is includes a tiny fix to open the block explorer link in a new tab. This somehow got lost during merge conflict resolution.

### DIFF
--- a/frontend/src/components/RequestFormInputs.vue
+++ b/frontend/src/components/RequestFormInputs.vue
@@ -152,8 +152,10 @@ const { selectedSourceChain, sourceChains, targetChains, switchChain } = useChai
   chains,
 );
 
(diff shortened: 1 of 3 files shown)


weilbith:
**Disclaimer:**
In result of the current personnel shortage, management has decided to empower me to merge less critical pull requests on my own without further reviews. This can be seen especially critical since we still have not enough test coverage. Especially a safety net with end-to-end tests is missing. I'll do my best job to find a balance between making progress and working secure as a single developer for the majority of the upcoming pull requests that are frontend related. Note that I'll have to use my permissions as repository administrator to circumvent the blocked merge button.

Issue #547, opened 2022-04-11
Implement Non-Fill Proof


fredo:
# Implementation summary
This is a implementation summary out of #346.

## Problem description
If a request is not filled, the protocol is vulnerable to exploits. The protocol secured such cases by the real filler resolving all false claims eventually. If there is no real filler, then there is no incentive to do so.

1) Alice does an unfillable claim
2) Charlie claims
3) Charlie challenges own claim
4) Charlie counter challenges
5) Charlie wins as the claimer and is able to withdraw the deposit

It would only be prevented if there were a rightful filler who would call L1 eventually by escalating its own claims.

### Fill Manager
We need to add the possibility to invalidate fill hashes. Under the assumption that a fill hash cannot be foreseen, fill hashes from future blocks will be will be invalidated. fill hashes with older block number will be compared to the fills mapping. The mapping stores request hash to fill hash.

#### Change `fillId` to be hash of previous block
This change will give the `fillId` its original purpose, to prevent a false claim from becoming valid in the future (by filling in the future). This is under the assumption that future block hashes cannot be guessed.

#### Change `FillManager.fills` to `mapping(bytes32 => bytes32)`
We store the correct `fillHash` which filled the request in the mapping. With this we can invalidate all other fill hashes. 

#### `invalidateFillHash()`
Expose a function to the public to invalidate any fill hash. This will call `IProofSubmitter.submitNonFillProof(bytes32 fillHash)`.

```sol
function invalidateFillHash(requestHash, fillId) {
    fillHash = _createFillHash(requestHash, fillId);

    if(fills[requestHash] != fillHash){
        IProofSubmitter.submitNonFillProof(fillHash);
    }
}
```

#### `submitNonFillProof()`
Submits a L2 -> L1 message containing a fill hash. This means that the request hash (which was used to create the fill hash) was not filled by this fill hash.

#### resolveNonFill
forward nonFillProof to the registry.

calls `ResolutionRegistry.addNonFillProof()`

#### mapping invalidFillHashes `mapping(bytes32 => bool)`
mapping of invalid fill hashes. 

#### addNonFillProof()

adds a fill hash to `invalidFillHashes`

#### Multiparty bidding
Quite a big change. Allow multiple parties to bid on a claim as the challenger.
The bidding scheme doesnt change but it is not only the challenger who will be able to bid if the claimer is leading.
We need to store all stakes of all challengers and we need to store the last bidder.

[...]

PR #725, opened 2022-06-08
Withdraw invalidated claims

### LINKED ISSUE #674: Non-Fill: Withdraw claims when L1 resolved Non-Fill arrives
Once a L1 transmitted non-fill proof arrives, the agent needs to decide whether or not to withdraw a claim.

### Tasks
- [ ] Check if `withdraw` should be triggered when a claim is in the `invalidated_l1_resolved` state
- [ ] Implement necessary changes to `maybe_withdraw`
- [ ] Add tests

palango:
Resolves #674 
Resolves https://github.com/beamer-bridge/beamer/issues/547

### DIFF
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -260,7 +260,7 @@ def process_claims(context: Context) -> None:
             continue
 
         if claim.is_invalidated_l1_resolved:
-            # TODO: See https://github.com/beamer-bridge/beamer/issues/674
+            maybe_withdraw(claim, context)
             continue
 
         # Check if claim is an honest claim. Honest claims can be ignored.
@@ -452,6 +452,10 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
         # Claimer cheated
         if agent_is_challenger and claim.claimer != request.l1_resolution_filler:
             _withdraw(claim, context)
+
+    # Claim has a non-fill proof and the agent is challenging
+    elif claim.is_invalidated_l1_resolved and agent_is_challenger:
+        _withdraw(claim, context)
 
     # Otherwise check that the challenge period is over
     elif block["timestamp"] >= claim.termination:
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -459,10 +459,6 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
         if context.address in winning_addresses:
             _withdraw(claim, context)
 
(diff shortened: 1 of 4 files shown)


beamer/chain.py
```diff
@@ -459,6 +459,9 @@ def maybe_withdraw(claim: Claim, context: Context) -> None:
         if context.address in winning_addresses:
             _withdraw(claim, context)
 
+    if claim.is_invalidated_l1_resolved and agent_is_challenger:
+        _withdraw(claim, context)
+
```
fredo, line 464: I think we have to include that into the `if / elif condition`. Equal to the `has_l1_resolution` the timestamp doesn't matter if the fill hash is invalidated. So it should be prioritized higher than the claim termination.
In addition to that I think it is possible that we call `_withdraw()` twice the way it is right now.

beamer/tests/agent/unit/test_invalidate.py
```diff
@@ -159,3 +163,39 @@ def test_maybe_invalidate_claim_wrong_fill_id_but_timed_out(mocked_invalidate):
...
+    ],
+)
+def test_maybe_withdraw_after_invalidation(mocked_withdraw, test_data):
+    context, _ = make_context()
+
+    request = make_request()
+    context.requests.add(request.id, request)
+
+    participant1 = test_data[0]
+    participant2 = test_data[1]
```
fredo, line 184: renaming it to claimer and challenger would be more explicit

palango: ✔️

beamer/tests/agent/unit/test_invalidate.py
```diff
@@ -159,3 +163,39 @@ def test_maybe_invalidate_claim_wrong_fill_id_but_timed_out(mocked_invalidate):
...
+
+    claim = make_claim_challenged(
+        request=request,
+        claimer=participant1,
+        challenger=participant2,
+        # Claim should not be withdrawable yet
+        termination=Termination(TIMESTAMP + 1),
+    )
+    context.claims.add(claim.id, claim)
+    claim.start_challenge(make_tx_hash())
```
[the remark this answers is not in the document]
palango: ✔️

beamer/chain.py
palango: ✔️

fredo (approved):
LGTM!

Issue #475, opened 2022-03-30
Add option to block creation of new requests in `RequestManager`


palango:
In case we want to switch to new contracts it might make sense to disable to possibility of creating new requests. This would not stop the lifecycle of already running requests and is therefore different from #474

PR #733, opened 2022-06-09
contracts: Add deprecation switch to RequestManager

fredo:
fixes: #475


palango:
Looks good to me, but should we also have code to deprecate the fill manager?

fredo:
> Looks good to me, but should we also have code to deprecate the fill manager?

I cannot see a real benefit for it, assuming that we redeploy all the contracts for a new version. But it certainly does not harm.

palango (approved):

PR #750, opened 2022-06-13
deployment: Rinkeby deployment

fredo:


istankovic:
It would be nice to run the E2E test manually, to make sure that the everything was deployed correctly.
See https://github.com/beamer-bridge/beamer/blob/main/docker/optimism/optimism.sh#L51-L54.

fredo:
> It would be nice to run the E2E test manually, to make sure that everything was deployed correctly. See https://github.com/beamer-bridge/beamer/blob/main/docker/optimism/optimism.sh#L51-L54.

I'm not sure if I understand how that would work on an actual testnet. Based on the code I can see that a L1 resolution is tested. Neither is there a guaranteed L1 relayer for testnet rollups nor is the finalization time short enough that this would make sense. I think for both Metis and Boba finalization times are 7 days on testnet. This means the e2e test would need to wait 7 days right now.

fredo:
In any way, I still just let the e2e run and it didn't ran into a failure. So at least the FillManager seemed to work.

istankovic:
> This means the e2e test would need to wait 7 days right now.

I mean, for testnet we could say we simply don't care. Not ideal, but if we want to go out ASAP, we can pretend it's not important.
For mainnet, however, we definitely want to test the deployment and make absolutely sure that L1 resolution works.
And the sooner we try that process (of manually testing L1 resolution), the better.

> I do think that we want to test such things on testnet, but we need our own infrastructure for that. I suggested a solution a while ago to you.

Could you please remind me, do we have an issue for that?

> The e2e test finally worked.

Awesome, thanks!

> Though it caused a problem in the agent as the e2e test runs a Boba -> Boba transfer, and the Agent cannot process the events as no request exists.

That should not be an issue. With the E2E test we're not really testing the agent(s), we're testing the contracts and the deployment. Of course, it would be nice to have a complete test with agents and everything, but we are not there (yet).
But at least making sure that the message flow L2 -> L1 -> L2 works correctly allows us to be more confident about L1 resolution. What remains is the agent, which we can test rather more easily in our brownie-based test environment.

fredo:
In addition to my comment, the e2e-test.py tests an e2e test for the same chain id. For manual tests I would rather like to see rollup A -> rollup B L1 resolution and vice versa. Because that's the case we are gonna have in real life

palango (approved):
LGTM

Issue #711, opened 2022-06-03
Minting token fails


palango:
As found by @weilbith 

#### Expected behavior

Mint script works

#### Actual behavior

```
/Users/paul/Library/Caches/pypoetry/virtualenvs/beamer-1vah2Hfk-py3.9/lib/python3.9/site-packages/web3/eth.py:656: UserWarning: There was an issue with the method eth_maxPriorityFeePerGas. Calculating using eth_feeHistory.
  warnings.warn(
Traceback (most recent call last):
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/beamer-1vah2Hfk-py3.9/lib/python3.9/site-packages/web3/eth.py", line 654, in max_priority_fee
    return self._max_priority_fee()
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/beamer-1vah2Hfk-py3.9/lib/python3.9/site-packages/web3/module.py", line 57, in caller
    result = w3.manager.request_blocking(method_str,
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/beamer-1vah2Hfk-py3.9/lib/python3.9/site-packages/web3/manager.py", line 198, in request_blocking
    return self.formatted_response(response,
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/beamer-1vah2Hfk-py3.9/lib/python3.9/site-packages/web3/manager.py", line 171, in formatted_response
    raise ValueError(response["error"])
ValueError: {'code': -32601, 'message': 'the method eth_maxPriorityFeePerGas does not exist/is not available'}

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/paul/Projects/brainbot/beamer/scripts/call_contracts.py", line 375, in <module>
    cli()
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/beamer-1vah2Hfk-py3.9/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/beamer-1vah2Hfk-py3.9/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/beamer-1vah2Hfk-py3.9/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
[...]

PR #751, opened 2022-06-13
Fix call_contracts.py 

fredo:
`call_contracts.py` did not work because of a misconfigured web3 object. The code from e2e-test and the deployment script was reused to setup web3. 

Since all the 3 scripts share the same code to setup web3, load contract information and create an account, I moved the code to `scripts/_util.py` to remove the redundant code.

fixes: #711

### DIFF
--- a/docker/optimism/optimism.sh
+++ b/docker/optimism/optimism.sh
@@ -50,7 +50,8 @@ create_deployment_config_file() {
 
 e2e_test() {
     l2_rpc=http://localhost:8545
-    poetry run python "${ROOT}/scripts/e2e-test.py" ${DEPLOYMENT_DIR} ${KEYFILE} ${l2_rpc}
+    password=""
+    poetry run python "${ROOT}/scripts/e2e-test.py" ${DEPLOYMENT_DIR} ${KEYFILE} "${password}" ${l2_rpc}
 }
 
 usage() {
--- a/scripts/_util.py
+++ b/scripts/_util.py
@@ -1,9 +1,13 @@
+from pathlib import Path
 from typing import Optional
 
 import click
 from eth_utils import decode_hex, is_checksum_address, is_hexstr, to_canonical_address
+from web3 import Web3
+from web3.contract import Contract
 
-from beamer.typing import Address
+import beamer.contracts
+from beamer.typing import Address, ChainId
 
 
 def validate_address(
@@ -27,3 +31,9 @@ def validate_bytes(
         raise click.BadParameter("not a hex string")
 
     return decode_hex(value)
+
+
+def contracts_for_web3(web3: Web3, deployment_dir: Path) -> dict[str, Contract]:
+    deployment_info = beamer.contracts.load_deployment_info(deployment_dir)
+    chain_id = ChainId(web3.eth.chain_id)
+    return beamer.contracts.make_contracts(web3, deployment_info[chain_id])
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -2,11 +2,7 @@ import threading
 from concurrent.futures import ThreadPoolExecutor
 
 import structlog
-import web3
-from eth_account.signers.local import LocalAccount
 from eth_typing import Address
-from web3.gas_strategies.rpc import rpc_gas_price_strategy
-from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware
 
 import beamer.metrics
 from beamer.chain import EventMonitor, EventProcessor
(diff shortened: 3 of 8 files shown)


docker/optimism/optimism.sh
```diff
@@ -50,7 +50,8 @@ create_deployment_config_file() {
 
 e2e_test() {
     l2_rpc=http://localhost:8545
-    poetry run python "${ROOT}/scripts/e2e-test.py" ${DEPLOYMENT_DIR} ${KEYFILE} ${l2_rpc}
+    password=""
```
istankovic, line 53:
```suggestion
    password="''"
```

scripts/_util.py
```diff
@@ -27,3 +36,33 @@ def validate_bytes(
...
+    rpc: str, keystore_file: Path, password: str, deployment_dir: Path
+) -> tuple[LocalAccount, Web3, dict[str, Contract]]:
+    account = account_from_keyfile(keystore_file, password)
+    web3 = web3_for_rpc(rpc, account)
+    deployment_info = beamer.contracts.load_deployment_info(deployment_dir)
+
+    chain_id = ChainId(web3.eth.chain_id)
+    contracts = beamer.contracts.make_contracts(web3, deployment_info[chain_id])
+
+    return account, web3, contracts
```
istankovic, line 68: Do we really need this function?
I feel it does too much: it loads the keyfile, constructs a `Web3` instance, queries the chain ID, loads the deployment info and then constructs the contracts. The name is also not reflecting the functionality of this function.

I think just moving `account_from_keyfile` and `web3_for_rpc` to a common place already improves things a lot, to the point that we really do not this "setup-everything" function. What do you think?

fredo: Yes I thought about the same thing last night, and are about to change it back. It's not really useful

fredo: I composed account creation and web3 creation into one, because whenever we create an account we almost always want to create web3 as well. Let me know if I even should split up further, I dont really mind

scripts/_util.py
```diff
@@ -27,3 +36,33 @@ def validate_bytes(
...
+def _account_from_keyfile(keyfile: Path, password: str) -> LocalAccount:
+    with open(keyfile, "rt") as fp:
+        privkey = Account.decrypt(json.load(fp), password)
+    return cast(LocalAccount, Account.from_key(privkey))
+
+
+def web3_for_rpc(rpc: str, account: LocalAccount) -> Web3:
+    web3 = Web3(HTTPProvider(rpc))
+
+    web3.eth.set_gas_price_strategy(rpc_gas_price_strategy)
```
palango, line 50: For the rollups this is probably right, but on mainnet we should use something smarter.

scripts/_util.py
```diff
@@ -27,3 +36,33 @@ def validate_bytes(
         raise click.BadParameter("not a hex string")
 
     return decode_hex(value)
+
+
+def _account_from_keyfile(keyfile: Path, password: str) -> LocalAccount:
+    with open(keyfile, "rt") as fp:
+        privkey = Account.decrypt(json.load(fp), password)
+    return cast(LocalAccount, Account.from_key(privkey))
```
istankovic, line 44: I would actually make this public, move it to `beamer.util`, and make use of it both under `scripts/` and in `beamer/cli.py`, which already has `_account_from_keyfile`. It is a generally useful function.

scripts/_util.py
```diff
@@ -27,3 +36,33 @@ def validate_bytes(
...
+
+def web3_for_rpc(rpc: str, account: LocalAccount) -> Web3:
+    web3 = Web3(HTTPProvider(rpc))
+
+    web3.eth.set_gas_price_strategy(rpc_gas_price_strategy)
+    web3.middleware_onion.inject(geth_poa_middleware, layer=0)
+    web3.middleware_onion.add(construct_sign_and_send_raw_middleware(account))
+    web3.eth.default_account = account.address
+
+    return web3
```
istankovic, line 55: This function is suspiciously similar to the following from `beamer.agent`:

```python
def _make_web3(url: URL, account: LocalAccount) -> web3.Web3:
    w3 = web3.Web3(web3.HTTPProvider(url, request_kwargs=dict(timeout=5)))
    w3.eth.set_gas_price_strategy(rpc_gas_price_strategy)
    # Add POA middleware for geth POA chains, no/op for other chains
    w3.middleware_onion.inject(geth_poa_middleware, layer=0)
    w3.middleware_onion.add(construct_sign_and_send_raw_middleware(account))
    w3.eth.default_account = account.address
    return w3
```
Similarly to the `account_from_keyfile`, I would suggest to introduce a `beamer.util.make_web3` function and use it both in `beamer.agent` and under `scripts/`.

scripts/_util.py
istankovic: Yeah, I think we should just go without `setup_web3` -- it's ok to have to write a few lines extra now and then.

istankovic:
A general comment on commit sequencing: it is often helpful to separate the introduction of new functions and the actual usage of those functions, especially if old code is being replaced. To give an example (oldest commit first):

```
- beamer: util: add X
- beamer: agent: use beamer.util.X
- scripts: use beamer.util.X
```

This way, it is much easier to follow the commit (contents) and do the review, as well as inspect the git history later.

scripts/_util.py
fredo: I added an optional parameter

fredo:
@istankovic I addressed your comments. Please have a look again, also let me know if the commit sequence is fine grain enough.

docker/optimism/optimism.sh
fredo: fyi that doesn't work. It puts the string `''` as the password. 

Quoting the parameter did the job `..... "${password}" ...`

PR #754, opened 2022-06-14
agent: Wait to trigger L1 resolution until finalization

### LINKED ISSUE #749: Only start the relayer when transaction is finalized
Currently the relayer is started whenever L1 criteria are fulfilled. 

We should add one more criteria, which is we should only call the relayer when the message is finalized. 

A message is approximately finalized when `finalization period` of the roll up has passed since the message was sent.

palango:
Resolves #749

Delay calling the relayer until the finalization period is over.

I decided to store the timestamps in the request and claim objects instead of recalculating it from the transaction every time to save some rpc requests.

### DIFF
--- a/beamer/state_machine.py
+++ b/beamer/state_machine.py
@@ -34,7 +34,7 @@ from beamer.models.claim import Claim
 from beamer.models.request import Request
 from beamer.tracker import Tracker
 from beamer.typing import ChainId, ClaimId, FillHash, RequestHash, RequestId
-from beamer.util import SECONDS_PER_DAY, TokenMatchChecker
+from beamer.util import TokenMatchChecker
 
 log = structlog.get_logger(__name__)
 
@@ -121,21 +121,27 @@ def _find_request_by_request_hash(
     return None
 
 
-def _invalidation_ready_for_l1_relay(claim: Claim) -> bool:
+def _invalidation_ready_for_l1_relay(
+    claim: Claim, target_chain_id: ChainId, request_manager: Contract
+) -> bool:
+    target_chain_finalization = request_manager.functions.finalizationTimes(target_chain_id).call()
     return (
         not claim.is_invalidated_l1_resolved
         and claim.invalidation_tx is not None
         and claim.invalidation_timestamp is not None
-        and int(time.time()) > claim.invalidation_timestamp + 7 * SECONDS_PER_DAY
+        and int(time.time()) > claim.invalidation_timestamp + target_chain_finalization
     )
 
 
-def _proof_ready_for_l1_relay(request: Request) -> bool:
+def _proof_ready_for_l1_relay(request: Request, request_manager: Contract) -> bool:
+    target_chain_finalization = request_manager.functions.finalizationTimes(
+        request.target_chain_id
+    ).call()
     return (
         not request.is_l1_resolved
         and request.fill_tx is not None
         and request.fill_timestamp is not None
-        and int(time.time()) > request.fill_timestamp + 7 * SECONDS_PER_DAY
+        and int(time.time()) > request.fill_timestamp + target_chain_finalization
     )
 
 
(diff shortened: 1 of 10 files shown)


beamer/state_machine.py
```diff
@@ -235,7 +265,7 @@ def _handle_claim_made(event: ClaimMade, context: Context) -> HandlerResult:
         return False, None
 
     events: list[Event] = []
-    if not claim.is_invalidated_l1_resolved and claim.invalidation_tx is not None:
+    if _invalidation_ready_for_l1_relay(claim, request.target_chain_id, context.request_manager):
```
fredo, line 268: If the condition is false, then `InitiateL1InvalidationEvent` will not be created. And if no other ClaimMade event will ever arrive, then it will not trigger L1 resolution. 

I think what would happen now is, that as long as the above condition is false, the agent will counter challenge again, Probably leading to drain all the funds (depending on the attacker).

fredo: Well, I'm not exactly sure about the solution. But currently what happens is: 

As long as the condition is not true, which is finalization time has not passed yet, then we will ultimately engage in the challenge again. That's not what we want. As soon as the agent has gathered enough funds (l1_criteria_fulfilled), we want him to wait and not counter challenge. 

That's what happened before by moving to the l1_resolved state. But since we are not emitting the event before finalization time is passed we will either stay in `claimer_winning` or `challenger_winning` and this will lead to another challenge transaction.

palango: Nice find, I pulled out the wait into the handling of said events.

beamer/state_machine.py
```diff
@@ -367,9 +367,17 @@ def _handle_initiate_l1_resolution(
 
     # A request should never be dropped before all claims are finalized
     assert request is not None, "Request object missing"
-
     assert request.fill_tx is not None, "Request not yet filled"
-    if _l1_resolution_criteria_fulfilled(claim, context):
+    assert request.fill_timestamp is not None, "Request not yet filled"
+
+    # Check that message is finalized on L1
+    if not _timestamp_is_l1_finalized(
```
fredo, line 374: this will cost one RPC call per second for each claim whose request was filled (can also be handled in a later PR)

fredo (approved):
Most of my comments are optimizations for the future.

beamer/state_machine.py
```diff
@@ -121,21 +121,27 @@ def _find_request_by_request_hash(
     return None
 
 
-def _invalidation_ready_for_l1_relay(claim: Claim) -> bool:
+def _invalidation_ready_for_l1_relay(
+    claim: Claim, target_chain_id: ChainId, request_manager: Contract
+) -> bool:
+    target_chain_finalization = request_manager.functions.finalizationTimes(target_chain_id).call()
```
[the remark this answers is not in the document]
palango: We don't necessarily need an event for this, we can just query it on startup and store it somewhere

beamer/tests/agent/unit/test_handlers.py
```diff
@@ -138,7 +138,6 @@ def test_handle_request_resolved():
 @pytest.mark.parametrize("timestamp", [TIMESTAMP, int(time.time())])
```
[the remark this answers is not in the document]
palango: Before your earlier comment, this was actually testing the emission of the event. But I thought it wouldn't hurt to have this as the tests are fast. 
I can remove it though.

beamer/state_machine.py
palango: Created #784 for this

beamer/tests/agent/unit/test_handlers.py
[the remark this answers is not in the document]
palango: On second thought, I'll remove it because it isn't clear.

beamer/state_machine.py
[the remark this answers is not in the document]
istankovic: > We don't necessarily need an event for this, we can just query it on startup and store it somewhere

Those were mine thoughts too, but then I remembered that `setFinalizationTime` can be called multiple times.
Which is maybe not what we'd like to allow -- in that case, yes, I'd say look it up on startup and we're done.

Otherwise, we need to consider the case where the finalization time has been changed. (Not sure whether we really need that.)

PR #755, opened 2022-06-14
Agent needs to sleep when there are unprocessed events in the queue

fredo:
fixes: #752


palango (approved):
👍 
Feel free to add a test or do it after the deployment is fixed.

Issue #756, opened 2022-06-14
Agent crashes when ClaimWithdrawn event is processed in started state


fredo:
The agent crashes when a ClaimWithdrawn event is processed in the started state. But this can happen during sync.

PR #759, opened 2022-06-14
agent: Allow claim state transition from started to withdrawn

fredo:
fixes: #756


istankovic:
Please do not forget to update the claim state machine in the docs.

PR #762, opened 2022-06-15
Deployment script improvements

fredo:
We need to improve the deployment script to work on mainnet (different behavior due to other timing and gas price behavior).

Some of the optimizations:

- expose the timeout of `wait_for_transaction()` in `transact()`
- increase the timeout upon deploying
- better gas price strategy for L1 deployments


beamer/util.py
```diff
@@ -30,13 +30,20 @@ def cause(self) -> str:
         return str(self.__cause__)
 
 
-def transact(func: Union[ContractConstructor, ContractFunction], **kwargs: Any) -> Any:
+def transact(
+    func: Union[ContractConstructor, ContractFunction],
+    timeout: float = 120,
```
palango, line 35: Maybe just make 600 the default here

palango: I'm not sure if 2 or 10 mins make a big difference, both are way to long if that should happen.

If we want to assume that this is faster on rollups, then I would lower the default.

fredo: okay. 120 was just the default from the function itself. I chose the value to not change the default.

fredo:
@palango Could you review this again, please? I'd like to do two separate PRs as changing the fundamental structure of the script will be a big change. This PR only contains small adjustments

scripts/call_contracts.py
```diff
@@ -309,5 +309,44 @@ def whitelist(
...
+    contracts: dict[str, Contract],
+    protocol_fee_ppm: int,
+    lp_fee_ppm: int,
+    min_lp_fee: int,
+) -> None:
+    """Update fee of Request Manager. Must be done by owner."""
+
+    request_manager = contracts["RequestManager"]
+
+    tx_hash = request_manager.functions.updateFeeData(
```
palango, line 343: Might be worthwhile to check the owner here, but we can add this later.

palango (approved):
LGTM

Issue #301, opened 2022-03-04
Add natspec docs to contracts


palango:
- Function params and what they return (public API docs)
- Generate HMTL docs and integrate them in our current documentation.

PR #785, opened 2022-06-21
Contract docs

istankovic:
Closes: #301

### DIFF
--- a/contracts/contracts/FillManager.sol
+++ b/contracts/contracts/FillManager.sol
@@ -7,9 +7,19 @@ import "OpenZeppelin/openzeppelin-contracts@4.5.0/contracts/access/Ownable.sol";
 import "../interfaces/IProofSubmitter.sol";
 import "./BeamerUtils.sol";
 
+/// The fill manager.
+///
+/// This contract is responsible for keeping track of filled requests. In addition to allowing
+/// agents to (eventually) prove that they filled requests, it also allows anyone to invalidate
+/// a claim that a request was filled.
+///
+/// It is the only contract that agents need to interact with on the target chain.
 contract FillManager is Ownable {
     using SafeERC20 for IERC20;
 
+    /// Emitted when a request has been filled.
+    ///
+    /// .. seealso:: :sol:func:`fillRequest`
     event RequestFilled(
         uint256 indexed requestId,
         bytes32 fillId,
@@ -19,28 +29,64 @@ contract FillManager is Ownable {
         uint256 amount
     );
 
+    /// Emitted when a fill hash has been invalidated.
+    ///
+    /// .. seealso:: :sol:func:`invalidateFillHash`
     event HashInvalidated(
         bytes32 indexed requestHash,
         bytes32 indexed fillId,
         bytes32 indexed fillHash
     );
 
+    /// Emitted when a liquidity provider has been added to the set of allowed
+    /// liquidity providers.
+    ///
+    /// .. seealso:: :sol:func:`addAllowedLP`
     event LPAdded(address lp);
 
+    /// Emitted when a liquidity provider has been removed from the set of allowed
+    /// liquidity providers.
+    ///
+    /// .. seealso:: :sol:func:`removeAllowedLP`
     event LPRemoved(address lp);
 
+    /// The L1 :sol:contract:`Resolver` contract to be used for L1 resolution.
     address public l1Resolver;
+
+    /// The proof submitter contract to be used for submitting L1 proofs.
+    ///
+    /// The specific implementation of the :sol:interface:`IProofSubmitter` interface
+    /// is chain-dependent.
     IProofSubmitter public proofSubmitter;
 
-    // mapping from request hash to fill hash of filled requests
+    /// Maps request hashes to fill hashes.
     mapping(bytes32 => bytes32) public fills;
+
+    /// The set of liquidity providers that are allowed to fill requests.
     mapping(address => bool) public allowedLPs;
 
+    /// Constructor.
+    ///
+    /// @param _l1Resolver The L1 resolver contract.
+    /// @param _proofSubmitter The proof submitter.
     constructor(address _l1Resolver, address _proofSubmitter) {
         l1Resolver = _l1Resolver;
         proofSubmitter = IProofSubmitter(_proofSubmitter);
     }
 
+    /// Fill the specified request.
+    ///
+    /// The caller must have approved at least ``amount`` tokens for :sol:contract:`FillManager`
+    /// with the ERC20 token contract at ``targetTokenAddress``. The tokens will be immediately
+    /// sent to ``targetReceiverAddress`` and a fill proof will be generated, which can later
+    /// be used to trigger L1 resolution, if needed.
+    ///
+    /// @param requestId The request ID.
+    /// @param sourceChainId The source chain ID.
+    /// @param targetTokenAddress Address of the token contract on the target chain.
+    /// @param targetReceiverAddress Recipient address on the target chain.
+    /// @param amount Amount of tokens to transfer. Does not include fees.
+    /// @return The fill ID.
     function fillRequest(
         uint256 requestId,
         uint256 sourceChainId,
@@ -81,6 +127,17 @@ contract FillManager is Ownable {
         return proofReceipt.fillId;
     }
 
+    /// Invalidate the specified fill.
+    ///
+    /// In cases that a claim has been made on the source chain, but without a corresponding fill
+    /// actually happening on the target chain, anyone can call this function to mark the fill
+    /// as invalid. This is typically followed by a challenge game on the source chain, which
+    /// the dishonest claimer is guaranteed to lose as soon as the information about the invalid
+    /// fill (so called "non-fill proof") is propagated to the source chain via L1 resolution.
+    ///
+    /// @param requestHash The request hash.
+    /// @param fillId The fill ID.
+    /// @param sourceChainId The source chain ID.
     function invalidateFillHash(
         bytes32 requestHash,
         bytes32 fillId,
@@ -97,12 +154,24 @@ contract FillManager is Ownable {
         emit HashInvalidated(requestHash, fillId, fillHash);
     }
 
+    /// Add a liquidity provider to the set of allowed liquidity providers.
+    ///
+    /// Only allowed liquidity providers can fill requests.
+    ///
+    /// .. note:: This function can only be called by the contract owner.
+    ///
+    /// @param newLP The liquidity provider.
     function addAllowedLP(address newLP) public onlyOwner {
         allowedLPs[newLP] = true;
 
         emit LPAdded(newLP);
     }
 
+    /// Remove a liquidity provider from the set of allowed liquidity providers.
+    ///
+    /// .. note:: This function can only be called by the contract owner.
+    ///
+    /// @param oldLP The liquidity provider.
     function removeAllowedLP(address oldLP) public onlyOwner {
         delete allowedLPs[oldLP];
 
--- a/contracts/contracts/RequestManager.sol
+++ b/contracts/contracts/RequestManager.sol
@@ -9,6 +9,13 @@ import "OpenZeppelin/openzeppelin-contracts@4.5.0/contracts/access/Ownable.sol";
 import "./BeamerUtils.sol";
 import "./ResolutionRegistry.sol";
 
+/// The request manager.
+///
+/// This contract is responsible for keeping track of transfer requests,
+/// implementing the rules of the challenge game and holding deposited
+/// tokens until they are withdrawn.
+///
+/// It is the only contract that agents need to interact with on the source chain.
 contract RequestManager is Ownable {
     using Math for uint256;
     using SafeERC20 for IERC20;
@@ -43,6 +50,10 @@ contract RequestManager is Ownable {
     }
 
     // Events
+
+    /// Emitted when a new request has been created.
+    ///
+    /// .. seealso:: :sol:func:`createRequest`
     event RequestCreated(
         uint256 requestId,
         uint256 targetChainId,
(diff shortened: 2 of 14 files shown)


contracts/contracts/RequestManager.sol
```diff
@@ -66,53 +83,99 @@ contract RequestManager is Ownable {
...
+    /// provide when making a claim, as well in each round of the challenge game.
     uint256 public claimStake;
+
+    /// The period for which the claim is valid.
     uint256 public claimPeriod;
+
+    /// The period by which the termination time of a claim is extended after each
+    /// round of the challenge game. This period should allow enough time for the
+    /// other parties to counter-challenge.
+    ///
```
fredo, line 108: There is a missing piece. The termination time is only extended by it if current terminition time is shorter than `challengePeriodExtension`

contracts/contracts/RequestManager.sol
```diff
@@ -66,53 +83,99 @@ contract RequestManager is Ownable {
...
+    /// Finalization times are in seconds.
+    mapping(uint256 => uint256) public finalizationTimes;
 
+    /// Maps request IDs to requests.
     mapping(uint256 => Request) public requests;
+
+    /// Maps claim IDs to claims.
     mapping(uint256 => Claim) public claims;
 
+    /// The minimum fee paid to the liquidity provider.
```
fredo, line 145: To prevent confusion, would also be good to state that this value is denominated in transfer token.

contracts/contracts/RequestManager.sol
```diff
@@ -138,6 +208,16 @@ contract RequestManager is Ownable {
...
 
+    /// Create a new transfer request.
+    ///
+    /// @param targetChainId ID of the target chain.
+    /// @param sourceTokenAddress Address of the token contract on the source chain.
+    /// @param targetTokenAddress Address of the token contract on the target chain.
+    /// @param targetAddress Recipient address on the target chain.
+    /// @param amount Amount of tokens to transfer. Does not include fees.
+    /// @param validityPeriod The number of seconds the request is to be considered valid.
+    ///                       A request cannot be fulfilled once its validity period has elapsed.
```
fredo, line 219: That sounds a bit confusing. Technically the request can be filled anytime. But the request cannot be claimed anymore and the user is entitled to withdraw the tokens if there are not active claims.
I'm not sure how to describe this properly.

fredo, line 219: Maybe just change it to 

`A request cannot be claimed once its validity period has elapsed`

contracts/contracts/RequestManager.sol
```diff
@@ -270,6 +364,33 @@ contract RequestManager is Ownable {
...
+    /// can join the game and make another challenge.
+    ///
+    /// The caller must provide enough native tokens as their stake.
+    /// For the original claimer, the minimum stake is
+    /// ``challengerStakeTotal - claimerStake + claimStake``.
+    ///
+    /// For challengers, the minimum stake is
+    /// ``claimerStake - challengerStakeTotal + 1``.
+    ///
+    /// An example (time flows downwards, claimerStake = 10)::
```
fredo, line 383: claimerStake = claimStake

contracts/contracts/RequestManager.sol
```diff
@@ -270,6 +364,33 @@ contract RequestManager is Ownable {
...
+    ///
+    ///   claimRequest() by Max [stakes 10]
+    ///   challengeClaim() by Alice [stakes 11]
+    ///   challengeClaim() by Max [stakes 11]
+    ///   challengeClaim() by Bob [stakes 16]
+    ///
+    /// In this example, if Max didn't want to lose the challenge game to
+    /// Alice and Bob, he would have to challenge with a stake of at least 16.
```
fredo, line 391: Nice description!

contracts/contracts/FillManager.sol
```diff
@@ -7,9 +7,19 @@ import "OpenZeppelin/openzeppelin-contracts@4.5.0/contracts/access/Ownable.sol";
 import "../interfaces/IProofSubmitter.sol";
 import "./BeamerUtils.sol";
 
+/// The fill manager.
+///
+/// This contract is responsible for keeping track of filled requests. In addition to allowing
+/// agents to (eventually) prove that they filled requests, it also allows anyone to invalidate
+/// a claim that a request was filled.
```
fredo, line 14: More accurately, it would invalidate a fill hash (or a fill id in combination with a request (hash) , respectively)

contracts/contracts/RequestManager.sol
```diff
@@ -43,6 +50,8 @@ contract RequestManager is Ownable {
     }
 
     // Events
+
+    /// Emitted when a new request has been created.
```
fredo, line 54: should we also add the `see also: ` keyword, as done in the fill manager?

contracts/contracts/FillManager.sol
```diff
@@ -19,28 +29,64 @@ contract FillManager is Ownable {
...
+    /// @param _l1Resolver The L1 resolver contract.
+    /// @param _proofSubmitter The proof submitter.
     constructor(address _l1Resolver, address _proofSubmitter) {
         l1Resolver = _l1Resolver;
         proofSubmitter = IProofSubmitter(_proofSubmitter);
     }
 
+    /// Fill the specified request.
+    ///
+    /// The caller must approve ``amount`` tokens for :sol:contract:`FillManager` with
```
fredo, line 79: `... must have approved at least...`

contracts/contracts/ResolutionRegistry.sol
```diff
@@ -4,16 +4,43 @@ pragma solidity ^0.8.12;
 import "./CrossDomainRestrictedCalls.sol";
 import "./BeamerUtils.sol";
 
+/// The resolution registry.
+///
+/// This contract resides on the source L2 chain and is tasked with keeping track of results
+/// of L1 resolution. In particular, it stores the information about known fills and fillers,
+/// as well as fills that were marked invalid. This information is used by the :sol:contract:`RequestManager`
```
fredo, line 11: would probably also go with fill hashes instead of fills.

contracts/contracts/ResolutionRegistry.sol
```diff
@@ -4,16 +4,43 @@ pragma solidity ^0.8.12;
...
+    /// Maps request hashes to fill infos.
     mapping(bytes32 => BeamerUtils.FillInfo) public fillers;
-    // mapping from fillHash to validity flag
+
+    /// The set of invalid fill hashes.
     mapping(bytes32 => bool) public invalidFillHashes;
 
+    /// Resolve the specified request.
+    ///
+    /// Callable only by the :sol:contract:`Resolver` contract via a messenger, this
```
fredo, line 37: Technically, it is only callable by the `l1_messenger_contract`. I know that this contract can only be called by the resolver, but in the l1 messenger contract you also describe that the messenger can only be called by the resolver. In order to stay consistent how the message chain is described throughout the whole documentation, I would actually prefer to make it clear that the function is only allowed to be called by the l1 messenger

contracts/contracts/ResolutionRegistry.sol
```diff
@@ -4,16 +4,43 @@ pragma solidity ^0.8.12;
...
+/// The resolution registry.
+///
+/// This contract resides on the source L2 chain and is tasked with keeping track of results
+/// of L1 resolution. In particular, it stores the information about known fills and fillers,
+/// as well as fills that were marked invalid. This information is used by the :sol:contract:`RequestManager`
+/// to resolve claims.
+///
+/// .. note::
+///
+///   This contract can only be called by the :sol:contract:`Resolver` contract, via a
```
fredo, line 16: see https://github.com/beamer-bridge/beamer/pull/785/commits/aab394ac52221b23f83701a2274fcc2006e80bc1#r905148152

contracts/contracts/ResolutionRegistry.sol
```diff
@@ -32,6 +59,14 @@ contract ResolutionRegistry is CrossDomainRestrictedCalls {
         emit RequestResolved(requestHash, filler, fillId);
     }
 
+    /// Mark the fill as invalid.
+    ///
+    /// Callable only by the :sol:contract:`Resolver` contract via a messenger, this
```
fredo, line 64: see https://github.com/beamer-bridge/beamer/pull/785/commits/aab394ac52221b23f83701a2274fcc2006e80bc1#r905148152

fredo (changes requested):
Thanks for this work! I gave a couple of nitpicks and suggestions. 

Is there a reason why the other contracts are not documented? There is no documentation for the contracts which implement interfaces, the BeamerUtils library and RestrictedCalls.sol.

contracts/contracts/RequestManager.sol
istankovic: Added a note on that.

contracts/contracts/RequestManager.sol
istankovic: > 



> That sounds a bit confusing. Technically the request can be filled anytime. But the request cannot be claimed anymore and the user is entitled to withdraw the tokens if there are not active claims. I'm not sure how to describe this properly.

This is a very good point. Yeah, it's a bit tricky to convey the meaning.
I went with
```
Once its validity period has elapsed, the request cannot be claimed
anymore and will eventually expire, allowing the request submitter 
to withdraw the deposited tokens if there are no active claims.    
```

contracts/contracts/FillManager.sol
istankovic: Right, but this is high-level description of the contract, we're not really concerned about details on that level.

contracts/contracts/ResolutionRegistry.sol
```diff
@@ -32,6 +59,14 @@ contract ResolutionRegistry is CrossDomainRestrictedCalls {
...
+    /// marks the fill as invalid.
```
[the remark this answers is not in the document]
istankovic: > why not `` marks the fill identified by ``fillHash`` as invalid ``

Eh... if we only had ``fillHash`` in the arguments... :smile: 
But yeah, I see your point, changed to mention ``fillId``.

contracts/contracts/ResolutionRegistry.sol
istankovic: > Technically, it is only callable by the `l1_messenger_contract`. I know that this contract can only be called by the resolver, but in the resolver you also describe that it can only be called by the proof submitter. So ultimately the addresses which are stored as `callers`. In order to stay consistent how the message chain is described throughout the whole documentation, I would actually prefer to make it clear that the function is only allowed to be called by the l1 messenger

Hm, I thought I was already being consistent -- by including the "... via a messenger ..." wording everywhere.
Technically, yes, the messengers are the immediate neighbours in the call chain, however, in the grand scheme
of things I feel it's more important to highlight where the call _really_ comes from (i.e. the proof submitter, or the resolver).

fredo: I thought about a bit more, how could one argue that in the grand scheme the resolver the *real caller*. One could argue that the *real caller* is defined where the call is constructed. 

In the IProofSubmitter the call to the resolver is constructed (`abi.encode...`). 
In the same sense the resovler contract constructs the resolution registry call. Following this argumentation, your documentation makes sense. 

I always try to read the docs as if I was an external developer getting familiar with our codebase. Reading in the docs that we have `Resolver -> Messenger -> ResolutionRegistry` and then reading in the contract documentation that the Resolver is the caller of the ResolutionRegistry would confuse me and I really want to make sure that we can prevent this confusion.
Defining the caller as the one who constructs the call would separate the definition of the "real caller" from the technical message flow. I think it is important to explicitly distiguish to not confuse readers.

istankovic:
> Is there a reason why the other contracts are not documented? There is no documentation for the contracts which implement interfaces, the BeamerUtils library and RestrictedCalls.sol.

I omitted the `BeamerUtils` because that's internals and private. I've just now added docs for restricted calls contracts.

contracts/contracts/ResolutionRegistry.sol
istankovic: I updated the wording.

contracts/contracts/RestrictedCalls.sol
```diff
@@ -3,16 +3,37 @@ pragma solidity ^0.8.12;
...
+    /// The set of callers, in the form of hashed pairs ``(chainId, address)``.
     mapping(bytes32 => bool) public callers;
 
+    /// Add a caller for the given chain ID.
+    ///
+    /// .. note:: There can only be one caller per chain.
+    ///
+    /// @param chainId The chain ID.
+    /// @param caller The caller.
     function addCaller(uint256 chainId, address caller) external onlyOwner {
```
fredo, line 21: this is not part of the documentation, but actually the documentation does not fit to the logic of the function. 
`.. note:: There can only be one caller per chain.`. This is not true as we never check against the chain ID, we only check against the hash of `(chainId, caller)` which will be different for any caller.

We should either change the code or the documentation.

fredo (changes requested):
please have a look at my comment on restricted calls

contracts/contracts/RestrictedCalls.sol
istankovic: Good point, removed.

fredo: Interesting point though, which we should investigate. CrossDomain is restricted to one caller but RestrictedCalls isnt.

contracts/contracts/RestrictedCalls.sol
```diff
@@ -3,16 +3,35 @@ pragma solidity ^0.8.12;
...
         bytes32 key = keccak256(abi.encodePacked(chainId, caller));
 
         require(!callers[key], "RestrictedCalls: caller already exists");
         callers[key] = true;
     }
 
+    /// Mark the function as restricted.
+    ///
+    /// Calls to the restricted function can only come from one address, that
```
fredo, line 28: Please also change this one here

istankovic: Done.

contracts/contracts/RestrictedCalls.sol
istankovic: Yup, do you thing that is something that we need to change?

fredo (approved):
LGTM now.

PR #792, opened 2022-06-23
Switch request state to claimed if own ClaimMade event was received

### LINKED ISSUE #782: Switch request state to claimed if own ClaimMade event was received
This was a bug happening on startup. Currently the agent only jumps into the `claimed` state directly after calling the claim transaction. That is fine as long as the agent is running. But on start up the agent potentially will have called the claim transaction for a request already. The `claimed` state represents the state where the agent claimed the request (because he is the filler). We should also jump into that state whenever we receive a `CLaimMade` event where the agent is the claimer.

palango:
Resolves #782


istankovic:
Could you please update the docs for the state machine?

beamer/tests/agent/unit/test_handlers.py
```diff
@@ -289,3 +298,76 @@ def test_maybe_claim_l1_as_challenger(
...
+
+@patch("beamer.metrics")
+def test_handling_claim_during_sync(_mocked_metrics):
+    """
+    Tests the problem encountered in https://github.com/beamer-bridge/beamer/issues/782
+    """
+    context, config = make_context()
+    # Assert that ACCOUNT.address is the agent's address
+    # Is used as an assurance that in case ACCOUNT is changed
+    assert config.account.address == ACCOUNT.address
```
istankovic, line 311: The comment looks a bit strange, I'm not sure I understand what is being checked here. What purpose does `ACCOUNT` have and how does that relate to the agent?

docs/source/request_state_machine.dot
```diff
@@ -4,11 +4,17 @@ digraph request_state_machine {
...
     filled -> claimed [label = try_to_claim];
+
     claimed -> withdrawn [label = withdraw];
     claimed -> l1_resolved [label = resolve];
+
     l1_resolved -> withdrawn [label = withdraw];
+
     filled -> withdrawn [label = withdraw];
     filled -> ignored [label = ignore];
+
```
istankovic, line 18: This does not look complete. You added 3 new state transitions in `request.py`, yet here I see only 1 transition added.

beamer/tests/agent/unit/test_handlers.py
palango: I just copied this from another test, but I'm fine with removing it.

beamer/tests/agent/unit/test_handlers.py
```diff
@@ -289,3 +298,76 @@ def test_maybe_claim_l1_as_challenger(
         assert mocked_withdraw.called
     else:
         assert not mocked_withdraw.called
+
+
+@patch("beamer.metrics")
+def test_handling_claim_during_sync(_mocked_metrics):
```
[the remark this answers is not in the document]
palango: For example in `_handle_request_filled` we update metrics, but in the unit tests those are not initialized.

docs/source/request_state_machine.dot
palango: We currently don't display loop transitions (i.e. state to itself)

beamer/tests/agent/unit/test_handlers.py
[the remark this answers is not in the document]
palango: Ideally we should decouple this

istankovic (approved):

PR #796, opened 2022-06-23
Fix wrong balance displayed when sending a tx

### LINKED ISSUE #598: Wrong value in token balance label
#### System information

OS & Version: All  
Browser & Version: All

#### Actual behavior

After a successful transaction, target chain token balance label shows double the amount of transferred token sum instead of the actual amount. For example (first screenshot), `1 TST` from METIS to BOBA, balance is `22 TST` on Metamask, while token balance label shows `23 TST = 21 + (1 x 2)`. 2nd screenshot, `3 TST` from METIS to BOBA, balance on Metamask is `25 TST`, while token balance label shows `28 TST = 22 + (3 x 2)`. This behavior occurs for a few minutes and then the balance returns to the right one matching Metamask! 

#### Expected behavior

Token balance label value should match Metamask one at all times!

Create a request and check the target chain balance just after making it!

compojoom:
fixes #598

It turns out that the filter for the events is also being executed for "past" events. According to this here: https://github.com/ethers-io/ethers.js/issues/1504 events are emited for the current block.

So what happens is - I execute a transfer, reload the page and my balance is wrong because I received the current block. But my balance was already updated.

My workaround for now is to only fetch the balance from the contract and not to try to calculate it myself.

### DIFF
--- a/frontend/src/composables/useTokenBalance.ts
+++ b/frontend/src/composables/useTokenBalance.ts
@@ -26,6 +26,28 @@ export function useTokenBalance(
 
   let tokenContract: Contract;
 
+  async function updateTokenBalance(
+    provider: IEthereumProvider | undefined,
+    tokenAddress: string | undefined,
+    signerAddress: string | undefined,
+  ) {
+    if (!provider || !tokenAddress || !signerAddress) {
+      balance.value = BigNumber.from(0);
+      return;
+    }
+
+    try {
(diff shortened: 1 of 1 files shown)


compojoom:
+1 on every transfer event. Shouldn't be that dramatic, since the events are emitted from the current block in the future.

weilbith (approved):
Seems fine to me. 👍🏾 
Should do the job and nothing to comment.

Issue #757, opened 2022-06-14
Adapt deployment script to receive contracts' constructor parameters as arguments


fredo:
The deployment script has hard coded values such as 

- claimPeriod = 1 hour
- claimStake = 0.0047 ETH
- challengePeriodExtension = 1 hour

With deploying on test and mainnet, it would be good to make those arguments consumable

fredo:
totally agree

PR #797, opened 2022-06-24
Deployment script arguments

fredo:
resolves: #757

### DIFF
--- a/scripts/deployment/rinkeby_testnets.json
+++ b/scripts/deployment/rinkeby_testnets.json
@@ -13,9 +13,11 @@
             "l1_messenger": ["OptimismL1Messenger", "0xF10EEfC14eB5b7885Ea9F7A631a21c7a82cf5D76"],
             "l2_messenger": "OptimismL2Messenger",
             "finalization_time": 604800,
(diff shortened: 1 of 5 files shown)


scripts/deployment/rinkeby_testnets.json
```diff
@@ -25,9 +27,11 @@
             "l1_messenger": ["OptimismL1Messenger", "0xBa6f1aAF3F909F208b224F3985eCda745FA9101A"],
             "l2_messenger": "OptimismL2Messenger",
             "finalization_time": 604800,
-            "claim_stake": 4700000000000000,
-            "claim_period": 3600,
-            "challenge_period_extension": 3600
+            "request_manager_arguments": {
+                "claim_stake": 0.00047,
```
istankovic, line 31: Is this correct?
```
>>> eth_utils.from_wei(4700000000000000, 'ether')
Decimal('0.0047')
```

fredo: 0.00047 is the correct number

probably the other was not intended. But it's not that important since it is on testnet.

PR #798, opened 2022-06-24
relayer: Shutdown on ppid change

### LINKED ISSUE #609: Relayer bundling
We have to make the relayer accessible for the agent.

Ideally, we can create a static binary out of the relayer, so that it doesn't need the whole node runtime when used with in the agent.

- [ ] Investigate tools for bundling node packages
- [ ] Create make target to bundle the relayer

### LINKED ISSUE #748: Kill subprocesses (relayer) when Agent dies
If the agent dies, all running relayers should be killed as well. Upon restart the agent will assume that no relayers have been started and restart all relayers.

palango:
The relayer should shut down when the agent crashes for whatever reason.
In that case the relayer process will be reattached to a process higher
in the process tree. This is used to check for the ppid, and shutdown in
case it changes.

Resolves #748 

Also adds basic bundling to the relayer.
Resolves #609

### DIFF
--- a/relayer/src/service.ts
+++ b/relayer/src/service.ts
@@ -2,6 +2,7 @@ import type { Signer } from "ethers";
 import { BaseServiceV2, validators } from "@eth-optimism/common-ts";
 import { CrossChainMessenger, MessageStatus } from "@eth-optimism/sdk";
 import type { Provider } from "@ethersproject/abstract-provider";
+import { ppid } from "process";
 
 type MessageRelayerOptions = {
   l1RpcProvider: Provider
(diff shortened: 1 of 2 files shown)


relayer/src/service.ts
```diff
@@ -105,11 +107,21 @@ export class MessageRelayerService extends BaseServiceV2<
   }
 }
 
+async function returnOnPpidChange(start_ppid: number) {
+  while (start_ppid === ppid) {
+    await sleep(500);
+  }
```
weilbith, line 113: Should we maybe log something in this case?

relayer/src/service.ts
```diff
@@ -105,11 +107,21 @@ export class MessageRelayerService extends BaseServiceV2<
   }
 }
 
+async function returnOnPpidChange(start_ppid: number) {
+  while (start_ppid === ppid) {
+    await sleep(500);
```
weilbith, line 112: Nitpick: if you want to do it without a dependency it is trivial `await new Promise(resolve => setTimeout(resolve, 500))` 🤷🏾‍♂️

weilbith (commented):
How do I test this best?

relayer/src/service.ts
palango: Good point, will add some log.

weilbith (commented):
👌🏾

weilbith (approved):

palango:
Just tested this again manually, the change of the ppid was found:
>  Relayer parent pid changed, shutting down. Old ppid: 39986, new ppid: 1

Issue #781, opened 2022-06-21
Mark sync done when events are processed


fredo:
We do not mark the sync as done when events are processed. We do it upon receiving the events

PR #799, opened 2022-06-25
agent: Sync AND process all events from both chains before processing…

fredo:
… the state (requests and claims)

fixes: #781


beamer/chain.py
```diff
@@ -153,18 +153,24 @@ def add_events(self, events: list[Event]) -> None:
...
     def _thread_func(self) -> None:
         self._log.info("EventProcessor started")
+
         chain_id = self._context.web3_l1.eth.chain_id
         poll_frequency = POLL_FREQUENCY_MAINNET if chain_id == 1 else POLL_FREQUENCY_TESTNET
 
+        # Wait until all past events are fetched and in the queue
+        # so the agent can sync to the current state
+        while not self._synced and not self._stop:
+            self._have_new_events.wait(poll_frequency)
```
palango, line 163: Do we really have to wait for `poll_frequency` here? This will always make it wait before the loop is exited, right?

palango (approved):
Looks good, left one comment.

beamer/chain.py
fredo: Yeah thats right. In a sense we dont have to wait for poll frequency here. But if we have a lower number the loop would run multiple times while we know that there wont be any new events in this time (due to the actual poll frequency). So I decided to sync them on the frequency.

PR #802, opened 2022-06-25
Fix UI flashing

compojoom:
When submitting a request the UI input was being set to disabled. Unfortunately this caused the UI to "glitch" since the input field and selector dropdown icon backgroudns were white.

By inheriting the background-color from the parent elements the UI stays the same when disabled and one no longer has the feeling that the app is broken.

Without the changes:

image

With the changes:
image

the fields are still disabled and if the user goes with the mouse over them he would receive the normal feedback that he cannot do anything in the field.

### DIFF
--- a/frontend/src/components/inputs/Selector.vue
+++ b/frontend/src/components/inputs/Selector.vue
@@ -72,6 +72,10 @@ const props = defineProps<Props>();
   @apply mt-0 pl-8 border-0 text-2xl text-teal;
 }
 
+.selector.vs--disabled .vs__search {
+  background-color: inherit;
+}
+
 .selector .vs__dropdown-option {
   @apply px-8;
 }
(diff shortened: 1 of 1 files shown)


istankovic:
Thanks @compojoom, that bug was annoying. :-)

weilbith:
Sorry, I ignored this glitch so far, because I knew we will throw away this part of the code anyway. But now as we have the fix, I don't see a reason not to just merge it. We need to ensure that #817 is not breaking it again.

weilbith (approved):

istankovic:
Is there a reason why we're not merging this?

weilbith:
I usually don't merge PRs if I'm not the author. :shrug:

istankovic:
@compojoom ?

compojoom:
So, if you receive a PR from the public - you would expect the public to merge?

image

So, if it is approved - someone from the team needs to merge it.

weilbith:
> So, if you receive a PR from the public - you would expect the public to merge?

I haven't perceived you as an external from the public. 🙂 
I would also expect the author to interact with the team if he wonders why his PR doesn't get merged. 🤷🏾

PR #812, opened 2022-06-27
Add FinalizationTimeUpdated event

### LINKED ISSUE #784: Optimize number of requests needed for L1 resolution checks
See 
- https://github.com/beamer-bridge/beamer/pull/754#discussion_r902404372
- https://github.com/beamer-bridge/beamer/pull/754#discussion_r902622035=
- https://github.com/beamer-bridge/beamer/pull/754/files/e94c0cb4e0c5225ccd66d062753116efc3ffcf89..eff09627d2eb3b7233dc0ec7a1c245ea3bec01ff#r902584553=

palango:
Resolves #784

### DIFF
--- a/beamer/events.py
+++ b/beamer/events.py
@@ -52,6 +52,12 @@ class TxEvent(Event):
 
 
 @dataclass(frozen=True)
+class FinalizationTimeUpdated(TxEvent):
+    target_chain_id: ChainId
+    finalization_time: int
+
+
+@dataclass(frozen=True)
 class RequestEvent(TxEvent):
     request_id: RequestId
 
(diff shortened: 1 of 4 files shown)


fredo (approved):
Looks good. One nitpick, in theory finalization time should never be none, because it is checked upon request creation. Rquest which do not have finalization time for the tareget chain id are forbidden.

palango:
> finalization time should never be none, because it is checked upon request creation. Rquest which do not have finalization time for the tareget chain id are forbidden.

A nice, I didn't even think about that. I'll change the code to take that into account.

Issue #810, opened 2022-06-27
agent: agent gets DoS'd when a not supported request gets claimed


fredo:
Our testnet agent suffered a DoS attack because it consumed a `ClaimMade` event where no request object exists. While for valid supported requests (by the agent) this is not possible because the events arrive in order, we introduced an assert.
For unsupported requests though it can be the case, because they are still valid requests from a contracts point of view, but the agent will discard them.

When at a later point in time, a `ClaimMade` event arrives it will hit the assert and crashes the agent.

PR #815, opened 2022-06-28
beamer: don't expect there will be a request if we encounter a claim

istankovic:
We may have ignored a request (e.g. due to an unsupported token pair) so
we need to be able to handle cases where we see claims for requests we
do not have. Note that we already do this in _handle_claim_withdrawn.

Fixes: #810

### DIFF
--- a/beamer/state_machine.py
+++ b/beamer/state_machine.py
@@ -246,7 +246,7 @@ def _handle_request_filled(event: RequestFilled, context: Context) -> HandlerRes
 def _handle_deposit_withdrawn(event: DepositWithdrawn, context: Context) -> HandlerResult:
     request = context.requests.get(event.request_id)
     if request is None:
-        return False, None
+        return True, None
 
     try:
         request.withdraw()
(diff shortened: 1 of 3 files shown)


beamer/state_machine.py
```diff
@@ -261,7 +261,10 @@ def _handle_claim_made(event: ClaimMade, context: Context) -> HandlerResult:
     # RequestCreated event must arrive before ClaimMade
     # Additionally, a request should never be dropped before all claims are finalized
     request = context.requests.get(event.request_id)
-    assert request is not None, "Request object missing upon ClaimMade event"
+    if request is None:
+        # We might not find the request because we dropped it for some reason
+        # (e.g. an invalid token pair).
+        return False, None
```
palango, line 267: With this we will process it again and again, should we return true here?

palango: Can you also fix that for `_handle_claim_withdrawn` while you're at it?

palango (approved):

PR #816, opened 2022-06-28
Contract naming

### LINKED ISSUE #809: Contracts: review & improve naming
Go over the contracts and improve naming.
For example, during contract docs work we found `ClaimWithdrawn` would be better named `ClaimStakeWithdrawn`.
The former would be confusing to someone new to the project, since claims cannot actually be withdrawn.

istankovic:
Closes #809.


palango (approved):
LGTM, but I'd squash this into two commits only

Issue #739, opened 2022-06-10
Package agent and relayer for external usage


palango:
For external parties that want to run an agent, we should provide a package that contains both the agent and required dependencies such as the relayer.

We decided that we want to package the agent and publish it on PyPI. This package should include the relayer binary (#609), so that we can ensure compatibility and users only have one package to install.

- [x] Package the relayer, handled in #609
- [x] Include relayer into the agent python package
- [x] Create CI workflow, that builds relayer and agent packages when a tag is pushed to `main`

istankovic:
Done.

PR #819, opened 2022-06-29
Agent packaging

palango:
Ref #739 

This PR adds basic packaging to the agent.
This is done by creating a bundled version of the relayer and including that in the agent package. The code to get the relayer executable is changed, so that it can find the bundles version.

Tested in a local VM:
```
Python 3.9.13 (main, Jun 23 2022, 11:12:54)
Type 'copyright', 'credits' or 'license' for more information
IPython 8.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import beamer.l1_resolution

In [2]: beamer.l1_resolution.get_relayer_executable()
2022-06-29 11:53.45 [debug    ] Found packaged relayer         relayer=PosixPath('/abc/venv/lib/python3.9/site-packages/beamer/data/bin/relayer-node16-linux-x64')
Out[2]: PosixPath('/abc/venv/lib/python3.9/site-packages/beamer/data/bin/relayer-node16-linux-x64')
```

### DIFF
--- a/.github/workflows/agent-release.yml
+++ b/.github/workflows/agent-release.yml
@@ -1,0 +1,48 @@
+name: Agent Packaging
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  package:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'false'
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+
+      - name: Bundle relayer
+        working-directory: relayer/
+        run: |
+          yarn install
+          yarn build
+          yarn add pkg
+          yarn run pkg -t node16-linux-x64 -o relayer-node16-linux-x64 build/src/service.js
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+
+      - uses: Gr1N/setup-poetry@v7
+
+      - name: Cache Python packages
+        id: cache-python-packages
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/pypoetry/virtualenvs
+          key: poetry-${{ runner.os }}-${{ hashFiles('.github/workflows/agent-release.yml', 'poetry.lock') }}
+
+      - name: Build agent package
+        run: |
+          cp relayer/relayer-node16-linux-x64 beamer/data/relayers
+          poetry install
+          poetry build
+          poetry publish -u '__token__' -p ${{ secrets.PIP_PASSWORD }}
--- a/.github/workflows/agent-release.yml
+++ b/.github/workflows/agent-release.yml
@@ -26,6 +26,7 @@ jobs:
           yarn build
           yarn add pkg
           yarn run pkg -t node16-linux-x64 -o relayer-node16-linux-x64 build/src/service.js
+          yarn run pkg -t node16-macos-x64 -o relayer-node16-macos-x64 build/src/service.js
 
       - uses: actions/setup-python@v4
         with:
@@ -43,6 +44,7 @@ jobs:
       - name: Build agent package
         run: |
           cp relayer/relayer-node16-linux-x64 beamer/data/relayers
+          cp relayer/relayer-node16-macos-x64 beamer/data/relayers
           poetry install
           poetry build
           poetry publish -u '__token__' -p ${{ secrets.PIP_PASSWORD }}
--- a/beamer/cli.py
+++ b/beamer/cli.py
@@ -103,7 +103,7 @@ def main(
 ) -> None:
     beamer.util.setup_logging(log_level=log_level.upper(), log_json=False)
 
-    if get_relayer_executable() is None:
+    if not get_relayer_executable().exists():
         log.error("No relayer found")
         sys.exit(1)
 
--- a/beamer/cli.py
+++ b/beamer/cli.py
@@ -1,4 +1,5 @@
 import signal
+import sys
 from pathlib import Path
 from typing import Optional
 
@@ -9,7 +10,7 @@ import beamer.contracts
 import beamer.util
 from beamer.agent import Agent
 from beamer.config import Config
-from beamer.l1_resolution import relayer_executable_exists
+from beamer.l1_resolution import get_relayer_executable
 from beamer.typing import URL
 from beamer.util import account_from_keyfile
 
@@ -102,8 +103,9 @@ def main(
 ) -> None:
     beamer.util.setup_logging(log_level=log_level.upper(), log_json=False)
 
-    # TODO: use return value and exit if relayer is not available
-    relayer_executable_exists()
+    if get_relayer_executable() is None:
+        log.error("No relayer found")
+        sys.exit(1)
 
     account = account_from_keyfile(keystore_file, password)
     log.info(f"Using account {account.address}")
--- a/beamer/l1_resolution.py
+++ b/beamer/l1_resolution.py
@@ -1,6 +1,8 @@
-import shutil
 import subprocess
+import sys
 
+from pathlib import Path
+from typing import Optional
 import structlog
 from hexbytes import HexBytes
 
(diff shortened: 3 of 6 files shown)


pyproject.toml
```diff
@@ -2,8 +2,12 @@
 name = "beamer"
```
palango, line 2: Should we call this `beamer-agent`?

istankovic: > Should we call this `beamer-agent`?

Good question. There's already https://pypi.org/project/beamer/ which also makes it tough to just leave this as "beamer".
Let's see what we provide as part of this package:

- the `beamer-agent` executable
- the relayer executable (albeit in an "implementation-detail" manner)
- the `beamer` Python package

In the future there might be more things in there.
Considering all of this, I would go with `beamer-bridge`.
(Also, it's fine if the name of our PyPI package differs from the Python package we ship. See https://pypi.org/search/?q=bridge&o=.)

beamer/cli.py
```diff
@@ -102,8 +102,8 @@ def main(
 ) -> None:
     beamer.util.setup_logging(log_level=log_level.upper(), log_json=False)
 
-    # TODO: use return value and exit if relayer is not available
-    relayer_executable_exists()
+    if get_relayer_executable() is None:
+        log.warn("No relayer available")
```
istankovic, line 106: I'm thinking here, the agent should probably refuse to start if there's no usable relayer.
Because the relayer is critical for L1 resolution and if we don't have it...
So I would actually prefer that we log an error here and immediately exit with an error code.

palango: I totally agree, I most mostly worried about creating more problems for the testnet deployments.

.github/workflows/agent-release.yml
```diff
@@ -0,0 +1,57 @@
...
+          python-version: 3.9
+
+      - uses: Gr1N/setup-poetry@v7
+
+      - name: Cache Python packages
+        id: cache-python-packages
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/pypoetry/virtualenvs
+          key: poetry-${{ runner.os }}-${{ hashFiles('.github/workflows/backend.yml', 'poetry.lock') }}
```
istankovic, line 44: Should we hash this file (`agent-release.yml`) instead of `backend.yml`?

pyproject.toml
palango: I like ` beamer-bridge`! 👍

.github/workflows/agent-release.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Build agent package
+        run: |
+          cp relayer/relayer-node16-linux-x64 beamer/data/bin
+          poetry install
+          poetry build
+          # poetry publish
```
istankovic, line 51: Extra comment?

beamer/l1_resolution.py
```diff
@@ -12,22 +15,33 @@
...
-        return True
+        return Path(maybe_relayer)
 
     log.warning("Relayer executable not found in PATH")
-    return False
+    return None
 
 
 def run_relayer_for_tx(l1_rpc: URL, l2_rpc: URL, privkey: str, tx_hash: HexBytes) -> None:
+    relayer = get_relayer_executable()
```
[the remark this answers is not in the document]
palango: I thought more about https://github.com/beamer-bridge/beamer/pull/819/files#r909561849 , then this will never happen.

.github/workflows/agent-release.yml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          node-version: '16'
+
+      - name: Bundle relayer
+        working-directory: relayer/
+        run: |
+          yarn install
+          yarn build
+          yarn add pkg
+          yarn run pkg -t node16-linux-x64 -o relayer-node16-linux-x64 build/src/service.js
```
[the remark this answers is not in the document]
palango: I felt it's clearer this way, where we first bundle the relayer, and then build the pyhton package

beamer/l1_resolution.py
```diff
@@ -12,22 +15,33 @@
...
+        if path.exists():
+            log.debug("Found packaged relayer", relayer=path)
+            return path.resolve()
+
+    # Otherwise check if the relayer is available in PATH
     maybe_relayer = shutil.which(_RELAYER_EXECUTABLE)
     if maybe_relayer is not None:
         log.debug("Relayer found in PATH", relayer=maybe_relayer)
-        return True
+        return Path(maybe_relayer)
```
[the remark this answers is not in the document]
istankovic: > Currently yes, at least for mac users.

What's the issue with mac?

beamer/l1_resolution.py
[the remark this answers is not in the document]
istankovic: Yeah, I'm thinking here we should behave the same as if we couldn't find the relayer on startup: log an error and terminate the agent.

In theory, only a broken installation (e.g. somebody deletes the relayer while the agent is running) could result in this.
(Also -- what about mac situation here?)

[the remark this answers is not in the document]
palango: done

palango:
All done, ready for final review.

beamer/tests/agent/test_cli.py
```diff
@@ -38,6 +40,23 @@ def _generate_deployment_dir(output_dir, root, contracts):
...
+    executable = target / _RELAYER_NAMES[sys.platform]
+    if executable.exists():
+        return
+
+    executable.write_text("text")
+
+    yield
+
+    executable.unlink()
+
```
istankovic, line 57: Nitpick: I was thinking more along the lines of

```python
 def setup_relayer_executable():
    path = beamer.l1_resolution.get_relayer_path()
    if path.exists():
       return

    path.parent.mkdir(parents=True, exist_ok=True)
    path.write_text("")
    yield
    path.unlink()
```
where we don't need to know the explicit location of the relayer nor the platform stuff. This would also
avoid importing private items.

.github/workflows/agent-release.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: |
+          cp relayer/relayer-node16-linux-x64 beamer/data/relayers
+          poetry install
+          poetry build
+          poetry publish -u '__token__' -p ${{ secrets.PIP_PASSWORD }}
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: agent
```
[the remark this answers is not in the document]
palango: Ah good point. Do we even want to store the artifact here?

beamer/l1_resolution.py
```diff
@@ -9,25 +11,37 @@
...
-    maybe_relayer = shutil.which(_RELAYER_EXECUTABLE)
-    if maybe_relayer is not None:
-        log.debug("Relayer found in PATH", relayer=maybe_relayer)
-        return True
+    bin_dir = Path(__file__).parent.joinpath("data/relayers/")
+    try:
+        maybe_relayer = bin_dir.joinpath(_RELAYER_NAMES[sys.platform])
+    except KeyError:
+        log.warning("Unsupported platform", platform=sys.platform)
+        return None
```
istankovic, line 25: Here I'd actually `raise RuntimeError("Unsupported platform ...` and let the agent automatically exit. Meaning we don't need to return `None`.

.github/workflows/agent-release.yml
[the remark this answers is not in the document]
istankovic: I'm not even sure whether we have a use case for that, so unless we can make use of those artifacts, I think we should not store them.

istankovic (approved):
Great work!
A couple of comments left, to improve internals.

.github/workflows/agent-release.yml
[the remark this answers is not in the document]
istankovic: Removed.

beamer/l1_resolution.py
```diff
@@ -9,25 +11,37 @@
 
 log = structlog.get_logger(__name__)
 
-_RELAYER_EXECUTABLE = "beamer-relayer"
+_RELAYER_NAMES = {"linux": "relayer-node16-linux-x64", "darwin": "relayer-node16-macos-x64"}
 
 
-def relayer_executable_exists() -> bool:
-    """Checks if the relayer executable is in PATH, return `False` if not"""
+def get_relayer_executable() -> Optional[Path]:
```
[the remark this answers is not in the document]
istankovic: Done.

beamer/l1_resolution.py
istankovic: Done.

PR #833, opened 2022-07-05
ci: Fix relayer packaging

palango:
Try to fix the current CI problems

### DIFF
--- a/.github/workflows/agent-release.yml
+++ b/.github/workflows/agent-release.yml
@@ -4,6 +4,9 @@ on:
   push:
     tags:
       - '*'
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
 
 jobs:
   prepare-package:
@@ -12,11 +15,16 @@ jobs:
   package:
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    needs: prepare-package
 
     steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'false'
+
       - uses: actions/download-artifact@v3
         with:
-          name: source-with-relayers
+          name: relayers
 
       - uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/agent-release.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     tags:
       - '*'
+  pull_request:
```
istankovic, line 7: This is going to publish a Python package on every push, no?

palango: Yes, this is just for testing.

istankovic (approved):

.github/workflows/agent-release.yml
```diff
@@ -12,11 +15,16 @@ jobs:
   package:
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    needs: prepare-package
```
istankovic, line 18: Ah, that was the needed incantation... :+1:

.github/workflows/prepare-package.yml
```diff
@@ -21,6 +22,6 @@ jobs:
 
       - uses: actions/upload-artifact@v3
         with:
-          name: source-with-relayers
-          path: .
-          retention-days: 1
+          name: relayers
+          path: relayer/relayer-node16-*
```
istankovic, line 26: Oh yes, that's exactly what I had in mind, should be plenty quick now!

.github/workflows/prepare-package.yml
```diff
@@ -21,6 +22,6 @@ jobs:
...
+          retention-days: 7
```
istankovic, line 27: Any reason we like to keep the relayers for 7 days?

docker/Dockerfile.agent
```diff
@@ -2,7 +2,7 @@ FROM python:3.9
 
 COPY . /src
 RUN mkdir -p /src/beamer/data/relayers
-COPY ./relayer/relayer-node16-*-x64 /src/beamer/data/relayers/
+COPY ./relayer-node16-*-x64 /src/beamer/data/relayers/
```
istankovic, line 5: How can this work? Does local invocation of `make container-image` still work?

istankovic: > Why would we ever run this locally?

I have been using this to test our Docker images. Far easier than to mess with github CI.
Also, it is not just about us -- this is intended to be usable by other developers too, and they may well expect this kind of thing to work. Think of the way that we are using Optimism's Docker infrastructure.
Anyone should be able to generate their own, possibly customised, Docker images. Why should we restrict this only to our CI?

palango: Ok, tested it again. works both locally and on ci.

PR #834, opened 2022-07-05
agent: Fix flakyness of `test_challenge_5[False]`

palango:
Sometimes this seems to take a while, so let's increase the timeout.


istankovic (approved):
Awesome, man. :+1:

PR #835, opened 2022-07-06
A few fixes

istankovic:
Found during a recent deployment.

### DIFF
--- a/scripts/call_contracts.py
+++ b/scripts/call_contracts.py
@@ -314,19 +314,16 @@ def whitelist(
     "protocol-fee-ppm",
     type=int,
     required=True,
-    help="Protocol fee percentage described in parts per million",
 )
 @click.argument(
     "lp-fee-ppm",
     type=int,
     required=True,
-    help="LP fee percentage described in parts per million",
 )
 @click.argument(
     "min-lp-fee",
     type=int,
     required=True,
-    help="Minimum LP fee in Wei",
 )
 @pass_args
 def update_fee(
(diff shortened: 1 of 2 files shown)


scripts/call_contracts.py
```diff
@@ -314,19 +314,16 @@ def whitelist(
     "protocol-fee-ppm",
     type=int,
     required=True,
-    help="Protocol fee percentage described in parts per million",
```
palango, line 317: Why remove these?

scripts/e2e-test.py
```diff
@@ -28,7 +28,8 @@ def main() -> None:
     request_amount = 123
 
     token = l2_contracts["MintableToken"]
-    token.functions.mint(deployer.address, request_amount).transact()
+    tx_hash = token.functions.mint(deployer.address, request_amount).transact()
+    web3.eth.wait_for_transaction_receipt(tx_hash)
```
palango, line 32: Do we really need this? There's the approve afterwards, and as the nonces increase this should wait implicitly.

scripts/call_contracts.py
istankovic: > Why remove these?

It's literally in the commit message:
```
scripts: call_contracts: click.argument does not support help text
```
:smile: 

Without the fix, you get an exception at runtime when trying to use that command.

scripts/e2e-test.py
istankovic: > Do we really need this? There's the approve afterwards, and as the nonces increase this should wait implicitly.

I thought it is just good practice. Theoretically, it is possible that the mint tx is not included in a block, and the approve is then reverted, which is wasteful, right?

palango (approved):

PR #836, opened 2022-07-06
Prepare for release test

palango:
Still need to add version number logging to the agent

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -2,7 +2,10 @@ name: Build beamer-agent container image
 
 on:
   push:
-    branches: [ main ]
+    branches:
+      - 'main'
+    tags:
+      - 'v*'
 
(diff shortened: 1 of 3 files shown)


beamer/cli.py
```diff
@@ -103,6 +104,9 @@ def main(
 ) -> None:
     beamer.util.setup_logging(log_level=log_level.upper(), log_json=False)
 
+    version = pkg_resources.get_distribution("beamer-bridge").version
```
istankovic, line 107: Nitpick: since we rely on a recent-enough version of Python, I think we should use `importlib.metadata.version` here, that stuff is intended to replace functionality offered by the old `pkg_resources` package (see https://docs.python.org/3/library/importlib.metadata.html).

istankovic (approved):

beamer/cli.py
palango: How do you remember such things 🤯

PR #839, opened 2022-07-07
frontend: remove floating-vue dependency

weilbith:
We don't use any of its components/features anymore. The probably last one removed was the tooltip that got replaced with a custom component. Removing this also reduces final bundled package size.


manuelwedler (approved):
LGTM

PR #849, opened 2022-07-12
Add cli argument to start agent via config file

### LINKED ISSUE #653: Introduce an agent config file
In order to make the agent configuration easier and pave the way to multiple chain pairs support, we need to introduce a configuration file for the agent:

```bash
$ beamer-agent -c agent.conf
```

It should be possible to fully configure an agent just by using the configuration file i.e. no additional command line options should be necessary.

Additionally, we need to allow the user to configure multiple chain pairs:
```toml

log-level = "debug"                                                                     
deployment-dir = "deployments/rinkeby"                                                  
fill-wait-time = 120                                                                    
source-chain = "metis"
target-chain = "boba-rinkeby"
                                                                                        
[account]                                                                               
path = "/keys/UTC--2022-03-09T20-51-23.967Z--0x6626079bcf8c3241b082c73b74dfea46cefa4f02"
password = "foobar"                                                                     
                                                                                        
[metrics]                                                                               
prometheus-port = 9101                                                                  
                                                                                        
[chains.l1]                                                                             
rpc-url = "https://api-rinkeby.etherscan.io"                                            
                                                                                        
[chains.metis]                                                                          
rpc-url = "https://stardust.metis.io/?owner=588"                                        
                                                                                        
[chains.boba-rinkeby]                                                                   
rpc-url = "https://rinkeby.boba.network/"                                               
                                                                                        
[tokens]                                                                                
TST = [                                                                                 
    ["boba-rinkeby", "0x2644292EE5aed5c17BDcc6EDF1696ba802351cf6"],                     
    ["metis", "0xAcF5e964b76773166F69d6E53C1f7A9114a8E01D"]                             
]                                                                                       
                                                                                        
USDC = [                                                                                
    ["boba-rinkeby", "0x1a65113Fb92916EF0D3043D651b469b653763F16"],                     
    ["metis", "0x6bCE0F297a204E1374860E0259EC31047a87B50F"]                             
]                                                                                       
```
The above is in TOML format, to make it easier for humans.
Note that the above also includes the token match data, which obviates the need for a separate token match file.
It also has the added benefit of naming the tokens so that implementing a specific policy (e.g. forbidding a token on a specific chain) is made easier.

How we might go about implementing this issue:
- [ ] add a new command line option `-c`, `--config`
- [ ] have a default configuration in the agent that can be overridden by the config file and command line options, in that order
- [ ] replace the token match file with the `tokens` table from the config file

cducrest:
Add example config file. The possible values in the config file match what was previously the arguments of the cli
related: https://github.com/beamer-bridge/beamer/issues/653

I don't get how we should have `chains.metis` and `chains.boba-rinkeby` in the config as described in the issue, because l2a and l2b are treated differently in the agent. I don't get how the token matching should use names instead of chain ids as described in the issue, should I replace the chain ids there and then add a new config field for mapping names to chain id? I don't feel that will gain much.

We can have a call if you want to explain to me better what is needed to close #653


cducrest:
I tried to improve the code based on your comment @istankovic , and added source and target chain name in the config. 

Replacing chain id with names in the token mapping of the config is a little more involved because the `TokenMatchChecker.is_valid_pair()` method uses chain ids (probably rightfully so, since they take them from events). So we need to somehow instantiate the TokenMatchChecker with chain names and later replace the names with the chain ids once connected to the chain, or instantiate it later altogether. I find this challenging while maintaining the legacy `token_match_file` compatibility.

palango (approved):

PR #853, opened 2022-07-14
frontend: New selector component

### LINKED ISSUE #542: frontend: Add logos of rollups in selectors
In Olympe's design, there are the logos of the rollups in the selector fields. We should add them. I think only the config needs to get adapted for that. The selector component already supports displaying an image.

### LINKED ISSUE #660: frontend - custom selection input component
Have a look at the [mock-ups](https://www.figma.com/file/Tx5V8SUrD1HrvDVqX7iEjn/Beamer-App?node-id=0%3A1) for details.

A small PR should do:
  - define a custom input component to select from a list
  - add a search input on top of the list
  - emit a close event when a value got selected
  - make list scrollable
  - ensure that component draws 100% over the whole parent (form) component

manuelwedler:
Closes #660
Closes #542 

The selector uses a new overlay to render all options in a list as it is in the design.

The selector still uses the old SelectorOption<T>. This is because we need to have a uniform way to display a label and an image for any type T. I also considered to only make the output of the modelValue the type T. But this would lead to some problems: We want to be able to set the selector value from the outside, but to get the correct corresponding SelectorOption<T> for displaying it from the props.options we would need to find the value inside the options. This is not possible for an arbitrary type T or would need some interface to allow for a uniform filter for any T that implements that interface. As this is much more effort than the value it would bring, I decided to keep the SelectorOption as the modelValue. Also, the SelectorOption is still in the form.ts file, because we cannot import it from a .vue file.

Lastly, this adds images for the rollups to display in the selector. When we later add images to config of real tokens (in the config files), the images would also be displayed in the selector. 

As in the design all fields in the dialog are now right aligned.


palango (approved):

weilbith:
When I access the `frontend-dev` version, I have the following issues:
  - my selection is not accepted, should mean after making a selection, the input field is still empty
  - the icons are broken and look like on the screenshot below
  
  
image

manuelwedler:
Thanks for noticing @weilbith ! 
1. I see, the issue is that you are not connected to a provider and the source chain selector now only displays the connected chain of the provider. 3 options to solve this:
   - Disable the source selector before the user connects
   - Let the user select a value (when not connected) and after connecting switch the chain after.
   - Let the user select a value (when not connected) and after connecting overwrite the selected source chain with the connected the chain of the provider.
2. Probably the image url is not correct in the deployed version.
Will look into these. Also created an issue #892 

Any opinion on 1?

weilbith:
I would choose 1b). But honestly I think this is a question to the UX responsible people. Maybe just implement a solution and wait if people complain. 😜

PR #855, opened 2022-07-15
frontend: fix issue of a too low token allowance

### LINKED ISSUE #851: Frontend: Transfer history lazy loaded only when label in view
Screencast from 14 07 2022 13 33 29

#### Actual behavior
When in the transfer history and you scroll to the bottom, the upper elements won't load until the label of the transfer group is in the view.

#### Expected behavior
The transfers should load immediately when they would be in the view.

Make a couple of transfers. Let the some time pass so they are grouped by older and last week. Scroll to the bottom. Slowly scroll up again.

### LINKED ISSUE #854: Setting allowance limit
#### System information
Currently it seems that the token allowance is being set at the total amount of tokens one has. 

I wanted to test the bridge with 3 test tokens. I tried to send 1 - the allowance for the token was set to 3. Later when I added more tokens to be able to pay for the transfer (one need 5 tokens) it was not possible to execute the transfer since the allowance limit was set to 3 and I was basically trying to transfer 6.

#### Expected behavior
The token limit should be set to a higher enough number and maybe an option to change it should be available somewhere in the UI.

weilbith:
Fixes #854

How to verify the bug got fixed:
- use first an old version of the frontend
- use an account not used yet for Beamer (in fact that has no token allowance for the request manager contract)
- transfer some ETH to this account
- connect the app to the account
- mint some test tokens for this account
- send all tokens to a different account accept 2
- try to make a transfer of 1 TST
  - this will set the allowance to 2, as currently it is 0, is needs 1 and sets it to the current token balance of the user
  - the allowance step should be green
  - the request transaction step should fail (error in UI is `insufficient balance`)
  - the error in the console should reveal the contract reverted message because the token allowance is too low
  - this could be repeated, the allowance step gets skipped, because the allowance of 2 is greater than 1

- use this fixed version now
- send back all the TST tokens to the active account
- try to make another transfer of 1 TST
- it should now increase the allowance it requires a minimum of 6 (1TST transfer plus 5 TST fees)
- request transaction should succeed

### DIFF
--- a/frontend/src/actions/transfers/transfer.ts
+++ b/frontend/src/actions/transfers/transfer.ts
@@ -226,11 +226,13 @@ export class Transfer extends MultiStepAction implements Encodable<TransferData>
       throw new Error('Missing wallet connection!');
     }
 
+    const totalAmount = this.sourceAmount.uint256.add(this.fees.uint256);
+
     await ensureTokenAllowance(
       signer,
       this.sourceAmount.token.address,
       this.sourceChain.requestManagerAddress,
-      this.sourceAmount.uint256,
+      totalAmount,
     );
   }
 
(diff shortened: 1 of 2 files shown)


weilbith:
Alright, I got the review by Daniel and he tested it. But as he is not a team member he can't give an "official" approval on GitHub. So I'll use my administrator permissions to merge this anyway.

PR #861, opened 2022-07-19
beamer: util: attempt to send a transaction up to 5 times by default

istankovic:
This should help with "nonce too low" issues.


beamer/util.py
```diff
@@ -41,10 +43,19 @@ def transact(
...
     timeout: float = 120,
     poll_latency: float = 0.1,
+    attempts: int = 5,
     **kwargs: Any,
 ) -> Any:
     try:
-        txn_hash = func.transact(cast(Optional[TxParams], kwargs))
+        while attempts > 0:
+            try:
+                txn_hash = func.transact(cast(Optional[TxParams], kwargs))
```
palango, line 52: should there be some wait before the retry?

fredo: just to get clarity. Can you confirm that this will actually use a new nonce everytime it is called?

istankovic: Good question. Do you have any suggestions?

istankovic: > just to get clarity. Can you confirm that this will actually use a new nonce everytime it is called?

AFAICT, any time you call `transact`, this will actually result in a call to `eth_getTransactionCount` which is not cached with our current setup so it should be fine. This is done by web3 middleware. If you use `construct_latest_block_based_cache_middleware`, this is _not_ going to be the case, so one needs to be careful.

istankovic: I added a random wait between 0.5 and 3 seconds between each attempt, meaning the total run time for `transact` (with the default settings) would be 12 seconds in the worst case. We probably don't need to optimise here (yet), but it's good to have in mind since this will actually block the event loop.

PR #862, opened 2022-07-19
Trim logging

istankovic:
We now log only when we have something interesting to report, like changing states of requests or claims.
This should significantly reduce storage requirements when using debug log level.
Also, some of the messages have been promoted to info log level to give them more visibility since most people are not really expected to run agents with debug logging turned on all the time.


palango (approved):
Awesome!

PR #865, opened 2022-07-20
agent: drop RequestFilled event if request is in state withdrawn

fredo:
fix for removing unusable RequestFilled events. Happens especially during start up and fethcing older events


istankovic (approved):
Looking good, MAN! :smile:

PR #866, opened 2022-07-21
frontend: Fix chain switching logic and keep source chain selector in sync

### LINKED ISSUE #852: Frontend: Cancelling chain switch lets user send tx to wrong chain
#### Actual behavior
A change in the "From" selector asks MetaMask to switch the chain. When the user cancels switching the chain in MetaMask, our "from" selector shows the new chain while the user is still connected to the old one. After that they may make a transfer, which will fail. The reason in the transfer history is "Insufficient balance!".

The whole of the "switchChain" functions of our providers can be reworked in my opinion. It returns either a boolean or a null, or may throw an error. It is not clear in which case the error is thrown and in which the null is returned. In the successful case a true is returned, but it never returns false. This function is really confusing and the expected behavior is not clear to me.

#### Expected behavior
The selector should be in sync with the connected chain from MetaMask.

Select another source chain. Cancel the switch in MetaMask. Try to make a transfer.

manuelwedler:
Closes #852 

Not related to this PR but I also fixed the config file for production which wasn't parsable  since my last PR.

### DIFF
--- a/frontend/src/composables/useChainSelection.ts
+++ b/frontend/src/composables/useChainSelection.ts
@@ -1,16 +1,11 @@
 import type { Ref } from 'vue';
 import { computed } from 'vue';
 
-import type { IEthereumProvider } from '@/services/web3-provider';
 import type { ChainConfigMapping } from '@/types/config';
 import type { Chain } from '@/types/data';
 import type { SelectorOption } from '@/types/form';
 
-export function useChainSelection(
-  provider: Ref<IEthereumProvider | undefined>,
-  chains: Ref<ChainConfigMapping>,
-  ignoreChains: Ref<Chain[]>,
-) {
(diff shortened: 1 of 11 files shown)


frontend/src/composables/useChainSelection.ts
```diff
@@ -27,7 +27,7 @@ export function useChainSelection(
     if (provider.value && chain.identifier !== provider.value.chainId.value) {
       try {
         const isSuccessfulSwitch = await provider.value.switchChain(chain.identifier);
-        if (isSuccessfulSwitch === null) {
+        if (isSuccessfulSwitch === false) {
```
palango: Wouldn't it be better to write `if (!isSuccessfulSwitch)`?

palango (approved):
One nit, otherwise looks good.

frontend/src/composables/useChainSelection.ts
manuelwedler: yes, definitely better! thx

PR #869, opened 2022-07-21
beamer: chain: log the duration (in milliseconds) of each event loop …

istankovic:
…iteration

This should help us notice problems with the event loop.

### DIFF
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -177,6 +177,7 @@ class EventProcessor:
         iteration = 0
         created_events: list[Event] = []
         while True:
+            t1 = time.time()
             with self._lock:
                 events = self._events[:]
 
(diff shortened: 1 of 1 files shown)


palango (approved):

beamer/chain.py
```diff
@@ -200,11 +201,13 @@ def _process_events(self) -> None:
                 del self._events[: len(events)]
                 self._events.extend(unprocessed)
 
+            t2 = time.time()
             self._log.debug(
                 "Finished iteration",
                 iteration=iteration,
                 any_state_changed=any_state_changed,
                 num_events=len(self._events),
+                time=round((t2 - t1) * 1e3, 3),
```
palango, line 210: maybe call this runtime or duration.

PR #874, opened 2022-07-21
frontend: Fix transfer history lazy loading

### LINKED ISSUE #851: Frontend: Transfer history lazy loaded only when label in view

manuelwedler:
Closes #851 

Every transfer group label and every transfer element is now wrapped in a LazyWrapper itself. This fixes the problem that a user needed to scroll to the label to see the whole group. Also, this removes the overhead of every transfer being added to the DOM when the respective group is displayed. As all transfers go eventually into the "older" group, this could have been an performance issue in my opinion.

I would like to have the opinion of @weilbith next week when he is back on this PR. This bug is nothing critical, so this should be fine. I am just wondering how he intended the LazyWrapper to be uses, because my version will create an extra div for every element.

### DIFF
--- a/frontend/src/components/TransferHistory.vue
+++ b/frontend/src/components/TransferHistory.vue
@@ -16,27 +16,22 @@
         <span>Nothing here yet.</span>
       </div>
 
-      <LazyWrapper
-        v-for="group of groupedAndSortedTransfers"
-        :key="group.label"
-        :root-element="listElement"
-        :threshold="0.0"
-        class="mb-12"
-        data-test="transfer-group"
-      >
-        <div v-if="group.transfers.length > 0" class="text-2xl text-center">
-          {{ group.label }}
-        </div>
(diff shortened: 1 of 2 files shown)


weilbith:
This is exactly what I had in mind too. When I implemented this first, I wasn't careful enough and the history was just not a couple of days old. But after I saw this bug myself I also thought that this is the issue and the approach how to solve it. So great job from my perspective. 👍🏾

weilbith (approved):

PR #888, opened 2022-07-27
frontend: Extend input field implementation

### LINKED ISSUE #887: Frontend: Alphabet character input inside request form breaks app
https://user-images.githubusercontent.com/18466333/181225686-fa822ba5-b136-4ac7-ba68-2e5a4d73bd30.mp4

#### Expected behavior
Alphabet character input should be forbidden inside the amount input fields to prevent exposure to unforeseen errors. 
By forbidding alphabet character input we will satisfy both the underlying amount parser implementation, which expects a strictly numeric value provided as `string` and also mobile UX design (numeric keyboard pops out on input focus instead of normal keyboard).

#### Actual behavior
User is allowed to enter alphabet characters inside the token amount inputs. 
Due to the implemented amount parsing which, in fact, expects only numeric values as input the app breaks and becomes unresponsive to user interaction.

- Select a source chain and token. 
- Start entering alphabet characters instead of numeric characters inside the amount input.
- App becomes unresponsive to user interaction

GabrielBuragev:
The implementation of the TextInput component has been adjusted to accept other type attributes rather then having a static `text` attribute. Therefore its name has changed as well since it has to serve more purposes than being a text input only.

This fixes #887 - an issue that makes the app unresponsive/unusable after the user has selected the `token` & `sourceChain` and then types an alphabet character inside the `amount` input field (by accident or intentionally).
It also delivers better UX on mobile devices by showing the numeric keyboard to the user for the number inputs.

The Input component still receives/emits the value as string in order to not break current code where `TokenAmount` parsers are being used for serializing the request payload.
Can be changed in future based on needs.

Tests are added for all the basic test cases for the Input component.


frontend/tests/unit/components/inputs/Input.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { mount } from '@vue/test-utils';
+
+import Input from '@/components/inputs/Input.vue';
+
+const createWrapper = (props = {}, attachToBody = false) => {
```
manuelwedler, line 5: For consistency, I rather would like to see a typed options object as a parameter for this function. Just as the `createWrapper` of the `TextInput` tests looked like with an additional `type` option. So everything should be in the options object, also the `attachToBody` parameter. Most important is that we had typings for the options' properties.

manuelwedler (changes requested):
Wow, I didn't know that it is possible to have a HTML number input which outputs JS Strings. I thought the input field would be limited to the maximum JS number. Very good to know! Thanks a lot for the PR.
There is just one minor comment to keep consistency. And please run `yarn lint:fix` before pushing. There is one linting issue. I recommend to create a fixup commit to make these changes to keep the history clean. 
Just in case you don't know about the [fixup](https://blog.sebastian-daschner.com/entries/git-commit-fixup-autosquash) commits, the process is the following: you push your fixup commit(s), I will review again, after that you can autosquash, force push and merge.

frontend/tests/unit/components/inputs/Input.spec.ts
GabrielBuragev: Thanks for reminding me that this pattern is the preferred one for our tests. Kinda missed it :see_no_evil:.
Will keep that in mind next time i am writing tests!

PR #894, opened 2022-08-01
frontend: Add target address prefill function

### LINKED ISSUE #878: Having the "Target Address" be pre-filled so users do not need to copy and paste
Feedback from several community members

Many have been saying:

since most of the time you are transferring to the same wallet address, it would be better to have the "target address" to be pre-filled

@istankovic - feedback from one of the users on discord

GabrielBuragev:
Fixes #878

The target address input inside the request form is now reacting
and prefilling the value for the user on wallet address change
(connect/disconnect).
When the user manually entered a different address than his currently
connected wallet address the input will not be touched until
input value was cleared.

### DIFF
--- a/frontend/src/components/RequestDialog.vue
+++ b/frontend/src/components/RequestDialog.vue
@@ -157,7 +157,10 @@ function parseTokenFromChainConfiguration(
 
 function resetForm() {
   requestSource.value = EMPTY_SOURCE_DATA;
-  requestTarget.value = EMPTY_TARGET_DATA;
+  requestTarget.value = {
+    ...EMPTY_TARGET_DATA,
+    toAddress: requestTarget.value.toAddress === signerAddress.value ? signerAddress.value : '',
+  };
 }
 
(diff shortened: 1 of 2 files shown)


manuelwedler (commented):
The changes look good!
Just one addition we might want to add. Not resetting to the signer address makes definitely sense in the scenario which you explained in the issue. But we could still leave the address (in the reset function) if it is the same as the signer address. This simple check would remove one step for second time users. What do you think?

manuelwedler (approved):
Perfect!

istankovic:
This seems to be blocked by a fixup commit.

@GabrielBuragev we typically use `git rebase -i` to clean up commits before we finalize PRs. That is not to say that everything needs to be squashed into a single giant commit, but rather that the history should be looking nice and clean.

GabrielBuragev:
@istankovic thank you for informing me. Still haven't rebased so Manuel has a clear overview of the steps for better reviewing experience. 
Will rebase now since he finished reviewing.

PR #901, opened 2022-08-03
frontend: Implement debounced fee calculation & adjust UI

### LINKED ISSUE #895: Frontend: Inconvenient fee amount apperance
[rec.webm](https://user-images.githubusercontent.com/18466333/182129170-78933425-d0ca-4328-b46a-04158d6c54ca.webm)

I noticed a "choppy" reaction time feeling inside the request form when the app is trying to calculate the request fees based on user's input.

Currently the fee calculation for the user's entered `token amount` is done once per character input. 
That behavior comes from inside the composable which is responsible for calculating the fees by watching the `token amount` value inside a `watch` function which causes that event to be fired once for each input action.

A solution could be implemented by using [debounced](https://lodash.com/docs/4.17.15#debounce) function. Current `useRequestFee` composable could be extended to support debounced functions.
Probably, there are also other solutions. Therefore, this can be discussed further.

**IMO** the request fee value should turn into a "spinner" indicator until the calculating operation finished. Its always nice to show feedback to the user that there is an ongoing process, gives the app more interactive feeling.

Addressing issue #741 as it significantly helps in reducing stress on RPC endpoints.

GabrielBuragev:
fixes #895

Extended implementation of `useRequestFee` composable to support
debounced fee calculation in order to provide better UX and to tackle
the "choppy" reaction when the fees are being updated on the UI.
By adding this we also reduce the stress on the RPC endpoints significantly.

As a side effect, the Spinner component was refactored and another composable
`useDebouncedTask` was created in order to split the implementation logic
for the debounced functionality.

The Spinner component was refactored to provide better documentation and
configurability. It is now used in the request form to indicate to the user
that a fee recalculation is ongoing whenever he changes the amount of tokens
for sending.

Added tests accordingly for the `useRequestFee` composable, `Spinner` component
& `useDebouncedTask` composable.


frontend/src/components/RequestSourceInputs.vue
```diff
@@ -15,9 +15,7 @@
             :disabled="!faucetAvailable"
             @click="runFaucetRequest"
           >
-            <div v-if="faucetRequestActive" class="h-5 w-5">
-              <spinner></spinner>
-            </div>
+            <spinner v-if="faucetRequestActive" size="6" border="2"></spinner>
```
GabrielBuragev, line 18: Looks better with `size` of 6

frontend/src/composables/useRequestFee.ts
```diff
@@ -1,22 +1,28 @@
...
 
+import { useDebouncedTask } from '@/composables/useDebouncedTask';
 import { getRequestFee } from '@/services/transactions/request-manager';
 import { TokenAmount } from '@/types/token-amount';
 
 export function useRequestFee(
   rpcUrl: Ref<string | undefined>,
   requestManagerAddress: Ref<string | undefined>,
   requestAmount: Ref<TokenAmount | undefined>,
+  debounced?: boolean | undefined,
```
manuelwedler, line 12: Nitpick: for an optional parameter you don't need to add the undefined type.

frontend/src/composables/useRequestFee.ts
```diff
@@ -32,11 +38,24 @@ export function useRequestFee(
...
       error.value = errorMessage ?? 'Unknown failure.';
     }
+
+    loading.value = false;
   };
 
-  watch([rpcUrl, requestManagerAddress, requestAmount], updateRequestFeeAmount, {
-    immediate: true,
-  });
+  const cb = computed(() =>
```
manuelwedler, line 45: There is nothing reactive inside, so you don't need a computed property here. I guess that happened because you had the debounced parameter as a ref before. I would just move the call down inside the watch. 

Also, I'd like to mention that I think it is better to have full names (callback instead of cb), because even if in most cases the meaning is clear it might not be clear to everyone. So it's just a small difference but provides better readability imo.

frontend/src/composables/useRequestFee.ts
```diff
@@ -32,11 +38,24 @@ export function useRequestFee(
...
+  );
+
+  watch(
+    [rpcUrl, requestManagerAddress, requestAmount],
+    () => {
+      loading.value = true;
+      cb.value();
+    },
+    {
+      immediate: true,
```
manuelwedler, line 56: I was wondering if there is an easy way to not have the spinner on page load. Maybe we don't need this immediate option and it would work?

manuelwedler (commented):
Thanks for the work. Looks good! Just two nitpicks and a question.

frontend/src/composables/useRequestFee.ts
GabrielBuragev: Ah, thanks for the info. I am pretty fresh in typescript, don't mind if you see this beginner mistakes :smiley:

frontend/src/composables/useRequestFee.ts
GabrielBuragev: True, true and true. Will do the necessary changes for this!

frontend/src/composables/useRequestFee.ts
GabrielBuragev: I just checked this. 
Removing the `immediate: true` option would not change the behavior since the `rpcUrl` and `requestManagerAddress` are changing on page load. This happens only when user is connected to a network in his wallet that we support inside the sourceChain selector. In that case, the app prefills the sourceChain selection input with his currently connected address and therefore those values change and the `loading` value is set to true, even tho not needed when amount is still `undefined`.

However, i can solve this by adding a simple condition inside the watcher function which checks if the token amount is empty. If so, set the fee amount to `undefined` in that case.

manuelwedler: Go for it if you want to. It's really not a requirement, just looks a littler nicer to have a stable page on load I think.

frontend/src/composables/useRequestFee.ts
manuelwedler: Yea all good. I just want to mention also such small things, so you can get a better feeling for typescript more quickly.

GabrielBuragev:
Note: `immediate: true` was left untouched in this fixup commit since it's removal caused test failures inside the test suite of `useRequestFee`. 
A couple of tests depend on the immediate watcher function.

GabrielBuragev:
Also, i spotted that the spinner inside `App.vue` was not shown correctly. So i rolled back to having a wrapper around the spinner component there.

manuelwedler (approved):
Great! Thanks for the work!

PR #902, opened 2022-08-03
Frontend: fix several problems with the selector

### LINKED ISSUE #892: frontend: New selector component problems
When I access the `frontend-dev` version, I have the following issues:
  - my selection is not accepted, should mean after making a selection, the input field is still empty
  - the icons are broken and look like on the screenshot below
  
  
image

_Originally posted by @weilbith in https://github.com/beamer-bridge/beamer/issues/853#issuecomment-1196491361_

Additionally the current selection should be highlighted in the options.

manuelwedler:
fixes #892 

Addresses three minor bugs related to the new selector component and adds selection highlighting and return by escape key to it.


frontend/src/components/RequestTargetInputs.vue
```diff
@@ -72,6 +72,15 @@ const ignoreChains = computed(() =>
 );
 const { chainOptions } = useChainSelection(chains, ignoreChains);
 
+watch(
```
GabrielBuragev, line 75: There is only one example where i find this to be useful: 
user has no wallet connected -> selects a target chain -> selects the source chain to be the same as target chain.

All other cases are impossible in `production` environment due to the already implemented `ignoreChains` approach (check 5 lines above).

But this can also be solved by using the same `ignoreChains` approach (just like in this file) inside the `RequestSourceInputs.vue` component. That would be way more elegant and makes more sense. That way we can significantly reduce exposure to unpredicted errors due to unpredicted user behavior. :sweat_smile: 
If you implement that you can safely remove this piece of code too.

What do you think?

frontend/src/components/inputs/Selector.vue
```diff
@@ -99,4 +107,5 @@ const optionClasses = `flex flex-row items-center gap-8 cursor-pointer
   flex-[0_0_4.5rem] w-full px-8 rounded-xl border border-teal-light text-mint text-2xl
   hover:border-teal-dark hover:bg-teal-dark
 `;
+const highlightedOptionClasses = `!text-teal-light font-semibold`;
```
GabrielBuragev, line 110: I think the UI would look more intuitive if we make the background of the selected option to be `bg-teal-dark`. What do you think? 
If it differs from the provided design then ignore this.

GabrielBuragev (commented):
You covered many bugs with this one, nice!
I have 1 code improvement suggestion and one design improvement suggestion. 
Let me know if they make sense to you.

frontend/src/components/RequestTargetInputs.vue
GabrielBuragev: Sorry, highlighted the wrong line with this one.

manuelwedler: We talked about this and will go with my approach.

The "no connected wallet" case is the only one which this is needed. The problem about the `ignoreChains` approach is there is an edge case where it is not safe with the current implementation. An unconnected user may select a target chain and the connect to the exact same chain which would select it in the source. This would be valid and the user would be able to create a request with the same source and target.

frontend/src/components/inputs/Selector.vue
manuelwedler: yes looks nicer :+1:

GabrielBuragev (approved):
Awesome job done!

PR #903, opened 2022-08-03
frontend: Design adaptions

### LINKED ISSUE #859: Frontend: small design adaptions
There are a few places where we differ from the design. Most prominent is the transfer history. Olympe gave also some points to improve there. For example the filled dots should not have a border and the text should be ordered slightly differently. Also the text is moving a bit when steps are being completed due to the dots' animation, but it should stay in place.
Where we use teal-light as the color, the design uses actually sea green (#05B0AB).

- [x] align with Olympe 
- [x] implement changes

manuelwedler:
Closes #859 

There were a few places where our implementation of the transfer history differed from the design. Besides, this fixes the dots animation to not move the text around in the history.

Also the tailwind color theme has been reworked to make it align with the figma design.


frontend/src/components/layout/Tabs.vue
```diff
@@ -52,9 +52,9 @@ const headers = computed(() =>
     click: () => (activeTab.value = tab),
     classes: {
       'cursor-pointer': tab.label != activeTab.value.label,
-      'bg-teal-very-dark/40': tab.label != activeTab.value.label,
-      'text-teal-light': tab.label == activeTab.value.label,
-      'text-teal-light/40': tab.label != activeTab.value.label,
+      'bg-black/40': tab.label != activeTab.value.label,
+      'text-sea-green': tab.label == activeTab.value.label,
+      'text-sea-green/40': tab.label != activeTab.value.label,
     },
```
GabrielBuragev, line 58: Maybe we can simplify this by creating `inactiveTabClasses` & `activeTabClasses` and then inside the dom element assign the classes based on the condition `header.label == activeTab.label`.
The current approach makes all the header objects carry all this info with them which is unnecessary. 
Also these 4 equality checks are being evaluated each time per header element.

GabrielBuragev (changes requested):
One minor adjustment that i think would be beneficial. Otherwise everything looks good.
Nice UI improvements btw!

frontend/src/components/layout/Tabs.vue
manuelwedler: Not really related to this PR, but yes, makes it much cleaner. I'll go for it.

GabrielBuragev (approved):
Perfect!

frontend/src/components/layout/Tabs.vue
GabrielBuragev: Shouldn't have requested changes, a simple comment would've been enough, but thanks for understanding the need for this!

PR #908, opened 2022-08-04
Add mainnet deployment files

istankovic:


deployments/mainnet/deployment.json
```diff
@@ -0,0 +1,108 @@
...
+      "address": "0x913b5aC7E1124Ea3429593193bf44eB967FB7658",
+      "deployment_block": 15082755,
+      "deployment_args": [
+        "0x25ace71c97B33Cc4729CF772ae268934F7ab5fA1"
+      ]
+    }
+  },
+  "L2": {
+    "288": {
+      "MintableToken": {
```
fredo, line 20: I think there shouldn't be a mintable token for mainnet, even if it was deployed

deployments/mainnet/deployment.json
```diff
@@ -0,0 +1,108 @@
...
+        ]
+      },
+      "OptimismL2Messenger": {
+        "address": "0xa6FAEEE4DF263CaB9be09B747f4Be760940c884A",
+        "deployment_block": 712239,
+        "deployment_args": []
+      }
+    },
+    "10": {
+      "MintableToken": {
```
fredo, line 64: same here

deployments/mainnet/deployment.json
istankovic: > I think there shouldn't be a mintable token for mainnet, even if it was deployed

Does it harm? Actually, it was not deployed the second time, we're reusing the 1st deployment that you did.
The Rinkeby one has ti too and it makes testing easier because we don't have to track the addresses manually.

However, I can drop that commit if you still think it should not be there.

fredo: After discussion, please remove it from the mainnet deployment information as we will use actual tokens with value for testing

istankovic: Done.

deployments/mainnet/deployment.json
istankovic: Done.

PR #912, opened 2022-08-05
docker: fix relayer binaries location

istankovic:


istankovic:
Found out during Agent Setup Kit testing that the relayers are not where they are supposed to be.

palango (approved):

docker/Dockerfile.agent
```diff
@@ -3,8 +3,8 @@ FROM python:3.9
 COPY README.md pyproject.toml poetry.lock /src/
 COPY beamer /src/beamer
 
-RUN mkdir -p ./beamer/data/relayers
```
palango, line 6: I'm not sure why this changed, but OK

istankovic: The thing is, we use `WORKDIR` to change the working directory to `/src` and run  poetry from there, so the agent will expect to find relayers there, not under `/beamer/...`.

Issue #919, opened 2022-08-08
Fix Agent Setup Kit issues found by Jannik


istankovic:
We need to:

- [x] improve clarity around the account keyfile (keystore is customary)
- [x] add a Rinkeby testnet config file and instructions
- [x] make the agent exit with a better error message when the chain ID is not found in the deployment data
- [x] say how people can check that everything works fine (just the command to show the latest logs and make them verify it says “Sync done” or something)

fredo:
this can be closed @istankovic , right?

PR #920, opened 2022-08-09
beamer: agent: improve error reported when we don't find contract info

istankovic:
This is one of the things Jannik ran into, part of #919.

### DIFF
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -3,17 +3,26 @@ from concurrent.futures import ThreadPoolExecutor
 
 import structlog
 from eth_typing import Address
+from web3 import Web3
 
 import beamer.metrics
 from beamer.chain import POLL_PERIOD_MAINNET, POLL_PERIOD_TESTNET, EventMonitor, EventProcessor
 from beamer.config import Config
-from beamer.contracts import make_contracts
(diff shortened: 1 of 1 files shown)


beamer/agent.py
```diff
@@ -16,6 +17,14 @@
 log = structlog.get_logger(__name__)
 
 
+def _get_contracts_info(config: Config, web3: Web3) -> dict[str, ContractInfo]:
+    chain_id = ChainId(web3.eth.chain_id)
+    info = config.deployment_info.get(chain_id)
+    if info is None:
+        raise RuntimeError(f"Deployment info for chain ID {chain_id} not available")
```
fredo, line 24: can we add the path being used to it? Or at least give a hint where to look at/the problem lies.

istankovic: > can we add the path being used to it? Or at least give a hint where to look at/the problem lies.

Heh, it's not that pretty to add the path in here, I was hoping the "Deployment info" part would be hinting enough
(`--deployment-info`), but generally, it could also be the chain configuration (e.g. wrong RPC). Not sure how to improve this.

fredo: ok lets go for it now like this

fredo (approved):

PR #923, opened 2022-08-12
frontend: Delay request expiry listener

### LINKED ISSUE #741: frontend - reduce stress on RPC endpoints
This primarily targets the waiting for fulfillment or expiration.
Currently we listen on every new block since we created the request. This does the job and works fine for test networks. On production roll-ups this looks different. Roll-ups can produce quite a number of blocks per time. Doing multiple requests per each block can lead to becoming blocked by the RPC endpoint.

For the fulfillment we already have adaptive syncing for past events, and a strict filter for future ones. That looks good and there is possibly no further improvement than that.

For the expiration part, we can improve two different parts. First of all, we can skip the whole time window where the request is still valid. So we can "sleep" until the expected expiration date. In-between it is basically not possible to expire. But due to the unsynchronized clocks, the sequencer can be behind our time, so we still need to verify the block timestamp as before.
Once the validity period is over and there are still active claims, we do not need to check the claim number with every block. We can listen to `ClaimStakeWithdrawn` events for our specific request. Only in case of such an event, the number of active claims gets reduced again and might become zero.
Putting those together, we can safe a mass of RPC requests and should be really good prepared for production environments.

GabrielBuragev:
Fixes #741

In order to avoid RPC endpoint spam there was a need of an adaptive
transfer expiry event listening.
Previously, the transfer expiry check was attached and called each
time a new block was produced. That is, from the moment a transfer
was created and until it expires we would have this listeners active
for each transfer stored in the `localStorage` that is still pending.
With this change we will use adaptive listeners attached once
the transfer enters the time-frame where it is actually expected
to expire.
All the possible scenarios are covered with this impl.
1. If transfer didn't enter expected expiration time-frame when
checking against sequencer clock && our clock
- We sleep until the transfer enters the expected expiration
time-frame by our clock
2. If transfer entered expected expiration time-frame when checking
against our clock but not against the sequencer clock
- We wait for the sequencer clock to update with the newest
blocks by waiting on and listening on the next few block mints
3. If transfer entered expected expiration time-frame when checking
against our clock && sequencer clock:
3.1. If request was already withdrawn
- Cancel all listeners and return a rejection to signal a
transfer expiry.
3.2. If request still has active claims
- Listen to `ClaimStakeWithdrawn` events until `activeClaims`
falls down to 0. Once 0, return a rejection in order to signal
a transfer expiry.
4. If transfer entered expected expiration time-frame when checking
against our clock and sequencer clock && no one withdrawn the funds
yet && there are no active claims
- Return a rejection in order to signal a transfer expiry
With this approach, we avoid listening and querying the RPC endpoints
unnecessarily by using a more adaptive approach based on our
prior knowledge about the transfers expected behavior throughout
its life-span.

Tested all possible cases locally with a ganache instance.

### DIFF
--- a/frontend/src/services/transactions/request-manager.ts
+++ b/frontend/src/services/transactions/request-manager.ts
@@ -104,23 +104,40 @@ export async function getRequestData(
     throw new Error('No request known for this identifier!');
   }
 }
-
-export async function checkIfRequestHasExpired(
+type RequestExpiryInfo = {
+  validityExpired: boolean;
+  noActiveClaims: boolean;
+  notWithdrawnBySomeoneElse: boolean;
+  timeToExpiredMillis: number;
+};
+export async function getRequestExpiryInfo(
   rpcUrl: string,
   requestManagerAddress: string,
   requestIdentifier: UInt256,
   requestAccount: string,
-): Promise<boolean> {
-  const request = await getRequestData(rpcUrl, requestManagerAddress, requestIdentifier);
+): Promise<RequestExpiryInfo> {
+  const { validUntil, activeClaims, withdrawInfo } = await getRequestData(
+    rpcUrl,
+    requestManagerAddress,
+    requestIdentifier,
+  );
   const provider = new JsonRpcProvider(rpcUrl);
   const block = await provider.getBlock('latest');
-  const validityExpired = request.validUntil.lt(block.timestamp);
-  const noActiveClaims = request.activeClaims.eq(0);
+  const validityExpired = validUntil.lt(block.timestamp);
+  const noActiveClaims = activeClaims.eq(0);
   const notWithdrawnBySomeoneElse =
-    request.withdrawInfo.filler.toLowerCase() == '0x0000000000000000000000000000000000000000' ||
-    request.withdrawInfo.filler.toLowerCase() == requestAccount.toLowerCase();
+    withdrawInfo.filler.toLowerCase() == '0x0000000000000000000000000000000000000000' ||
+    withdrawInfo.filler.toLowerCase() == requestAccount.toLowerCase();
 
-  return validityExpired && noActiveClaims && notWithdrawnBySomeoneElse;
+  const timeNow = BigNumber.from(Date.now());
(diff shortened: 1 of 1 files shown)


frontend/src/services/transactions/request-manager.ts
```diff
@@ -104,23 +104,45 @@ export async function getRequestData(
...
-    request.withdrawInfo.filler.toLowerCase() == requestAccount.toLowerCase();
+    withdrawInfo.filler.toLowerCase() == '0x0000000000000000000000000000000000000000' ||
+    withdrawInfo.filler.toLowerCase() == requestAccount.toLowerCase();
 
-  return validityExpired && noActiveClaims && notWithdrawnBySomeoneElse;
+  const timeNow = BigNumber.from(
+    Math.ceil(Date.now() / 1000), // to seconds
+  );
+  const timeToExpiredMillis = Math.max(
+    validUntil.sub(timeNow).toNumber() * 1000, // to milliseconds
```
manuelwedler, line 136: Why not convert the `validUntil` to milliseconds before the call of `sub` instead of having two conversions here?

manuelwedler (commented):
Looks great! Just two nitpicky comments.

frontend/src/services/transactions/request-manager.ts
GabrielBuragev: Pretty new to the operations with BigNumbers, don't mind this rookie mistake. Thanks a lot for pointing out!

frontend/src/services/transactions/request-manager.ts
```diff
@@ -130,27 +152,66 @@ export function failWhenRequestExpires(
...
-    checkExpiration();
-    provider.on('block', checkExpiration);
+    configureListeners();
   });
 
-  const cancel = () => provider.removeAllListeners();
+  const cancel = () => {
+    provider.removeAllListeners();
+    contract.removeAllListeners();
+    if (timeout) clearTimeout(timeout);
```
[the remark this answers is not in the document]
GabrielBuragev: Nice catch! Fixed it.

manuelwedler (approved):
Thanks! Good job!

PR #924, opened 2022-08-12
docs: update outdated "contributing" docs

GabrielBuragev:
Aligned contract compile command to the latest directory structure.
Added missing argument inside contract deploy command which will enable
deploying the mintable token by default.


istankovic (approved):
Thanks!

PR #925, opened 2022-08-14
beamer: events: convert received bytes to HexBytes

istankovic:
This improves the log output of fill IDs.
~(`fill_id` is typed as `FillId`, which is a newtype of `Hexbytes`, but we get `bytes` instances from `web3`, which means we must convert them)~

We need to convert received `bytes` from `web3` to proper `FillId` which is derived from `HexBytes` to improve repr.

### DIFF
--- a/beamer/events.py
+++ b/beamer/events.py
@@ -166,6 +166,17 @@ def _make_topics_to_abi(contract: Contract) -> dict[bytes, ABIEvent]:
     return event_abis
 
 
+def _convert_bytes(kwargs: dict) -> None:
+    for name, type_ in (
+        ("fill_id", FillId),
+        ("fill_hash", FillHash),
+        ("request_hash", RequestHash),
+    ):
+        value = kwargs.get(name)
+        if value is not None:
+            kwargs[name] = type_(value)
+
+
 def _decode_event(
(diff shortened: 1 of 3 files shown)


beamer/events.py
```diff
@@ -177,6 +177,9 @@ def _decode_event(
         kwargs["chain_id"] = chain_id
         kwargs["block_number"] = log_entry["blockNumber"]
         kwargs["tx_hash"] = log_entry["transactionHash"]
+        fill_id = kwargs.get("fill_id")
+        if fill_id is not None:
+            kwargs["fill_id"] = FillId(HexBytes(fill_id))
```
fredo, line 182: FillId is not a NewType of HexBytes. It's a NewType of bytes
see https://github.com/beamer-bridge/beamer/blob/c5aa401c4984993084a561492ed36bd14ad3f8d9/beamer/typing.py#L13

istankovic: You're totally right, this had a couple of iterations and the PR description as well as the content is just wrong.
Pushed another version.

fredo:
Looks good! We should consider to use this for other bytes representations in our code base as well. (Like request hash, fill hash)

fredo (approved):

istankovic:
Pushed a new version that also does the same for `FillHash` and `RequestHash`.

fredo:
awesome!

PR #929, opened 2022-08-15
ci: add empty circleci config to make github happy

fredo:
Just for the time being until we have the circle ci config ready


fredo:
Closing. Will push a PR with the real config.

PR #930, opened 2022-08-15
First part of circle ci config

fredo:
Since the webhook to circle ci was introduced it blocks PRs. Unfortunately the only way to test circle ci is via github pushing to a branch and then let the CI run. 

To not block other PRs this is the first part of circle CIs config. It runs until lint. 
Until the config is complete we simply let circle and github CI run.


istankovic (approved):
:heart:

PR #931, opened 2022-08-16
frontend: Move terms and imprint to website and store checkbox state

### LINKED ISSUE #847: Add dApp disclaimer
There is a disclaimer in place that only allows user to use the dApp once agreed on.

Details [here](https://github.com/beamer-bridge/product/issues/67).

manuelwedler:
Closes #847 

It was requested to show the terms and the imprint from the beamer website. This removes the integrated terms and imprint, which were in modals. As this was the only place where the Modal component was used, it is also removed. Instead we link to the respective pages on the beamer website.

Besides, the checkbox state should be stored now according to https://github.com/beamer-bridge/product/issues/67. The state is stored in the settings store which is persisted to localStorage. This removes a step for second-time users and improves UX.


frontend/src/components/RequestDialog.vue
```diff
@@ -19,8 +19,25 @@
     <div v-if="transferFundsButtonVisible" class="flex flex-col items-center">
       <div v-if="DISCLAIMER_REQUIRED" class="relative">
         <div class="absolute -top-11 -left-44 w-96 m-auto flex flex-row gap-5">
-          <input v-model="disclaimerChecked" type="checkbox" :class="checkboxClasses" />
-          <span class="text-lg">I agree to the <TermsModal /></span>
+          <!--  
+            There seems to be a problem with tailwind and escaping the backslash in the TS variable. 
+            For this reason the content rule is added separately here in the HTML.
+          -->
+          <input
```
GabrielBuragev, line 26: I believe that this issue comes from the fact that tailwind scans our code to pick up all classes familiar to it for bundling them up in the build. Something like tree-shaking in JS. When the bacsklash was in the code it couldn't figure out or treated that class name as invalid and therefore skipped it.
Check the `dynamic values` section here:
https://v2.tailwindcss.com/docs/just-in-time-mode#arbitrary-value-support

frontend/tests/unit/stores/settings.spec.ts
```diff
@@ -11,17 +11,26 @@ describe('settings store', () => {
     setActivePinia(pinia);
   });
   describe('Settings initial value', () => {
-    it('initialize settings store with undefined values', () => {
+    it('initialize settings store with default values', () => {
       const settings = useSettings();
       expect(settings.$state.connectedWallet).toBeUndefined();
+      expect(settings.$state.disclaimerChecked).toBe(false);
```
GabrielBuragev, line 17: When writing tests i prefer to find solutions where i only do a single `expect` check (unless i am checking against different values for the same target). 
In this case, i would separate the `disclaimerChecked` and `connectedWallet` checks into their own test.
With the current approach it will make it harder to debug when something breaks in this test. Of course at this moment it will be pretty simple to debug since we have only 2 cases we are testing but if we add more store variables it can get messier here.

GabrielBuragev (approved):
One nitpick only, other then that everything looks good!

frontend/src/components/RequestDialog.vue
manuelwedler: Yeah makes sense. Thanks for mentioning. Btw I also think that it's not that nice to move the classes into a variable in the TS code. Thore preferred that because of readability reasons. Where we have a lot of classes, they were staying in one line with prettier. But having a variable kind of breaks with the reason of having tailwind in the first place, namely having the styles in the HTML. Maybe we should discuss this in the future.

frontend/tests/unit/stores/settings.spec.ts
manuelwedler: Cool makes sense to me :+1:

PR #934, opened 2022-08-17
frontend: update mainnet config

GabrielBuragev:
Updated the mainnet config file with the correct contract addresses
and tokens.
Added missing icons for the new rollups.


frontend/public/config.production.json
```diff
@@ -1,34 +1,34 @@
...
-      "fillManagerAddress": "0xeB5D11D41d68429D82e274f4FFEE94cc1aC950d2",
-      "explorerTransactionUrl": "https://blockexplorer.rinkeby.boba.network/tx/",
-      "rpcUrl": "https://rinkeby.boba.network/",
-      "name": "Boba Rinkeby",
+    "288": {
+      "identifier": 288,
+      "requestManagerAddress": "0x8D7ca47D7d0D9951CEdFC1E6eDA598268dAADE78",
+      "fillManagerAddress": "0x4e32CEc57BCaE80928f58D3dbD84550D397F5416",
+      "explorerTransactionUrl": "https://bobascan.com/tx/",
+      "rpcUrl": "https://lightning-replica.boba.network",
```
istankovic, line 8: I think we should use `replica.boba.network` (I remember we had issues with `lightning-replica.boba.network` before and 
`replica.boba.network` was then recommended by Boba people).

istankovic (approved):
Looks good to me, module the Boba RPC.

frontend/public/config.production.json
GabrielBuragev: I checked the scores of the RPC's here https://chainlist.org/ and therefore i decided to go with `lightning-replica.boba.network`.

GabrielBuragev: I just saw this on their discord channel:
```
Mainnet RPC
Read RPC: https://lightning-replica.boba.network/
Write RPC: https://mainnet.boba.network/
```
Should we maybe use `https://mainnet.boba.network/` since that is their recommended write RPC endpoint?

GabrielBuragev: Switching over to `replica.boba.network` as suggested by the whole team.

Issue #943, opened 2022-08-18
Add password option to generate_account.py


fredo:
Agents might want to use this to generate an account locally. Requested by agent

PR #944, opened 2022-08-18
chore: Add password option to generate_account.py

fredo:
fixes: #943

### DIFF
--- a/scripts/generate_account.py
+++ b/scripts/generate_account.py
@@ -4,6 +4,7 @@ from pathlib import Path
 
 import click
 import eth_account
+from eth_utils import to_checksum_address
 
 
 @click.command()
@@ -13,17 +14,20 @@ import eth_account
     metavar="KEY",
     help="Specify the private key to use instead of generating one.",
 )
+@click.option(
+    "--password", type=str, required=True, help="Choose a password to encrypt the keystore file"
+)
 @click.argument(
     "output",
     type=click.Path(file_okay=True, dir_okay=False, path_type=Path),
 )
-def main(key: str, output: Path) -> None:
+def main(key: str, password: str, output: Path) -> None:
     if key is None:
         key = random.randbytes(32)
-    account = eth_account.Account.from_key(key)
-    obj = eth_account.account.create_keyfile_json(account.key, b"")
+
+    obj = eth_account.Account.encrypt(key, password)
     output.write_text(json.dumps(obj))
-    print(account.address)
+    print(to_checksum_address(obj["address"]))
 
 
 if __name__ == "__main__":
--- a/docker/arbitrum/arbitrum.sh
+++ b/docker/arbitrum/arbitrum.sh
@@ -19,7 +19,7 @@ PRIVKEY=0x3ff6c8dfd3ab60a14f2a2d4650387f71fe736b519d990073e650092faaa621fa
 KEYFILE=$(mktemp -d)/${ADDRESS}.json
 echo Beamer deployer\'s keyfile: ${KEYFILE}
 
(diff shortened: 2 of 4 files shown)


scripts/generate_account.py
```diff
@@ -13,15 +13,16 @@
...
     "output",
     type=click.Path(file_okay=True, dir_okay=False, path_type=Path),
 )
-def main(key: str, output: Path) -> None:
+def main(key: str, password: str, output: Path) -> None:
     if key is None:
         key = random.randbytes(32)
     account = eth_account.Account.from_key(key)
-    obj = eth_account.account.create_keyfile_json(account.key, b"")
+    obj = eth_account.account.create_keyfile_json(account.key, password)
```
istankovic, line 25: Are you sure this is going to work?
`password` is a `str`, whereas previously we gave `b""`, which is a different type. Unfortunately, there is no type checking here, but `create_keyfile_json` ultimately ends up giving the password to `hashlib.pbkdf2_hmac`, which requires that the password is a `bytes`-like object, not `str`.

Additionally, what happens if the user does not give a password? Will `click` provide `None` or and empty string?

fredo: You are perfectly right. I just realized that the code used a method which is one abstraction layer deeper than I expected. 
`eth_account.Account.encrypt()` will do all the conversion we need and is much more convenient to use.

Regarding the password, I chose to have the password required and not giving an implicit password (i. e. `""`), to enforce the user actively typing a password.

scripts/generate_account.py
```diff
@@ -13,17 +14,20 @@
     metavar="KEY",
     help="Specify the private key to use instead of generating one.",
 )
+@click.option(
+    "--password", type=str, required=True, help="Choose a password to encrypt the keystore file"
```
istankovic, line 18: Without additional changes, making this `required` will break our E2E tests. I would actually prefer if this option is left... optional and instead make the default password empty.

fredo: IMHO, if it's only about our own e2e test, I'm more than happy to change it to make it work. I would really not give this script into the hands of the user, with making default passwords. 

So we should either don't promote the option to the user at all and then we can use it as an internal script which is not meant for external usage. (The contributing page should be updated, because it implicitly sets the empty password and the command for running an agent).

But as one of our real agents pointed rightfully out, they asked for a tool to generate the key locally reducing the risk of external exposure by not needing to send keys through the network.
If we promote this option, I am really not a fan of setting implicit passwords, because the user might not be aware and this could be a potential agent key with funds on it.

Additionally, I think it's not a big deal to add `--password ""` to our e2e commands.

istankovic: > 



> IMHO, if it's only about our own e2e test, I'm more than happy to change it to make it work. I would really not give this script into the hands of the user, with making default passwords.
>
> So we should either don't promote the option to the user at all and then we can use it as an internal script which is not meant for external usage. (The contributing page should be updated, because it implicitly sets the empty password and the command for running an agent).

Yeah, I never really imagined the script to be used in this manner, but OTOH I see no issues. I somehow hoped
whoever is going to run an agent is going to be knowledgeable enough and also have other (more trusted) ways to generate an address, or already have an address ready. But I also don't have anything against them using the script.

> But as one of our real agents pointed rightfully out, they asked for a tool to generate the key locally reducing the risk of external exposure by not needing to send keys through the network. If we promote this option, I am really not a fan of setting implicit passwords, because the user might not be aware and this could be a potential agent key with funds on it.
> 
> Additionally, I think it's not a big deal to add `--password ""` to our e2e commands.

Sure, please go for it. :+1:

istankovic (approved):
Looks good!

Issue #891, opened 2022-07-27
Relayer does not find execution of message


fredo:
May be a bug in the optimism SDK https://github.com/ethereum-optimism/optimism/issues/3107
This issue is for tracking purpose.

PR #946, opened 2022-08-18
relayer: use Promise.race instead of Promise.any

istankovic:
The latter will block until one of the promises is fulfilled, or all are
rejected. Since the returnOnPpidChange never rejects, this means the
relayer will be forever stuck in case the main promise rejects (i.e.
throws). Promise.race will instead return as soon as one promise is
rejected, which is what we want here.

Closes: #891.


GabrielBuragev:
:+1:

GabrielBuragev (approved):

PR #948, opened 2022-08-19
relayer: don't do anything if the message has already been relayed

istankovic:
Also print message status and correctly exit the JS process as soon
as one of the promises is resolved.


GabrielBuragev (approved):
LGTM

PR #950, opened 2022-08-23
Frontend: Add eslint rule to require curly braces for blocks

manuelwedler:
Adding this because I saw this quite often in the recent PRs.
I think it is in general good practice to have the curly braces around any block, because it provides better readability, maintainability and is less prone to errors. See https://stackoverflow.com/questions/4797286/are-braces-necessary-in-one-line-statements-in-javascript for some arguments.

### DIFF
--- a/frontend/src/services/transactions/request-manager.ts
+++ b/frontend/src/services/transactions/request-manager.ts
@@ -169,7 +169,9 @@ export function failWhenRequestExpires(
           requestManagerAddress,
           requestIdentifier,
         );
-        if (activeClaims.eq(0)) cleanUpAndReject();
+        if (activeClaims.eq(0)) {
+          cleanUpAndReject();
+        }
       });
     };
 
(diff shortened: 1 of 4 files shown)


frontend/src/services/transactions/request-manager.ts
```diff
@@ -184,13 +186,18 @@ export function failWhenRequestExpires(
...
-        else timeout = setTimeout(configureListeners, requestExpiryInfo.timeToExpiredMillis);
+        else {
+          timeout = setTimeout(configureListeners, requestExpiryInfo.timeToExpiredMillis);
+        }
       } else {
-        if (!requestExpiryInfo.notWithdrawnBySomeoneElse) return cleanUpAndReject();
+        if (!requestExpiryInfo.notWithdrawnBySomeoneElse) {
+          return cleanUpAndReject();
+        }
         if (!requestExpiryInfo.noActiveClaims) {
```
GabrielBuragev, line 201: Especially for this part of the code :smile:

GabrielBuragev (approved):
I agree, lets use it!

PR #959, opened 2022-08-24
frontend: Correctly abbreviate long token amounts

### LINKED ISSUE #698: frontend - correctly abbreviate long token amounts
Currently, in the header of the transfer history, we display a shorten version of the token amount that was sent. The value get shorten to two decimals after the dot. While this seems fine for DAI, this falls of for WETH. So we need a smarter solution here.
Please note that within the expanded body of the transfer, the exact amount is always visible.

### LINKED ISSUE #964: "USDC Available" always shows as 0.0
#### System information

OS & Version: Mac 10.14.6
Browser & Version: Chrome
Commit hash:0x9cc02d55307468ed13e65ca0aeaacf2cd2e31f743fe8bef8ec0092e459710bef

#### Expected behavior
When connecting your wallet, you should be able to see how much USDC you have in your wallet.

#### Actual behavior
It always shows 0

Screen Shot 2022-08-25 at 2 41 57 PM

It's like this for all transactions

manuelwedler:
Closes #698, closes #964

default beahavior for the new format() function:
- for values >= 1 shows maximum 5 digits totally or at least 2 decimals
- for values < 1 show 8 decimal places max
- for values < 1 if decimal places would only be 0 show a less than sign (e.g. < 0.001)

Or displays the given decimals from the options.

@GabrielBuragev this does not include the tooltip for displaying the full amount yet. I simply didn't have enough time. Could you add it either here or in a new PR? It should be displayed for the token balance and in the headers in the transfer history

### DIFF
--- a/frontend/src/components/RequestSourceInputs.vue
+++ b/frontend/src/components/RequestSourceInputs.vue
@@ -52,7 +52,7 @@
           </InputValidationMessage>
           <div v-else class="self-end">
             <div v-if="showTokenBalance" class="text-base mr-5 mt-1">
-              {{ formattedTokenBalance }} {{ selectedToken?.label }} available
+              {{ formattedTokenBalance }} available
             </div>
           </div>
         </div>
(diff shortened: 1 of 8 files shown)


GabrielBuragev (approved):
Nice job Manu. I love that we moved the formatting logic to the TokenAmount class, and even more that we removed the direct usage of `BigNumber` inside some of the composables thanks to this approach.

I improved the test messages & splitted some tests that were mistakenly bound together.

On top of that i added the tooltip functionality on the user balance as discussed.

GabrielBuragev:
Also, i didn't add the tooltip on the transfer expandable card inside the activity tab since that info is already there when you expand it.
image

PR #969, opened 2022-08-25
deployments: mainnet: update the commit ID to correspond to the contr…

istankovic:
…act source

[skip ci]

Forgot to do this yesterday.


fredo:
Did you check whether all contracts correspond to that commit id?

istankovic:
> Did you check whether all contracts correspond to that commit id?

The second-to-last contract change was made in June, on 29th, which was before our deployment. The last change was the validity period change, which we deployed but did not reflect in the deployment file as the new commit ID. This commit ID is a little bit newer (few commits) than that, but for the purposes of contract source, that should not matter. Therefore, I am convinced that, with this patch, checking out the contract source will correspond to the current mainnet deployment content.

fredo (approved):

PR #979, opened 2022-08-31
frontend: Add back button on selector view & hide action button when selector view is opened

### LINKED ISSUE #898: Frontend: Chain selection view cannot be closed once opened
ss_20220801_164929

At this moment, there is no way of closing the chain selection view without selecting a chain from the list of options. 

#### Why should this feature exist?
It would be nice to provide this feature to the users so they don't need to select a chain just because they want to go back to the request form view.
Also, if we don't provide them with a `close` function they will be forced to choose a chain and therefore possibly need to add that chain's RPC config inside their browser wallet. This action is triggered by our app when we detect that the user doesn't have that network yet configured inside his wallet provider.

**IMO**: Users should have the option to close the selection view if they cannot find their chain under the list of available chains on the beamer app, hussle free.

#### Steps to reproduce this behavior
Click on any chain selection input and try to close the view that follows without selecting a chain.

#### Specification
Decision made in accordance with Olympe:
- A button showing a left arrow should be shown on the bottom left of the big box. Clicking it closes the selector
- The "Transfer" button should be hidden while the selection is open to reduce distraction

GabrielBuragev:
Fixes #898 

This brings a new approach to how the app handles multiple Teleports on same target. Until now, the first element inside the Teleport target was displayed.
This makes it harder to control and override the content of the Teleport target since Vue Teleport component works only in [append mode](https://vuejs.org/guide/built-ins/teleport.html#multiple-teleports-on-the-same-target). 

There was a need of hiding the current displayed content inside the form action button when the selector view was open. 
In order to avoid introducing another store for managing such minor misc things, or another complex solution which might require communication between 3+ levels of components i settled down on using an approach where the most recently added element to the Teleport target is displayed.

With this i enabled inserting empty DOM elements on the lowest "leafs" in the Component tree in order to hide the other content inside the Teleport target.

### DIFF
--- a/frontend/src/components/inputs/Selector.vue
+++ b/frontend/src/components/inputs/Selector.vue
@@ -44,7 +44,13 @@
         </div>
       </div>
       <div class="items-end">
-        <div class="text-5xl cursor-pointer text-sea-green" @click="closeSelector">&lt;</div>
+        <div
+          class="text-5xl cursor-pointer text-sea-green"
+          data-test="close-trigger"
+          @click="closeSelector"
+        >
+          &lt;
+        </div>
       </div>
     </div>
   </Transition>
--- a/frontend/src/components/inputs/Selector.vue
+++ b/frontend/src/components/inputs/Selector.vue
@@ -66,6 +66,8 @@ interface Props {
 
 interface Emits {
   (e: 'update:modelValue', value: SelectorOption<unknown>): void;
+  (e: 'opened'): void;
+  (e: 'closed'): void;
 }
 
 const props = defineProps<Props>();
@@ -75,9 +77,13 @@ const opened = ref(false);
 const openSelector = () => {
   if (!props.disabled) {
     opened.value = true;
+    emits('opened');
   }
 };
-const closeSelector = () => (opened.value = false);
+const closeSelector = () => {
+  opened.value = false;
+  emits('closed');
+};
 
 const searchFilter = ref('');
 const filteredOptions = computed(() =>
--- a/frontend/src/components/inputs/Selector.vue
+++ b/frontend/src/components/inputs/Selector.vue
@@ -23,7 +23,7 @@
         name="searchFilter"
         placeholder="Search"
         :focus-on-mount="true"
-        class="flex-[0_0_4.5rem] !text-left"
+        class="flex-[0_0_4.5rem]"
         data-test="search-field"
       />
       <div
@@ -42,6 +42,9 @@
           <img v-if="option.imageUrl" class="h-12" :src="option.imageUrl" />
           <span>{{ option.label }}</span>
         </div>
+      </div>
+      <div class="items-end">
+        <div class="text-5xl cursor-pointer text-sea-green" @click="closeSelector">&lt;</div>
       </div>
     </div>
   </Transition>
--- a/frontend/src/components/RequestSourceInputs.vue
+++ b/frontend/src/components/RequestSourceInputs.vue
@@ -27,6 +27,8 @@
         :options="chainOptions"
         placeholder="Source Rollup"
         required
+        @opened="hideActionButton"
+        @closed="showActionButton"
       />
       <InputValidationMessage v-if="v$.selectedSourceChain.$error">
         {{ v$.selectedSourceChain.$errors[0].$message }}
@@ -76,6 +78,8 @@
             :options="tokens"
             placeholder="Token"
             required
+            @opened="hideActionButton"
+            @closed="showActionButton"
           />
           <Tooltip
             class="self-end -mr-3"
@@ -156,6 +160,7 @@ import { useTokenBalance } from '@/composables/useTokenBalance';
 import { useTokenSelection } from '@/composables/useTokenSelection';
 import { useConfiguration } from '@/stores/configuration';
 import { useEthereumProvider } from '@/stores/ethereum-provider';
+import { usePortals } from '@/stores/portals';
 import type { Chain } from '@/types/data';
 import type { RequestSource, SelectorOption } from '@/types/form';
 import { TokenAmount } from '@/types/token-amount';
@@ -174,6 +179,7 @@ const emits = defineEmits<Emits>();
 
 const configuration = useConfiguration();
 const ethereumProvider = useEthereumProvider();
+const { hideActionButton, showActionButton } = usePortals();
 
 const { provider, signer } = storeToRefs(ethereumProvider);
 const { chains } = storeToRefs(configuration);
--- a/frontend/src/components/RequestTargetInputs.vue
+++ b/frontend/src/components/RequestTargetInputs.vue
@@ -9,6 +9,8 @@
         :options="chainOptions"
         placeholder="Target Rollup"
         required
+        @opened="hideActionButton"
+        @closed="showActionButton"
       />
     </div>
     <div class="flex flex-row gap-5">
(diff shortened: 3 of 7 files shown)


compojoom:
Do I have to do anything special to see the arrow?
image

That is on firefox.

compojoom:
also I have a small rant @olympe-silve 

// rant start

Why is this back arrow placed at the bottom left? Most of the content in the chain selection is on the top: search, networks and now a mile down on the left we have an arrow? What is this arrow now? Is it going to teleport me to the land of the blessed or to the carpal tunnel land? 
I would say it is cool on mobile since my finger won't have to travel to the top left, but for everything else this is a strange position and you are just playing mind-games with me. Why is there a need for a back arrow? We are not navigating to a new page? We've just overlay-ed some more content? Why can't we figure out a way to place a close button `X` next to the input where I started the navigation? 
We start the navigation by clicking in the input field, why not end it somewhere near by? I won't even see the back arrow when I start the chain selection and I'll hate you in my sleep for this! :)

// rant off

compojoom:
@GabrielBuragev this PR is too big in my opinion. 
The first sentence is `fixes https://github.com/beamer-bridge/beamer/issues/898` - so I went to this issue - read what it is all about and then started looking into the changed code. The whole time I'm wondering why you are dealing with the signers, when the only thing that should be there is the back arrow. 

Where is the issue for "hide action button when selector view is opened"?

I don't agree with your approach and I think that @manuelwedler should be the final instance on whether the PR should be accepted as is. 

I understand why you don't want to have a store for such a thing, but I think that a store is the correct thing to have. And despite it being some more code now, it's less code in the future. 

The whole point of redux/stores is to manage app state in a predictable manner. If the submit button was part of the form it would be easier for you to manage the state of that button right inside of the component - now that it is outside - nothing change in my opinion.  You should still manage the button state through component code and not through CSS code. 

Today it is the networks selection, tomorrow you could decide to hide it because a tooltip on the top right is displayed to the user. Having a store that deals with such UI elements would allow you to be flexible.

That's what I think. Other than this - the PR looks fine & if Manuel is OK with this approach - it's ok with me.

olympe-silve:
> also I have a small rant @olympe-silve
> 
> // rant start
> 
> Why is this back arrow placed at the bottom left? Most of the content in the chain selection is on the top: search, networks and now a mile down on the left we have an arrow? What is this arrow now? Is it going to teleport me to the land of the blessed or to the carpal tunnel land? I would say it is cool on mobile since my finger won't have to travel to the top left, but for everything else this is a strange position and you are just playing mind-games with me. Why is there a need for a back arrow? We are not navigating to a new page? We've just overlay-ed some more content? Why can't we figure out a way to place a close button `X` next to the input where I started the navigation? We start the navigation by clicking in the input field, why not end it somewhere near by? I won't even see the back arrow when I start the chain selection and I'll hate you in my sleep for this! :)
> 
> // rant off

So, this arrow is meant to bring us back to the main form, not to move forward to the next step. 
Once the user clicks on the network of their choice, they are automatically brought them back to the main form. 
To actually move forward with the transaction, the user clicks on the button bellow the main form container.

When they are reviewing the available options, the return arrow is just there so that they have the _very optional_ option to see that list _without_ making a decision just yet. We do want the user to choose a network though, so the return button not part of the "filling the form" UX flow. It's a side flow. 
The risk of not putting an arrow would be for the user who just wanted to check the networks list to be stuck there. And have no choice but to use their history navigation to go back to a form they might have already partially filled (= frustrating), because until they click on a network, they don't know it brings them back automatically. 

As for the arrow position: considering our audience, who in huge majority reads from left to right, it should be on the left because it's bringing the user back to the form they were on previously (ok you know that).
And It is located at the bottom and not up with the rest of the actionable items because :
- we don't want the user to click on "return" accidentally (=frustrating) when they are choosing the network
- we want them to consider the network list first. (Eyes read through a page following a Z pattern, we don't want this pattern interrupted right away with an option to return) 
- the upper part of the container is already busy enough.

So yeah it's not at all for for carpal tunnel comfort. 
It's there out of consideration for clumsy clickers, indecisive lurkers and people with fat fingers. And to entirely to ruin your sleep.
But of course if we get feedback from users as confused as you are, it will be changed.

frontend/src/components/inputs/Selector.vue
```diff
@@ -43,6 +43,15 @@
           <span>{{ option.label }}</span>
         </div>
       </div>
+      <div class="items-end">
+        <div
+          class="arrow-back text-[2.2rem] after:content-['\1F870'] cursor-pointer text-sea-green"
```
manuelwedler, line 48: Same issue as Daniel has. The character is not rendering as an arrow. For me both on Chrome and Firefox.

frontend/src/components/TransferHistory.vue
```diff
@@ -38,7 +38,7 @@
     <div class="absolute bottom-0 right-0 w-full h-10 z-10 bg-gradient-to-t from-teal" />
   </div>
 
-  <SafeTeleport to="#action-button-portal">
+  <SafeTeleport v-if="signer" to="#action-button-portal">
```
manuelwedler, line 41: This was also not clear to me why necessary. I removed it locally to try it. Switching to the history while the selector is open would not show the new transfer button without this line. This also looks pretty unpredictable. Not sure why it works. Maybe because the if is reevaluated when you switch and the button is appended last in this case?

This seems to me like an example that the approach could be prone to errors in the future. What happens for example if we add another tab or any other kind of view?

manuelwedler (changes requested):
I really see your efforts in trying to keep it simple in this case. Thanks for that. But I also think that this is a point where we could improve a lot on the architecture. Although a store would also not have been my first approach to this, Daniel made some good points. This looks like it can get easily out of our hands and show some unpredictable behavior. So I'm in for the store approach. Not sure how a store would look like though. Just storing if the button is showing or also what kind of button? @compojoom maybe has some more input?

Last, hiding the button is in the issue #898 in my last comment. I think we should not have issues split in such a way. It was one design decision to add the back button and reduce distraction by hiding the big action button. I see that it is not clear by the issue because the final decision is in a comment, so rather we should update our issue descriptions to make them clear.

frontend/src/components/inputs/Selector.vue
GabrielBuragev: Waiting for @olympe-silve to provide me with an svg so i can proceed with solving this problem.

olympe-silve: @GabrielBuragev thank you for the reminder. here it is https://drive.google.com/file/d/1P9002z_a2jcFToCRSH6IG2KAFKt5JCNj/view?usp=sharing

GabrielBuragev:
So, i tried several different approaches by using store to solve this issue but most of the times came to a brickwall due to the way how the whole view layout is implemented.
Instead of the `Button` components being rendered in their own `Tab` content container (one inside Transfer tab and one inside Activity tab) they are scattered in multiple components and sent/rendered in the portal which is located inside the  `Home` view component.

It's hard to achieve a better solution at the moment without refactoring the whole layout in order to be able to move the `Button` components down to their respective `Tab` content containers. The fixed heights of the `Tab` content containers are making it impossible to simply move the `Button` components at the end of the `Tab` content.

I believe these buttons should be located & rendered in the `Tab` content containers instead of the `Home` view component (via a portal) as they both hold a different action and should be expected to behave differently under certain conditions (which we might add in the future).

I pushed a temporary solution to this which uses a store approach for hiding/showing the portal but i am still not satisfied with the layout implementation and will discuss with @manuelwedler on how we can improve this (refactor) in a future issue.

What do you think @manuelwedler? Will the latest changes be sufficient for closing this?

manuelwedler (approved):
For me this looks good for now and much better than the CSS solution. But sure let's discuss later how we can solve the layout issues. Thanks for the effort :)

GabrielBuragev:
Not optimal i agree, but lets focus on refactoring together with writing tests in the next sprints.
Thank you all for the review and valuable input!

PR #986, opened 2022-08-31
frontend: Fix error message overflow & progress line gaps

### LINKED ISSUE #904: Frontend: Error Message cut by borders
#### System information

OS & Version: (WIP)
Browser & Version: (WIP)

#### Expected behavior
Not have the message cut out by the borders

#### Actual behavior
There seems to be an issue with the UI; the error message was cut. Please see the image below

Below is also a comment from the member.
"Like this obvious address format error, it should be prompted in the previous step, not an error during the conversion process, and this is a suggestion hope it helps"

I have also opened a thread in Discord if you guys need to contact the guy.

"When I was operating, I deliberately entered the wrong address format, and it did not remind me the first time. It would be better if the user is told that the address is in the wrong format before processing the transaction."

GabrielBuragev:
Fixes #904, #995

### DIFF
--- a/frontend/src/components/layout/Progress.vue
+++ b/frontend/src/components/layout/Progress.vue
@@ -30,39 +30,48 @@ interface Props {
 defineProps<Props>();
 </script>
 
-<style scoped>
-/* 
+<style lang="scss" scoped>
+/*
 Please note that the used numeric values here are based on `rem` to scale with
 the application, but are very fixed an not variable. If any of these value gets
 touched, it is likely to break somewhere else. Using variables and calculation
 only barely helps with this problem. As this is a very specialized custom
 component, this inflexibility can be tolerated for now.
 */
+$ballHeight: 1.75rem;
+$ballWidth: 1.75rem;
+$ballBorderWidth: 2px;
+$liMarginY: 1.25rem;
 
 li {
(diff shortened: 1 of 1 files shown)


compojoom:
I coudn't reproduce the error, but I modified the source and  I can confirm that the error message itself is no longer cut:
image

Don't you however think that it is worth fixing the the height of the line on the left while on this issue?

compojoom (approved):

GabrielBuragev:
If i fix the height of the line does that mean i have to create another issue and link it to this one? If i follow your suggestions from the other PR (#979) then i should do it like that since fixing that is outside of the current scope of this PR.

GabrielBuragev:
Using only % doesn't solve all the cases. When set to a %, the bigger the `<li>` element height gets the bigger the gap gets once again.
My solution is using `calc` with dynamic variables in order to provide some sort of flexibility.

compojoom (approved):
Looks good!

compojoom:
> If i fix the height of the line does that mean i have to create another issue and link it to this PR?

No need to over do it. :) I mean in this case fixing the line is kind of related to the error message. In the other issue for me the second fix was just unrelated to the original issue.

PR #996, opened 2022-09-01
frontend: Validate amount when token not selected

### LINKED ISSUE #970: Amount not validated if token is not selected
Seems like the amount is not validated if a token is not selected:

image

I would expect that at least there is a "positive number" check performed even without a token.

GabrielBuragev:
Fixes #970


GabrielBuragev:
Thanks for the suggestion @compojoom.
I had that idea in mind as well while implementing the input validation. Even started implementing it that way, but halway there i figured there is no standard implementation cross browser/device and its really hard to make 100% coverage with it.

If you are interested you can read my comments on PR #918 where i explain why i chose to back off from implementing it and switch to aggressively validating only.

Thanks for the review!

PR #999, opened 2022-09-02
frontend: Fix several issues related to transfer history container

### LINKED ISSUE #965: Glitch after First Transaction. Second transaction does not show up on Activities page until you confirm on metmask
#### System information

System information
OS & Version: Mac 10.14.6
Browser & Version: Chrome
Commit hash:0x04d2d3faf51ae4dda52888a2cd7d9b35af7993272751384cf2eba585044ab498 

#### Expected behavior
When you start your second transaction, you should see the transaction happening on the activity page.

#### Actual behavior
You only can see it happen after you confirm on metamask (Sorry for the slow video, Got lost a little lol)

Do another transaction after the first one

But, when you switch the network and do a transaction, it goes fine.

### LINKED ISSUE #988: Transaction details disappear once you scroll down and up again
#### System information

OS & Version: Windows
Browser & Version: Chrome

#### Expected behavior
If I open a transaction detail view and scroll down and up again the details should still show OR it should not show but without a big gap between 2 transactions

#### Actual behavior
image

scroll down and up again

image

1. Have enough transactions that you can scroll
2. open detailed view of the transaction
3. scroll down, so details don't show
4. scroll up again and see bug

### LINKED ISSUE #998: Frontend: Box shadow is overlaying the label text making it harder to read
Due to an existing box-shadow implementation on top of the transfer history container, the labels and elements that are close to the top of the container are overlayed with shadow. 
The elements need to be pushed further down so the shadow is not falling on the labels by default when the container has not been scrolled.

Also the shadows on the bottom of the container are not visible due to incorrect paddings/margins of the container.

image

GabrielBuragev:
Fixes #965  
Fixes #998 
Fixes #988 

Check commit message for more info on  #965 solution.

### DIFF
--- a/frontend/src/components/TransferHistory.vue
+++ b/frontend/src/components/TransferHistory.vue
@@ -1,41 +1,43 @@
 <template>
-  <div class="relative w-full h-full pt-3" v-bind="$attrs">
-    <!-- Top gradient to create phase out effect for scrolling content. -->
-    <div
-      class="absolute top-0 right-0 w-full h-10 z-10 bg-gradient-to-t from-transparent to-teal"
-    />
-
-    <div
-      ref="listElement"
-      class="w-full h-full overflow-y-auto overflow-hidden pb-20 no-scrollbar"
-    >
+  <div class="relative w-full h-full pt-6" v-bind="$attrs">
+    <div class="gradient-wrapper relative w-full h-full">
+      <!-- Top gradient to create phase out effect for scrolling content. -->
       <div
-        v-if="transfers.length === 0"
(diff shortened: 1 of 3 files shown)


manuelwedler (approved):
Wow, I'm quite impressed that you managed to solve all the issues with the transfer history. It was buggy since the first implementation. Thank you!

GabrielBuragev:
Thanks for the review @manuelwedler! 
The fact that `nextTick` was not working was bothering me so much and i am not a big fan of using timeouts but unfortunately that was the only solution to this issue as described in the commit message.

Issue #505, opened 2022-04-01
Show commit/version in frontend


palango:
For debugging and bug reports it would help to display the version and/or commit hash in the frontend.

It's available on DO as a evn var: https://docs.digitalocean.com/products/app-platform/how-to/use-environment-variables/#app-wide-variables

# Acceptance criteria
- [ ] Commit version is visible on dapp and copyable

istankovic:
This is a good idea! :+1:

Issue #473, opened 2022-03-30
Introduce Versioning of compatibility between agents and contracts


fredo:
Problem description: 

We are currently working on the contract interfaces. Short term this will be done a couple of times more. At the same time we are fixing functionality or improving performance on the agent.
Just updating the production agent to the latest main will break compatibility with the deployed contracts. At the same time we want to have the fixes being updated. 

Outcome of the discussion:

General:
- Make sure that agents and contracts interface is always compatible for each version
- Create a forbid requests switch
- Upon each new deployment let the old agent run until all its requests/claims are resolved and deploy a new agent in parallel.
- Define a versioning scheme to align with the above

For Testnet: 

- have rapid development cycles
- Deploy any sprint (if necessary ) new contracts and agents

For Mainnet deployment:
- Define a more "long term" interface between contracts and agents

@palango @istankovic please add to this comment in case I'm missing something

fredo:
Outcome of today's discussion:

As a simple way to mitigate versioning problems between agents and contracts we can do the following:

- introduce version numbers for contracts
- Provide abstract interfaces on the agent's side for each contract version
- In the agent's core use abstract functions (withdraw, claim, etc)
- Having another module which handles the interaction with the contract specified by the version

This is applicable to events as well as contract calls

It would be nice to have a way to automatically check whether the interface between the agent and the contract has changed and that we need to upgrade the version. 

If I have forgotten anything important, please add @weilbith

fredo:
I think it is very cheap to introduce now compared to introducing in the future.

istankovic:
One potential issue that I see here is that if the request/claim state machine changes, it's going to be non-trivial to adjust for that, unless the state machine is cloned and modified, which not nice. In other words, how do we accommodate changes to the state machine(s) between versions?

Additionally, this issue seems more like an epic and it would be good to have much smaller, concrete and actionable issues written down as children of this one.

weilbith:
[...]

PR #1003, opened 2022-09-06
Release v0.1.5

istankovic:


GabrielBuragev (approved):
LGTM :heavy_check_mark:

GabrielBuragev:
One question to @manuelwedler though.
Does it make sense to update the `version` property of `package.json` with the corresponding version?

I would say it doesn't make much sense since we dont have the `frontend` published as npm package. What are your thoughts?

istankovic:
> Does it make sense to update the `version` property of `package.json` with the corresponding version?

Please note there are https://github.com/beamer-bridge/beamer/issues/505 and https://github.com/beamer-bridge/beamer/issues/473. We have been delaying work on those, but at some point it would be really nice to have them.

manuelwedler:
> One question to @manuelwedler though. Does it make sense to update the `version` property of `package.json` with the corresponding version?
> 
> I would say it doesn't make much sense since we dont have the `frontend` published as npm package. What are your thoughts?

Since we are using the same changelog for the frontend now, I think we should follow the same versioning for it. Even if we do not publish it to npm. So yes I would start to use the `version` field now.

Issue #893, opened 2022-08-01
RequestManager whitelist


fredo:
Currently, only the FillManager respects a whitelist of agents who are eligible to fill requests. If only a set of addresses is allowed to fill then only the same set should be allowed to claim.

In order to not need to maintain multiple lists, we can introduce the whitelisting process on L1 and send the allowances to the respective chains. 

It would also reduce an attack vector where the network could be spammed by false claims.

taleldayekh:
@fredo create new issue and close this one.

istankovic:
> In order to not need to maintain multiple lists, we can introduce the whitelisting process on L1 and send the allowances to the respective chains.

How would "send the allowances" actually look?

fredo:
similar to calling the resolution registry via L1, a L1 whitelist registration would send a message to the respective roll ups and call the "addWhitelist" function on the respective contracts.
same as for "resolve" , we would need to make sure the this function is restricted to be called by a certain L1 contract.

istankovic:
> In order to not need to maintain multiple lists, we can introduce the whitelisting process on L1 and send the allowances to the respective chains.

But we would still end up with "multiple lists", right? Both the fill manager and the request manager would have a list of allowed fillers/claimers, right?

istankovic:
After a discussion with @fredo we decided to

- [ ] add a whitelist to `RequestManager`
- [ ] modify the agent to check whether it is whitelisted in the source `RequestManager`

fredo:
can we also add 

- [ ] adapt `call_contracts.py whitelist` to whitelist directly on all corresponding contracts

PR #1005, opened 2022-09-07
Request manager whitelist

fredo:
resolves: #893 

For the tests I created a function called `alloc_whitelist_accounts` which takes contracts as an argument and calls `addAllowedLp()` on it. Since it is a util function in must be passed in the arguments.
The positive side on this is that I had to go through all the tests and bring more context to the addresses and give them better names. I also removed some overlappings where the requester was also the claimer (not that it is harmful in particular but it removes the possibgility of failures due to address misusage).

Regarding `call_contracts.py whitelist` I did not come up with a proper solution because the script only consumes one RPC endpoint but it would require two here. Since it is not critical for the audit I suggest to create a follow up issue for that.

### DIFF
--- a/beamer/tests/contracts/test_fill_manager.py
+++ b/beamer/tests/contracts/test_fill_manager.py
@@ -11,10 +11,10 @@ def test_fill_request(fill_manager, token, deployer):
     filler, receiver = alloc_accounts(2)
 
     with brownie.reverts("Ownable: caller is not the owner"):
-        fill_manager.addAllowedLP(deployer, {"from": filler})
+        fill_manager.addAllowedLp(deployer, {"from": filler})
 
-    whitelist_tx = fill_manager.addAllowedLP(deployer, {"from": deployer})
-    assert "LPAdded" in whitelist_tx.events
+    whitelist_tx = fill_manager.addAllowedLp(deployer, {"from": deployer})
+    assert "LpAdded" in whitelist_tx.events
 
     token.approve(fill_manager.address, amount, {"from": deployer})
     fill_manager.fillRequest(
@@ -36,7 +36,7 @@ def test_fill_request(fill_manager, token, deployer):
             {"from": deployer},
         )
 
-    fill_manager.removeAllowedLP(deployer, {"from": deployer})
+    fill_manager.removeAllowedLp(deployer, {"from": deployer})
     with brownie.reverts("Sender not whitelisted"):
         fill_manager.fillRequest(
             1,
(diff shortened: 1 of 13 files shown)


beamer/tests/util.py
```diff
@@ -28,6 +29,14 @@ def alloc_accounts(n):
     return tuple(_alloc_account() for _ in range(n))
 
 
+def alloc_whitelisted_accounts(n, contracts):
```
cducrest, line 32: This assumes each contracts is owned by the default address. Maybe that's an assumption you have everywhere in the tests, in which case it is fine.

beamer/tests/contracts/test_fill_manager.py
```diff
@@ -2,28 +2,32 @@
...
     with brownie.reverts("Ownable: caller is not the owner"):
-        fill_manager.addAllowedLP(deployer, {"from": filler})
+        fill_manager.addAllowedLp(filler, {"from": receiver})
 
-    whitelist_tx = fill_manager.addAllowedLP(deployer, {"from": deployer})
-    assert "LPAdded" in whitelist_tx.events
+    whitelist_tx = fill_manager.addAllowedLp(receiver)
+    assert "LpAdded" in whitelist_tx.events
+    blacklist_tx = fill_manager.removeAllowedLp(receiver)
+    assert "LpRemoved" in blacklist_tx.events
```
cducrest, line 20: It's boring to read tests that do a little bit of everything, but this PR does not change it.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -40,15 +41,22 @@ def test_request_invalid_target_chain(request_manager, token):
...
+    (claimer,) = alloc_whitelisted_accounts(1, {request_manager})
     request_id = make_request(
         request_manager, token=token, requester=requester, target_address=requester, amount=1
     )
 
+    with brownie.reverts("Ownable: caller is not the owner"):
+        request_manager.addAllowedLp(claimer, {"from": requester})
+
+    whitelist_tx = request_manager.addAllowedLp(claimer, {"from": deployer})
+    assert "LpAdded" in whitelist_tx.events
```
cducrest, line 56: Same as previous comment, this has little to do with `test_claim`

cducrest (approved):

beamer/tests/contracts/test_fill_manager.py
fredo: Yes, I'm also not happy about that we have some tests which do a little bit of everything. That's also the case on lots of other places, but I totally agree. 

The lines you mentioned add and remove the `receiver` though not the filler, in order to test that events are created. Still misplaced.

fredo: We should definitely refactor this

Issue #990, opened 2022-09-01
Use request hash as key in RequestManager


fredo:
part of #989 

We need to do the following

### Change key of mapping to request hash

* use the requestId as a nonce
* create request hash upon request creation
* change mapping to bytes32 => request
* adapt `validRequestId` modifier

### Adapt agent
* Agent must interpret key of mapping correctly
* Change internal dictionary as well.
* Remove `_find_request_by_request_hash` usage 

### Adapt frontend
* frontend must interpret key of mapping as hash bytes
@GabrielBuragev are there more changes necessary? 

### Additional
* Adapt tests
* Update documentation

## Summary of tasks
- [ ] Change mapping in Request Manager
- [ ] Adapt Agent
- [ ] Adapt frontend
- [ ] adapt agent and contract tests
- [ ] update relevant documentation

PR #1016, opened 2022-09-11
Unify request id and request hash

fredo:
resolves: #990 

The request id is now equal to the request hash. All details can be found in #990.


contracts/contracts/RequestManager.sol
```diff
@@ -192,16 +189,13 @@ contract RequestManager is Ownable, LpWhitelist {
 
     /// Check whether a given request ID is valid.
     modifier validRequestId(uint256 requestId) {
-        require(
-            requestId <= requestCounter && requestId > 0,
-            "requestId not valid"
-        );
+        require(requestId <= counter && requestId > 0, "requestId not valid");
```
istankovic: This is not a sufficient check for the validity of a request ID. Since we use the same counter for both claims and requests, we can have request with ID 2 and claim with ID 3. If we now try to do something with request ID 3, the check will return `true` and we will proceed, which should not happen.

I remember we talked about using one of the guaranteed nonzero fields in the `Request` struct (e.g. `validUntil` or `sender`) to check whether a given request ID is valid.

contracts/contracts/RequestManager.sol
```diff
@@ -192,16 +189,13 @@ contract RequestManager is Ownable, LpWhitelist {
...
         _;
     }
 
     /// Check whether a given claim ID is valid.
     modifier validClaimId(uint256 claimId) {
-        require(claimId <= claimCounter && claimId > 0, "claimId not valid");
+        require(claimId <= counter && claimId > 0, "claimId not valid");
         _;
```
istankovic, line 203: Same concerns as https://github.com/beamer-bridge/beamer/pull/1016/commits/4a2320fd55765a3ffe37866401a10ce10357c0d0#r968370758.

contracts/contracts/FillManager.sol
```diff
@@ -22,7 +22,7 @@ contract FillManager is Ownable, LpWhitelist {
     ///
     /// .. seealso:: :sol:func:`fillRequest`
     event RequestFilled(
-        uint256 indexed requestId,
+        bytes32 indexed requestId,
```
istankovic, line 25: I suggest to consider using an `uint256` representation for request IDs, i.e. `uint256 request_id = uint256(hash(...))`. No data would be lost and it would be easier to handle than a 32-byte sequence.

contracts/contracts/BeamerUtils.sol
```diff
@@ -7,53 +7,53 @@ library BeamerUtils {
...
-    function createRequestHash(
-        uint256 requestId,
+    function createRequestId(
         uint256 sourceChainId,
         uint256 targetChainId,
         address targetTokenAddress,
         address targetReceiverAddress,
-        uint256 amount
+        uint256 amount,
+        uint256 nonce
```
istankovic, line 16: What was the reason we did not include `sourceTokenAddress` here originally?

contracts/contracts/FillManager.sol
fredo: I would rather do that on the consumer side of the event.

contracts/contracts/RequestManager.sol
fredo: uh. Something went wrong here. I remember changing this to checking the sender only.

contracts/contracts/BeamerUtils.sol
fredo: We don't need to do this. The request id is supposed to contain all information one need to verify that the request has been fulfilled. 
The source token address is nothing the user requested, it's actually something the agent requests to fulfill the service.

contracts/contracts/RequestManager.sol
```diff
@@ -334,7 +348,7 @@ contract RequestManager is Ownable, LpWhitelist {
     /// @param requestId ID of the request.
     /// @param fillId The fill ID.
     /// @return The claim ID.
-    function claimRequest(uint256 requestId, bytes32 fillId)
+    function claimRequest(bytes32 requestId, bytes32 fillId)
```
istankovic, line 351: This type of thing would not be necessary if we used `uint256` instead of `bytes32` for the request ID, as mentioned above. I feel like treating the ID as an uint256 would just be nicer.

contracts/contracts/RequestManager.sol
fredo: Okay this is werid. I see the changes in the changed files. Did you maybe review on a different commit?

contracts/contracts/RequestManager.sol
fredo: see my comment above. I doesnt understand what "this type of thing" means

contracts/contracts/RequestManager.sol
```diff
@@ -188,8 +189,11 @@ contract RequestManager is Ownable, LpWhitelist {
...
 
     /// Check whether a given request ID is valid.
-    modifier validRequestId(uint256 requestId) {
-        require(requestId <= counter && requestId > 0, "requestId not valid");
+    modifier validRequestId(bytes32 requestId) {
+        require(
+            requests[requestId].sender != address(0),
+            "requestId not valid"
+        );
         _;
```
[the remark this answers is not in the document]
fredo: agree.

contracts/contracts/RequestManager.sol
istankovic: > see my comment above. I doesnt understand what "this type of thing" means. If you mean changing the type, that's not a problem to my understanding. Any different representation, if wished, can be done outside of the protocol. We also don't convert other types to int just for nicer representation. We should do this on the representation level.

The thing we should ask ourselves is: how we use this ID? In what ways do we imagine it would be used?

The only reason this is `bytes32` is because the hash function returns that type. Semantically, this is an ID, as indicated by the name and previous typing we had. So, the way I am looking at it, our only usage is going to be for equality comparisons and lookups. There is no additional value provided by the data being interpreted as bytes here.

It is also not unusual at all to convert hash values to unsigned integers (e.g. just see Arbitrum contracts).

contracts/contracts/RequestManager.sol
istankovic: > Okay this is werid. I see the changes in the changed files. Did you maybe review on a different commit?

I encountered the change on the first commit, I think.

contracts/contracts/BeamerUtils.sol
istankovic: Thinking further, I think in the end this may actually be fine because we have both the source chain ID _and_ the nonce there, which should suffice on its own to identify (and only identify) a request.

fredo: due to the nonce it is strictly impossible to get the same request id twice. The main purpose of the request id is to forbid the agent to not fulfill the service correctly (i.e. sending fewer tokens, sending the wrong token) but generating a proof for the request id. 
The source token here is not of interest, the agent wouldn't simply fill it, if he isnt happy with it

beamer/tests/agent/test_request.py
```diff
@@ -87,13 +87,14 @@ def test_challenge_own_claim(config, request_manager, token):
     agent_address = to_checksum_address(agent.address)
     claim_stake = request_manager.claimStake()
     request = Request(
-        RequestId(100),
+        RequestId(31 * b"" + b"1"),
```
istankovic, line 90: Why not just write here `b"1"`?
Or did you really think to write `b"\0"` which would be byte array of 31 zero bytes followed by a byte value 49?

beamer/tests/contracts/test_request_manager.py
```diff
@@ -49,6 +48,7 @@ def test_claim(token, request_manager, claim_stake, deployer):
         request_manager, token=token, requester=requester, target_address=requester, amount=1
     )
 
+    print(request_id)
```
istankovic, line 51: Leftover print from debugging?

contracts/contracts/RequestManager.sol
```diff
@@ -55,12 +55,13 @@ contract RequestManager is Ownable, LpWhitelist {
     ///
     /// .. seealso:: :sol:func:`createRequest`
     event RequestCreated(
-        uint256 requestId,
+        bytes32 requestId,
```
istankovic, line 58: We should probably make the request ID `indexed`.

contracts/contracts/RequestManager.sol
```diff
@@ -138,11 +139,8 @@ contract RequestManager is Ownable, LpWhitelist {
     /// cannot be used to create new requests.
     bool public deprecated;
 
-    /// The request counter, used to generate request IDs.
-    uint256 public requestCounter;
-
-    /// The claim counter, used to generate claim IDs.
-    uint256 public claimCounter;
+    /// The nonce counter, used to generate request and claim IDs.
+    uint256 public nonceCounter;
```
istankovic, line 143: Nitpick: it would be nice to have a bit more detail of the expected usage of `nonceCounter`. Currently it is not obvious whether we are supposed to read the counter and then increment it, or first increment it and then read the value.

One might also consider changing the name so it's more clear, e.g. `nextNonce` or something.

istankovic (approved):
Looks fine, with few minor things.

contracts/contracts/RequestManager.sol
fredo: good point

beamer/tests/agent/test_request.py
fredo: In some occasions the tests didnt work because of that. But in this case the byte creation doesnt make a difference. I think it fails when we compare a contract keccak to a locally created one.

Issue #975, opened 2022-08-29
Make resolver address in FillManager immutable and settable


fredo:
To optimize the deployment scripts, we will set the resolvers address by a setter. 

Make the variable immutable so that it can only be set once and `onlyOwner`.

Adapt the deployment scripts.

PR #1018, opened 2022-09-13
Fillmanager set resolver address manually

fredo:
as part of the new deployment design, it is simpler for us to set the resolver manually.
Unfortunately the planned immutable did not work out, can only be set in the constructor.

But a simple require in `setResolver()` does the same job.

resolves: #975


contracts/contracts/FillManager.sol
```diff
@@ -54,13 +54,16 @@ contract FillManager is Ownable, LpWhitelist {
...
-        l1Resolver = _l1Resolver;
+    constructor(address _messenger) {
         messenger = IMessenger(_messenger);
     }
 
+    function setResolver(address _l1Resolver) public onlyOwner {
+        require(l1Resolver == address(0), "Resolver already set");
+        l1Resolver = _l1Resolver;
+    }
+
```
istankovic, line 66: Could you please document the function?

istankovic (changes requested):
Looking good, but we need to also
- document the function
- test that `fillRequest` and `invalidateFill` revert if the resolver is not set

istankovic (approved):
Looks good, thanks!

Issue #994, opened 2022-09-01
Merge RequestManager and Resolution Registry


fredo:
We now have a mapping from request hash => request in the Request Manager and request hash => fillIInfo in the Resolution Registry. We can now unify the mappings and merge the resolution registry into the Request Manager.

### Removing FillInfo
No need for this struct anymore as it is only used inside the Request Manager

### Storing Filler Info
The L1 resolved data can be stored directly in the request's struct.

The request struct is added

```
address filler
bytes32 fillId
``` 

### Storing invalid fillhashes
A fill hash is a tuple of request hash and fill id. The invalidation of such can be again stored in the request's struct. The request is already assigned to its key, the request hash.
Since there can be multiple fill id invalidations associated to the request, we need a mapping

request.invalidFillIds = mapping (fillId => bool)

- [ ] Update docs and adjust the graph
- [ ] Adapt deployment script
- [ ] Adapt ResolveClaim function

Issue #993, opened 2022-09-01
Remove fill hashes from FillManager


istankovic:
We need to
- [x] change the `fills` mapping to be `mapping(requestHash => fillId)`
- [x] rename the `HashInvalidated` event to `FillInvalidated` and remove the `fillHash` field from it
- [x] update the agent
- [ ] update the tests
- [ ] update docs

fredo:
also remove any references to `fill hash` in the docs

PR #1032, opened 2022-09-15
Unify Request Manager and Resolution Registry

fredo:
resolves: #994 
resolves: #993 

Documentation updates will be handled in #1033.


istankovic:
There is this in `beamer.events`:
```python
@dataclass(frozen=True)            
class FillHashInvalidated(TxEvent):
    fill_hash: FillHash            
```
This can now be removed.

beamer/state_machine.py
```diff
@@ -329,9 +329,13 @@ def _handle_request_resolved(event: RequestResolved, context: Context) -> Handle
...
 
-def _handle_hash_invalidated(event: HashInvalidated, context: Context) -> HandlerResult:
+def _handle_fill_invalidated(event: FillInvalidated, context: Context) -> HandlerResult:
     fill_block = context.fill_manager.web3.eth.get_block(event.block_number)
-    claims = _find_claims_by_fill_hash(context, event.fill_hash)
+    # FIXME: Remove with https://github.com/beamer-bridge/beamer/issues/994
+    request = context.requests.get(event.request_id)
+    assert request is not None
+
+    claims = _find_claims_by_fill_hash(context, request.fill_hash_with_fill_id(event.fill_id))
```
istankovic, line 338: This PR is said to resolve #993 and #994 so I'm wondering why we have this here.

contracts/contracts/RequestManager.sol
```diff
@@ -325,8 +327,8 @@ contract RequestManager is Ownable, LpWhitelist {
         );
         require(request.activeClaims == 0, "Active claims running");
 
-        request.withdrawInfo.filler = request.sender;
         request.withdrawn = true;
+        request.filler = request.sender;
```
istankovic, line 339: Why do we need to set the `filler` here? The filler should only be set by the L1 fill proof, right?

contracts/contracts/RequestManager.sol
```diff
@@ -706,4 +711,33 @@ contract RequestManager is Ownable, LpWhitelist, CrossDomainRestrictedCalls {
...
         deprecated = true;
     }
+
+    /// Mark the request identified by ``requestId`` as filled by ``filler``.
+    ///
+    /// .. note::
+    ///
+    ///     This function is callable only by the native L2 messenger contract,
+    ///     which simply delivers the message sent from L1 by the
+    ///     Beamer's L2 :sol:interface:`messenger <IMessenger>` contract.
```
istankovic, line 732: I would remove this entire note as it is no longer correct.

contracts/contracts/RequestManager.sol
```diff
@@ -738,6 +744,43 @@ contract RequestManager is Ownable, LpWhitelist, CrossDomainRestrictedCalls {
...
     }
+
+    /// Mark the fill identified by ``requestId`` and ``fillId`` as invalid.
+    ///
+    /// .. note::
+    ///
+    ///     This function is callable only by the native L2 messenger contract,
+    ///     which simply delivers the message sent from L1 by the
+    ///     Beamer's L2 :sol:interface:`messenger <IMessenger>` contract.
+    ///
```
istankovic, line 760: I would remove this note since it is no longer correct.

contracts/contracts/RequestManager.sol
fredo: you are right. this is a leftover from the time where we set the withdrawer which is now reflected by the `MAX_INT` assignment.

contracts/contracts/RequestManager.sol
fredo: why is this no longer correct?

contracts/contracts/RequestManager.sol
fredo: This also needs to be removed in withdrawDeposit()

contracts/contracts/RequestManager.sol
fredo: why is this no longer correct?

istankovic: Because it's not the native L2 messenger contract that is going to be calling this in general (e.g. Arbitrum).

contracts/contracts/RequestManager.sol
istankovic: Same thing as in https://github.com/beamer-bridge/beamer/pull/1032#discussion_r973053687.

contracts/contracts/RequestManager.sol
```diff
@@ -5,18 +5,25 @@ import "OpenZeppelin/openzeppelin-contracts@4.7.3/contracts/token/ERC20/IERC20.s
...
 /// This contract is responsible for keeping track of transfer requests,
 /// implementing the rules of the challenge game and holding deposited
 /// tokens until they are withdrawn.
+/// The information passed by L1 resolution will be stored with the respective requests.
 ///
 /// It is the only contract that agents need to interact with on the source chain.
-contract RequestManager is Ownable, LpWhitelist {
+/// .. note::
+///
+///   The function resolveRequest and invalidateFill can only be called by
```
istankovic, line 23: Nitpick: function -> functions

contracts/contracts/RequestManager.sol
```diff
@@ -738,6 +744,43 @@ contract RequestManager is Ownable, LpWhitelist, CrossDomainRestrictedCalls {
...
+        restricted(resolutionChainId, msg.sender)
+    {
+        Request storage request = requests[requestId];
+        require(
+            request.filler == address(0),
+            "Cannot invalidate resolved fills"
+        );
+        require(
+            request.invalidFillIds[fillId] == false,
+            "FillId already invalidated"
```
istankovic, line 773: Nitpick: I would say `""Fill already invalidated"`, to make it more consistent.

contracts/contracts/RequestManager.sol
```diff
@@ -556,35 +556,24 @@ contract RequestManager is Ownable, LpWhitelist, CrossDomainRestrictedCalls {
...
-        (address filler, bytes32 fillId) = resolutionRegistry.fillers(
-            claim.requestId
-        );
-
-        if (filler == address(0)) {
-            (filler, fillId) = (request.filler, request.fillId);
-        }
-
-        if (resolutionRegistry.invalidFillHashes(fillHash)) {
+        // Claim is valid/invalid if either
```
istankovic, line 551: Nitpick: I'd reword this as "The claim is resolved as follows."

istankovic:
The commit `3100af629707708f83c64a4a6e8e41580544e8fb` has typos in its message: ` chore: add new constats indicies to access filler and fillid ` (constats -> constants, indicies -> indices)

istankovic:
Nitpick: the message of commit `73c9090e0b8179ea7e87d267ef49483a4c5c21e7` has too long first line. Per git convention,
it would be better as
```
contracts: Remove valid request id check on L1 resolution.

This means request ids can be filled before they exist. This concept is equal to as it was in resolution registry.
```
(I've also added a space in `resolutionregistry`.)

beamer/tests/util.py
```diff
@@ -233,35 +233,3 @@ def create_request_id(
...
-):
-    return keccak(
-        encode_abi_packed(
-            ["bytes32", "bytes32"],
-            [
-                request_id,
-                fill_id,
-            ],
-        )
-    )
```
istankovic, line 267: This is just great.

contracts/contracts/BeamerUtils.sol
```diff
@@ -7,6 +7,9 @@ library BeamerUtils {
         bytes32 fillId;
     }
 
+    uint256 public constant MAX_INT =
+        0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
+
```
istankovic, line 12: Why not just use built-in `uint256.max` and get rid of the constant?

istankovic: Also, why does this need to be in `BeamerUtils`?

istankovic:
We should remove `FillInfo` struct, nobody sems to use it:
```
$ rg FillInfo contracts/
contracts/contracts/BeamerUtils.sol:5:    struct FillInfo {
$
```

istankovic:
We should remove `RM_R_FIELD_WITHDRAWN`, nobody sems to use it:
```
$ rg RM_R_FIELD_WITHDRAWN
beamer/tests/constants.py:10:RM_R_FIELD_WITHDRAWN = 8
$
```

contracts/contracts/BeamerUtils.sol
```diff
@@ -30,35 +30,4 @@ library BeamerUtils {
...
-                    sourceChainId,
-                    targetChainId,
-                    targetTokenAddress,
-                    targetReceiverAddress,
-                    amount,
-                    nonce
-                ),
-                fillId
-            );
-    }
```
istankovic, line 60: :parrot:

istankovic (changes requested):
Epic work, but there are a few minor things to cleanup still.

Issue #1035, opened 2022-09-16
Frontend: Detect blacklist addresses & prohibit usage of dApp


GabrielBuragev:
# Why should this feature exist?
Due to legal reasons, there is the need of maintaining a blacklist of wallet addresses which should be prohibited from using our dApp 

# Implementation
https://github.com/brainbot-com/PoWDump/pull/12

PR #1042, opened 2022-09-19
frontend: block blacklisted addresses

GabrielBuragev:
Fixes: #1035 

Blacklisted wallet addresses should have limited access to our dApp:
- Transfer creation should be prohibited, on input validation level, when target address is detected to be a part of the blacklist.
- Completely hiding the forms & all the action components and showing an error message when the wallet provider's address is detected to be in the blacklist.

The address blacklist is going to be manually maintained for now.


frontend/src/services/web3-provider/metamask-provider.ts
```diff
@@ -34,7 +34,7 @@ export class MetaMaskProvider extends EthereumProvider implements ISigner {
     try {
       const accounts: string[] = await this.web3Provider.send('eth_requestAccounts', []);
       this.signer.value = this.web3Provider.getSigner(accounts[0]);
-      this.signerAddress.value = accounts[0];
+      this.signerAddress.value = getAddress(accounts[0]);
```
GabrielBuragev, line 37: Previously, addresses were stored here as lowercased strings instead of proper checksumed addresses which explains the usage of `getAddress` here.

frontend/src/composables/useRequestTargetInputValidations.ts
```diff
@@ -27,6 +32,10 @@ export const useRequestTargetInputValidations = (
       selectedTargetAddress: {
         required: helpers.withMessage('Target address is required', required),
         isValidEthAddress: helpers.withMessage('Invalid ETH address', isValidEthAddress),
+        makeIsNotBlacklistedEthAddressValidator: helpers.withMessage(
```
manuelwedler, line 35: Is it necessary to have this twice? The dialog won't be loaded anyway and a user would not be able to create a transfer. Or am I missing something?

frontend/src/utils/addressBlacklist.ts
```diff
@@ -0,0 +1,284 @@
+export const blacklistAddresses = [
```
manuelwedler, line 1: Nitpick: I would use UPPER_CASE for such constants.

Also, I am not sure if the `utils` folder is the right place. One approach would be to move such constant values to the place where they are used (separate file but same folder), but would not work if we need the check twice. Or we could only export a utility function `isAddressBlacklisted` which makes the complete call `blacklistAddresses.includes(getAddress(address))`. would also reduce the redundant `includes` calls (depending on if it is necessary to have the check twice)

manuelwedler (commented):
Looks good, only a question and a little improvement that could be done.

frontend/src/composables/useRequestTargetInputValidations.ts
GabrielBuragev: This is for the cases where a user comes with a non-blocked wallet address and tries to bridge funds to the blocked address. I believe that these cases fall under the scope of suspicious activity as well and therefore i implemented this for extra safety.

manuelwedler: Ah, true. Didn't get it right. Now it makes sense to me :)

frontend/src/utils/addressBlacklist.ts
GabrielBuragev: :heavy_check_mark:

manuelwedler (approved):
Perfect :)

PR #1045, opened 2022-09-20
Fix agent events

istankovic:


beamer/models/request.py
```diff
@@ -49,7 +49,12 @@ def __init__(
 
     fill = pending.to(filled) | filled.to(filled) | ignored.to(filled) | claimed.to(claimed)
     try_to_fill = pending.to(filled)
-    try_to_claim = filled.to(claimed) | pending.to(claimed) | claimed.to(claimed)
+    try_to_claim = (
+        filled.to(claimed)
+        | pending.to(claimed)
+        | claimed.to(claimed)
+        | l1_resolved.to(l1_resolved)
```
fredo, line 56: this can happen when someone challenges the agent's claim after it was already l1_resolved

fredo (approved):
Some logic things are missing when it comes to the point that claims can be made after l1 resolution happened. 

We will most likely have redeployed when this can happen but we should not forget about it. Especially when we add rollups with lower finality periods.

PR #1054, opened 2022-09-21
handle own claim made event after fill 

fredo:
handle own ClaimMade event only after corresponding RequestFilled event was processed


beamer/state_machine.py
```diff
@@ -317,11 +323,8 @@ def _handle_claim_made(event: ClaimMade, context: Context) -> HandlerResult:
 
 def _handle_claim_stake_withdrawn(event: ClaimStakeWithdrawn, context: Context) -> HandlerResult:
     claim = context.claims.get(event.claim_id)
-
-    # Check if claim exists, it could happen that we ignored the request because of an
-    # invalid token pair, and therefore also did not create the claim
     if claim is None:
-        return True, None
+        return False, None
```
istankovic, line 331: Eh, this can potentially result in a big number of events accumulating over time in an agent that does not support the token pairs involved. I'm not sure how to address that.

beamer/state_machine.py
```diff
@@ -272,11 +272,14 @@ def _handle_claim_made(event: ClaimMade, context: Context) -> HandlerResult:
     claim = context.claims.get(event.claim_id)
 
     if claim is None:
+        if event.last_challenger != ADDRESS_ZERO:
+            return False, None
+
```
istankovic, line 277: The commit message of `f19f7c434081f1402c77fa95762c01600a758740` says " agent: Only withdraw if there are no ClaimMade events in loop " but I do not understand how this relates to the above change. Could you please explain?

beamer/state_machine.py
fredo: That's true. I haven't better solution though.

beamer/state_machine.py
fredo: The relation is the following: 

There is a change above which don't processes ClaimMade Events of the agent if the fill isnt there yet. If that is the case then a second (or third, etc) ClaimMade event will be processed before the first one. 
We need to also push back those events, namely when a claim is no created yet, but we receive ClaimMade events with challengers in it. This means the initital ClaimMade event must have been pushed back, so also push back the others

beamer/state_machine.py
fredo: I'm working on a solution for this. Please do not merge yet.

beamer/state_machine.py
istankovic: Ah, that totally explains it. Thank you for such a nice explanation!

istankovic (approved):

PR #1058, opened 2022-09-21
frontend: Adjust protocol related code to reflect latest contract changes

### LINKED ISSUE #1025: Frontend: Adjust protocol related code to reflect latest contract changes
Due to refactoring of some of the contracts, the frontend code needs to be changed accordingly.
Most importantly, the PRs #1016 & #1032 are introducing changes to the `RequestManager` & `FillManager` contracts which affects the communication between the frontend and these contracts.

Common changes affecting all the files below:
- `requestId` type changed `uint256` -> `bytes32`.
- `withdrawInfo` struct was removed from `RequestManager` and a `withdrawn` boolean was introduced instead.

Update `fill-manager.ts` service
--
- `RequestFilled` - event filter parameter types
- `requestIdentifier` - type of all occurences

Update `request-manager.ts` service
--
- `ClaimStakeWithdrawn` - event filter parameter types
- `contract.requests(requestId)` - parameter type
- `contract.createRequest()` - return value
- `contract.withdrawExpiredRequest()` - parameter type
- `getRequestIdentifier()` - return type
- `requestIdentifier` - type of all occurences
- `withdrawInfo` - remove all occurences & replace with `withdrawnClaimId` instead

Update `request-information.ts`
--
- `_identifier` - class property type change & all occurences
- `identifier()` - return type
- `setIdentifier()` - parameter type

Update `transfer.ts` class
--
- remove all occurences of `withdrawInfo` & `filler`
- update `checkAndUpdateWithdrawState` to use `withdrawn` bool for checking
- remove `fulfillmentInformation` completely
- remove TODO's related to setting fulfillmentInformation

## Adjust tests accordingly

GabrielBuragev:
Fixes #1025 

Its discussable whether we should also remove the `fulfillmentInformation` from the code. 
It was nowhere used nor there is any plan to use it in the near future.
The only critical question is: is the code still backward compatible?

IMO, since no logic depended on this information except the tests, it is safe to say that it wont break compatibility.
Also the "encoding"/"decoding" of the persistent store will not break since we are just removing a field.
Would love to hear others opinions.

I removed it in this PR but if the discussions lead to bringing it back will do so.


GabrielBuragev:
Also spotted & fixed an edge case where the wallet provider is not properly initialized as a ref inside `Transfer.vue` component and therefore making withdraw requests fail when the user comes to the app directly to the "Activity" tab. 
To reproduce this you need to create & wait for a request to expire - shows up as expired and ready for withdrawing on the UI. After that refresh the page with `/#/?activeTabLabel=Activity` and try withdrawing funds. 
You should get this:
image

manuelwedler:
I never questioned the `fulfillmentInformation` because I thought Thore added it with a specific feature in mind. I'm in for removing it. There is a todo comment in the `transfer.ts`: `// TODO: set this.fulfillmentInformation` . It was never in use as far as I can see. Also it is an optional property of a transfer. Probably, never a transfer with this property was put into anyone's local storage. Anyway removing an property should not break anything as long as we just ignore unexpected fields.

manuelwedler:
These changes will definitely not work with local storage data of an earlier version of the dapp. An older transfer id cannot get parsed as the new byte based identifier for example. The old transfer would not be in the new contract anyway. But still we need to make sure that old local storage data is not accessed in a new dapp deployment.
How should we address this issue? I would like to find a solution to this before this is merged.

Interestingly, when we designed the transfer storage we considered so many cases where the frontend could have breaking changes for the local storage, but we never considered breaking changes from the smart contracts.

This makes my earlier comment about the `fulfillmentInformation` a bit irrelevant as we have breaking changes here anyway.

manuelwedler (commented):
The changes look fine to me, just as specified in the linked issue. I would only like to find an approach solving my earlier comment.

I just saw your discussion on rocketchat. Using adding a version to the url for the different deployments would work. I only think we cannot move the local storage data from one url to another. So the newest deployment would always need a new url. But don't know if there is a way around this.

manuelwedler (approved):
Approving since we have a basic strategy for branching and versioning now. It still might change if we decide to split repos, but this can still be approved now.

PR #1059, opened 2022-09-22
Frontend: fix tooltip for mobile

### LINKED ISSUE #987: "?" on webui does nothing on mobile
#### System information

OS & Version: Andoird
Browser & Version: Chrome

#### Expected behavior
If I click on the "?" on the webui from my phone i should get some info

#### Actual behavior
Not clickable at all

Open the webui on your phone, click on "?", nothing happens

manuelwedler:
Fixes #987

This does two things:
1. position the tooltip conditionally on the right of the main container or below the content element depending on if it fits in the viewport. (The tooltip did not have enough space on the right on mobile)
2. add the touch events to also make it work in an mobile environment.

### DIFF
--- a/frontend/tests/unit/components/layout/Tooltip.spec.ts
+++ b/frontend/tests/unit/components/layout/Tooltip.spec.ts
@@ -4,8 +4,7 @@ import Tooltip from '@/components/layout/Tooltip.vue';
 
 async function createWrapper(options?: {
   hint?: string;
-  position?: string;
-  maxTooltipWidth?: string;
+  tooltipWidth?: string;
   gap?: string;
   arrowSize?: number;
   defaultSlot?: string;
@@ -20,8 +19,7 @@ async function createWrapper(options?: {
     },
     props: {
       hint: options?.hint,
-      position: options?.position,
-      maxTooltipWidth: options?.maxTooltipWidth,
+      tooltipWidth: options?.tooltipWidth,
       gap: options?.gap,
       arrowSize: options?.arrowSize,
     },
@@ -59,7 +57,7 @@ describe('Tooltip.vue', () => {
     expect(tooltip.exists()).toBeFalsy();
   });
 
-  it('shows tooltip on hover', async () => {
+  it('allows controlling tooltip visibility based on mouse events', async () => {
     const wrapper = await createWrapper();
 
     expect(wrapper.find('[data-test="tooltip"]').exists()).toBeFalsy();
@@ -86,11 +84,11 @@ describe('Tooltip.vue', () => {
     expect(tooltip.html()).toContain('<span>test hint</span>');
   });
 
-  it('sets maximum width to tooltip', async () => {
-    const wrapper = await createWrapper({ maxTooltipWidth: '200px', showTooltip: true });
+  it('sets width to tooltip', async () => {
+    const wrapper = await createWrapper({ tooltipWidth: '200px', showTooltip: true });
     const tooltip = wrapper.get('[data-test="tooltip"]');
 
-    expect((tooltip.element as HTMLElement).style).toContain({ maxWidth: '200px' });
+    expect((tooltip.element as HTMLElement).style).toContain({ width: '200px' });
   });
 
   it('adapts the arrow size', async () => {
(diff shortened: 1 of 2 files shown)


frontend/tests/unit/components/layout/Tooltip.spec.ts
```diff
@@ -69,6 +69,16 @@ describe('Tooltip.vue', () => {
     expect(wrapper.find('[data-test="tooltip"]').exists()).toBeFalsy();
   });
 
+  it('shows tooltip on touch', async () => {
+    const wrapper = await createWrapper();
```
GabrielBuragev, line 73: Can we rename this test case to something like "allows controlling tooltip visibility based on touch events" since it covers both touchstart & touchend events?
Same goes for the hover test case above.

frontend/tests/unit/components/layout/Tooltip.spec.ts
```diff
@@ -5,7 +5,7 @@ import Tooltip from '@/components/layout/Tooltip.vue';
 async function createWrapper(options?: {
   hint?: string;
   position?: string;
```
GabrielBuragev, line 7: Can we remove all the occurences of `position` in the test file since we removed it from the props definition of the component?

frontend/tests/unit/components/layout/Tooltip.spec.ts
```diff
@@ -100,65 +110,24 @@ describe('Tooltip.vue', () => {
...
-     * in the test environment, the `getBoundingClientRect()` of an HTML element
-     * always returns zero for all properties.
-     */
-
-    it('sets gap as horizontal margin to tooltip', async () => {
-      const wrapper = await createWrapper({ gap: '20px', position: 'left', showTooltip: true });
-      const tooltip = wrapper.get('[data-test="tooltip"]');
-
-      expect((tooltip.element as HTMLElement).style).toContain({ margin: '0px 20px' });
+  describe('position bottom when not fitting in viewport', () => {
```
GabrielBuragev, line 113: I think we can rephrase & restructure the describe block here like so: 
- "when not fitting in viewport" 
  - "it positions itself on the bottom"
  - "sets gap ..." 
  - ...
- "when fitting in viewport"
  - "it positions itself on the right"
  - "sets gap ..."
  - ...

When nesting such `describe` blocks i always try to imagine the `describe` block as an environment under which the tests inside of it are made unless the `describe` block is targeting/defining a function/prop/variable name or a logic scope. 
If i imagine it as an environment i always end up in phrases like "when ...." which provides higher value to the tester/user running the tests.
In this case, i also believe the `'position bottom when not fitting in viewport'` is also a testable case and should not be used as a phrase for the `describe` block but instead there should be a test for the `position`.

What do you think?

GabrielBuragev (changes requested):
I love how you splitted the logic into smaller computed properties. Makes it way easier to understand whats going on!

I have a few small suggestions on the testing side regarding the phrases & nesting and spotted one bug on mobile devices.

GabrielBuragev:
On mobile devices, when you click on the tooltip indicator for the first time it correctly opens and closes. After the first try, every other try, the tooltip is forcibly closed directly on each open attempt.


https://user-images.githubusercontent.com/18466333/192268597-df9fc982-5d5f-457d-b07f-e00e5f075190.mp4

frontend/tests/unit/components/layout/Tooltip.spec.ts
manuelwedler: Sounds like a good approach to me. I just adapted the old tests and didn't think to much about the naming. But I like your approach.

frontend/tests/unit/components/layout/Tooltip.spec.ts
manuelwedler: Somehow missed some...

frontend/tests/unit/components/layout/Tooltip.spec.ts
manuelwedler: sure :heavy_check_mark:

frontend/tests/unit/components/layout/Tooltip.spec.ts
manuelwedler: Only thing is that I think it is not really possible to test the position because `getBoundingClientRect` does always return only 0 for all properties, like Thore's comment suggests. But will think again about it.

manuelwedler: With the behavior of the `getBoundingClientRect` I couldn't find a way to to test the position properly. With mocking I feel like I just end up testing the mocked value. But added the expected position as a comment in the tests.

frontend/tests/unit/components/layout/Tooltip.spec.ts
manuelwedler: :heavy_check_mark:

manuelwedler:
Actually the expected behavior is that you have to hold the button. Looks like the mouse events are also fired on mobile what is leading to the weird behavior. I just removed the touch events so when you touch it once it stays on the screen and on the second touch it is hidden.

GabrielBuragev:
Everything looks good now. Also mobile works as expected! 
Thanks for the edits.

GabrielBuragev (approved):

frontend/tests/unit/components/layout/Tooltip.spec.ts
GabrielBuragev: That's understandable. No need to waste more time on that. :heavy_check_mark:

Issue #1049, opened 2022-09-21
Set minimum fee to $1


czepluch:
# Rationale

The gas prices on rollups have changes quite a lot since we initially created the fee model. We should set the minimum fee lower than $5. Currently 1 tx on rollups costs around $0.10 and an agent does 3 per fill. So have some room for fluctuating gas prices, it is suggested that the minimum fee is set to $1.

# Implementation

- [x] Update the minimum fee in the mainnet deployment to be 1 USDC.
- [x] Add "The minimum fee for a transfer is 1 USDC" to the popup text. 
Implementation: (#1082)
- [x] Adapt docs in section `agent fee` in whitepaper

andfletcher:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-62442a69c5e6e60010e31cf3/issues/beamer-bridge/beamer/1049?planning-poker) with Zenhub @fredo @istankovic

andfletcher:
@GabrielBuragev please be aware that I added a component for this issue after aligning with Jacob.

We need the sentence "The minimum fee for a transfer is 1 USDC"  at the end of the popup (when you hover the mouse over "?")

andfletcher:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-62442a69c5e6e60010e31cf3/issues/beamer-bridge/beamer/1049?planning-poker) with Zenhub @GabrielBuragev

GabrielBuragev:
@andfletcher Should we create 2 separate issues and add them in this issue description and estimate them separately? 
Since this issue covers work on both frontend & backend.

andfletcher:
> @andfletcher Should we create 2 separate issues and add them in this issue description and estimate them separately? Currently this issue covers work on both frontend & backend.

i dont think its worth it. Its a good practice on how we can swarm on issues together to finish them since both acceptance criteria are unrelated but still go towards the same value.

GabrielBuragev:
Where are the fees defined?

andfletcher:
> defined

This is a good point. But currently, we only support USDC, so it should not matter for now.

andfletcher:
> Where are the fees defined?

I dont know but why do you need to know where the fees are defined (just asking)

czepluch:
They are defined in the contract. So you should be able to pull the number from there. Especially since we might need to manually update it in the future based on gas prices.

GabrielBuragev:
[...]

PR #1061, opened 2022-09-22
Update frontend mainnet

GabrielBuragev:


GabrielBuragev:
Draft until #1049 is finished and the solution is included here.

istankovic (approved):
Well, at least the commit messages look good to me. :-)

GabrielBuragev:
I would also like to include #1066 here since i found it to be pretty important as users connected with external desktop wallets (via WalletConnect) might find themselves left in the dark and not being able to disconnect easily.
If deadline for launch does not allow that then this will be merged as it is right now.

PR #1065, opened 2022-09-22
Store invalid fill ids in request

fredo:
We need to store invalid fillIds in the request object. This is due to the fact, that due to not processing preceding events, the claim creation can happen after processing invalidation events.

### DIFF
--- a/beamer/models/request.py
+++ b/beamer/models/request.py
@@ -37,8 +37,10 @@ class Request(StateMachine):
         self.fill_tx: Optional[HexBytes] = None
         self.fill_timestamp: Optional[Timestamp] = None
         self.fill_id: Optional[FillId] = None
+        self.invalid_fill_ids: dict[FillId, tuple[HexBytes, int]] = {}
         self.l1_resolution_filler: Optional[ChecksumAddress] = None
         self.l1_resolution_fill_id: Optional[FillId] = None
+        self.l1_resolution_invalid_fill_ids: set[FillId] = set()
 
     pending = State("Pending", initial=True)
     filled = State("Filled")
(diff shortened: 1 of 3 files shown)


beamer/state_machine.py
```diff
@@ -351,13 +351,31 @@ def _handle_request_resolved(event: RequestResolved, context: Context) -> Handle
...
             request.l1_resolve(event.filler, event.fill_id)
+
+            discard_fill_ids = list()
+            for invalid_fill_id in request.invalid_fill_ids:
+                if event.fill_id == invalid_fill_id[0]:
+                    discard_fill_ids.append(invalid_fill_id)
+
+            for invalid_fill_id in discard_fill_ids:
+                request.invalid_fill_ids.discard(invalid_fill_id)
+
```
istankovic, line 369: This could be more concise:
```python
request.invalid_fill_ids.difference_update(value for value in request.invalid_fill_ids 
                                                            if value[0] == event.fill_id)
```

beamer/models/request.py
```diff
@@ -37,8 +37,10 @@ def __init__(
         self.fill_tx: Optional[HexBytes] = None
         self.fill_timestamp: Optional[Timestamp] = None
         self.fill_id: Optional[FillId] = None
+        self.invalid_fill_ids: set[tuple[FillId, HexBytes, int]] = set()
```
istankovic, line 40: I'm thinking this might be better as a mapping `FillId` -> `tuple[HexBytes, int]`.

istankovic (approved):

fredo:
I created a new issue for the discussion around the HashInvalidated event

PR #1066, opened 2022-09-22
frontend: Add wallet disconnect feature & refactor home view

### LINKED ISSUE #1041: Frontend: Implement wallet disconnect feature
Currently, there is no proper wallet "disconnection" option for the user on the app's UI.
There is only one way the user can disconnect his wallet from our site and that is by tapping into the wallet provider extension/app configuration.

In order to improve UX we need to implement a more straight-forward & prominent flow for the users to be able to connect/disconnect from their chosen wallet providers.

UI design for this feature [is ready](https://www.figma.com/file/Tx5V8SUrD1HrvDVqX7iEjn/Beamer-App?node-id=808%3A421)

GabrielBuragev:
Resolves #1041 

While working on it, i seized the opportunity to slightly refactor the `Home.vue` code. I scattered the code inside it related to the wallet connection status into their own separate components. This chores will be very appreciated later on when we need to prepare a full test setup.
Also provided tests for all the new components & new features.

The folder/file naming could be done better IMO. Happily open for discussions/suggestions!

### DIFF
--- a/frontend/src/components/WalletConnectionDetails.vue
+++ b/frontend/src/components/WalletConnectionDetails.vue
@@ -1,0 +1,20 @@
+<template>
+  <div class="h-14">
+    <div v-if="signer" class="flex flex-row gap-4 justify-center items-center" data-test="details">
+      <div class="h-7 w-7 rounded-50 border-[3px] border-solid border-sea-green bg-green"></div>
+      <div class="flex flex-col w-full">
+        <span class="text-lg">You are currently connected</span>
+        <Disconnect> </Disconnect>
+      </div>
+    </div>
+  </div>
+</template>
+<script lang="ts" setup>
+import { storeToRefs } from 'pinia';
+
+import Disconnect from '@/components/wallet/Disconnect.vue';
+import { useEthereumProvider } from '@/stores/ethereum-provider';
+
+const ethereumProvider = useEthereumProvider();
+const { signer } = storeToRefs(ethereumProvider);
+</script>
--- a/frontend/src/components/WalletConnectionDetails.vue
+++ b/frontend/src/components/WalletConnectionDetails.vue
@@ -1,5 +1,11 @@
 <template>
-  <div class="h-14">
+  <div class="h-28 flex flex-col justify-center items-center">
+    <ConnectionError v-slot="{ message }" data-test="error-component">
+      <div class="text-center text-red p-2 text-lg h-12">
+        {{ message }}
+      </div>
+    </ConnectionError>
+
     <div v-if="signer" class="flex flex-row gap-4 justify-center items-center" data-test="details">
       <div class="h-7 w-7 rounded-50 border-[3px] border-solid border-sea-green bg-green"></div>
       <div class="flex flex-col w-full">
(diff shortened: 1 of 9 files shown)


frontend/src/components/wallet/Disconnect.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <slot data-test="disconnect-trigger" v-bind="{ disconnect }"> </slot>
```
manuelwedler, line 2: Hm, I think this is maybe split in a too complex way. I don't see why a function from a composable should be passed to its parent via the slot props in this context. This component does not do anything else than passing the function to the parent. The parent could just call the composable itself. A component not adding any actual HTML also looks wrong to me.

I think the right approach here would be to just move the `span` element, which is in the `Disconnect` component in the parent at the moment, to the `Disconnect` component directly (see my other comment). As a result you would not need to pass the disconnect function up to the parent.

frontend/src/components/WalletConnectionDetails.vue
```diff
@@ -0,0 +1,34 @@
...
+        {{ message }}
+      </div>
+    </ConnectionError>
+
+    <div v-if="signer" class="flex flex-row gap-4 justify-center items-center" data-test="details">
+      <div class="h-7 w-7 rounded-50 border-[3px] border-solid border-sea-green bg-green"></div>
+      <div class="flex flex-col w-full">
+        <span class="text-lg">You are currently connected</span>
+        <Disconnect v-slot="{ disconnect }">
+          <span
```
manuelwedler, line 14: I think this span element should just be inside the `Disconnect` component. Would make it simpler.

manuelwedler (changes requested):
I like how you always try to take refactoring opportunities and make our code base cleaner. This looks fine overall. See only my one comment. You might have been making it to complex at one point here.

frontend/src/components/wallet/Disconnect.vue
manuelwedler: > The parent could just call the composable itself

I think composables are a way to make code reusable, so it is a good idea to use them in multiple places. In the best case their interface / parameters just don't change. Anyway, I don't prefer this approach here.

> I think the right approach here would be to just move the span element

Hm, I would say just write another component that renders it as a button. Use the composable there, as we have it for reusing logic.
You could also just use the `button` tag anyway here and style without the look of a button. Would actually make sense semantically. And then let the parent component add styling if we really need it in multiple places.
Moving the `span` (or `button` if you like to change it) into the `Disconnect`  component is still my preference.

Having the logic split among multiple components feels not easily understandable to me. There is even a name for the pattern you used here. [Renderless Components](https://vuejs.org/guide/reusability/composables.html#vs-renderless-components). Only thing is that you mixed it with a composable. In the link, the vue documentation also not recommends this pattern:
> The recommendation is to use composables when reusing pure logic, and use components when reusing both logic and visual layout.

GabrielBuragev: Alright, switched to the suggested approach. Thanks for taking the time to elaborate, this was a valuable discussion for me.

GabrielBuragev:
Implemented the requested changes :heavy_check_mark:

manuelwedler (approved):
Looks good! Thanks for making the changes!

PR #1082, opened 2022-09-27
frontend: implement dynamic min. fee notification & refactor `RequestSourceInputs.vue`

GabrielBuragev:
Solves the frontend part of issue #1049.

Once again, i saw the opportunity to refactor the `RequestSourceInputs` and split some of the tooltip related code to its own component (since its not a plain html anymore).

Provided tests for the newly created components as well as composables.

### DIFF
--- a/frontend/src/components/RequestFeeTooltip.vue
+++ b/frontend/src/components/RequestFeeTooltip.vue
@@ -1,0 +1,38 @@
+<template>
+  <Tooltip class="-mr-3" show-outside-of-closest-reference-element>
+    <div class="h-full flex flex-col justify-center" data-test="tooltip-trigger">
+      <img class="h-6 w-6 cursor-help" src="@/assets/images/help.svg" />
+    </div>
+    <template #hint>
+      <div data-test="tooltip-text">
+        The fee amount is composed of two parts:<br />
+        <ul class="pl-5">
+          <li>• the liquidity provider fee</li>
+          <li>• the protocol fee</li>
+        </ul>
+        <br />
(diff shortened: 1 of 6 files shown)


frontend/src/components/RequestFeeTooltip.vue
```diff
@@ -0,0 +1,45 @@
...
+import { toRefs } from 'vue';
+
+import Tooltip from '@/components/layout/Tooltip.vue';
+import { useMinRequestFee } from '@/composables/useMinRequestFee';
+import type { Chain, Token } from '@/types/data';
+
+interface Props {
+  chain: Chain | undefined;
+  token: Token | undefined;
+}
```
manuelwedler, line 40: I'm thinking if it would make sense to call the `useMinRequestFee` composable in the `RequestSourceInputs` component and just pass the `formattedMinFee` as a prop. This would make the `RequestFeeTooltip` a purely visual component with no logic at all. We were trying to achieve this distinction in the past, see the `TransferSummary` for example.

manuelwedler (changes requested):
Just one idea, other than that looks good :)

frontend/src/components/RequestFeeTooltip.vue
GabrielBuragev: Done :heavy_check_mark:

GabrielBuragev:
We should also soon focus on grouping the components inside folders since we started introducing new components thanks to the refactoring in the recent PRs. This chore can be done as a part of the upcoming tasks related to writing tests.

manuelwedler (approved):
thanks for the change. Looks great!

PR #1087, opened 2022-09-28
frontend: recognize internal wallet disconnect

### LINKED ISSUE #692: Frontend: Detect wallet disconnection from inside the wallet provider
Currently, the user can disconnect the wallet from the wallet itself but such disconnects are not being recognized by our app (state is not being updated)
We should recognize that properly.

There is a simple event we can listen to. So if the user disconnects, we recognize this via an event handler in the EthereumProvider, we can unset the signer and most of the app is fine and allows to reconnect.

**Extra:**
We also might want to update the state that keeps the last connected wallet (used for auto-reconnect on page load). A reset to this state will make sure we dont spam the user, on every page load, when he chooses to disconnect from one wallet provider.
This requires a new relation. This means we either need to observe a reference of the provider instance or the provider must become an event emitter.

GabrielBuragev:
Fixes #692 

Read commit description.

### DIFF
--- a/frontend/src/composables/useWallet.ts
+++ b/frontend/src/composables/useWallet.ts
@@ -21,6 +21,7 @@ export function useWallet(
       metaMaskProvider.requestSigner();
       provider.value = metaMaskProvider;
       connectedWallet.value = WalletType.MetaMask;
+      metaMaskProvider.on('disconnect', disconnectWallet);
     } else if (withOnboarding) {
       onboardMetaMask();
     }
@@ -32,6 +33,7 @@ export function useWallet(
     if (walletConnectProvider) {
       provider.value = walletConnectProvider;
       connectedWallet.value = WalletType.WalletConnect;
+      walletConnectProvider.on('disconnect', disconnectWallet);
     }
   }
 
@@ -45,6 +47,11 @@ export function useWallet(
     }
   }
 
+  async function disconnectWallet() {
+    connectedWallet.value = undefined;
+    provider.value = undefined;
+  }
+
   const metamaskTask = useAsynchronousTask(connectMetaMask);
   const walletConnectTask = useAsynchronousTask(connectWalletConnect);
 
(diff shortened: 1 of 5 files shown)


manuelwedler (approved):
I'm thinking if it would be better to use a watcher in the `useWallet` composable instead of the event emitter api. We could just add a `connected` property to the providers and watch it inside the composable. I just feel like this could be the more vue-style approach. Since your solution looks perfectly fine, I'm approving. Feel free to make the change if you like it better.

GabrielBuragev:
So, i tried several different approaches for solving this:
- Introducing a new `connected` property which holds the connection status value inside the `EthereumProvider` class and then watching on that property inside the `useWallet` composable. This didn't result in a working solution since we use `shallowRef` when storing the `provider` inside the store and that means deeply nested refs are not reactive.
- Introducing a new `connected` getter property inside `ethereumProvider` store which watches on either the `connected` property inside the class or the `signerAddress` (checking for falsy values). Using this approach means we will somehow need to pass the getter to the `useWallet` composable in order to be able to watch on it and react.
- Other solutions on the store level

All of these solutions introduce a side effect which is not explicit and easy to understand as to why its needed. The `EventEmitter` approach is way more expressive IMO so thats why i decided on keeping this as it is. 
A small refactor to the `useWallet` composable can be done to remove duplication.

manuelwedler:
Fine for me that you merged it as I said your approach works perfectly. 
Just to inform you how the first bullet would have worked: wrapping the `connected` property in a `ref` again would be the solution. The `IEthereumProvider` interface already uses a lot of properties wrapped inside refs because of the outer `shallowRef`. The reason for the `shallowRef` is simply that we use complex library objects inside the providers and they didn't work together with a normal `ref`.

GabrielBuragev:
I tried exactly that but unfortunately i didn't come to a very clean & straight forward solution in the end because the watcher was firing with same new & old values of the provider's `connected` property (they were either both `false` or both `true`) so i couldn't do a proper check when a `disconnect` has happened (transition from `true` to `false` of `connected` ref). :confused: 
I thought the reason for this behavior was the usage of `shallowRef` for storing the `provider` in the store. Thanks for the hints though, i will definitely come back to this `composable` for refactoring so we can find a better solution for it.

PR #1097, opened 2022-09-30
docs: document the release process

istankovic:


istankovic:
@fredo could you please review?

docs/source/releasing.rst
```diff
@@ -0,0 +1,11 @@
+Making a new release
+--------------------
+
+* Make sure to populate ``CHANGELOG.md`` and change the version number
+  (see `this commit <https://github.com/beamer-bridge/beamer/commit/440b7ddffc01d16482d78ff9f18a8830670795bc>`_ for example).
+* Commit and push the changes above to Github, merge the PR and wait for the CI run to finish.
+* The previous step will have created a Docker image, which you can use for final testing before the release.
+  Once everything looks good, tag the commit (e.g. ``git tag v0.1.8``) and push it to Github (e.g. ``git push tag v0.1.8``).
```
fredo, line 8:
```suggestion
  Once everything looks good, tag the commit (e.g. ``git tag v0.1.8``) and push it to Github (e.g. ``git push --tag v0.1.8``).
```

docs/source/releasing.rst
```diff
@@ -0,0 +1,11 @@
+Making a new release
+--------------------
+
+* Make sure to populate ``CHANGELOG.md`` and change the version number
+  (see `this commit <https://github.com/beamer-bridge/beamer/commit/440b7ddffc01d16482d78ff9f18a8830670795bc>`_ for example).
+* Commit and push the changes above to Github, merge the PR and wait for the CI run to finish.
+* The previous step will have created a Docker image, which you can use for final testing before the release.
```
fredo, line 7: It would be nice to mention the docker image and running final tests references to the agent.

docs/source/releasing.rst
fredo: > > Do you want to release this to the public documentation?
> 
> Sure, why not, I don't see a reason to keep it private.
> 
> > Also should we document the process of creating docker images by hand?
> 
> Eh, I'd rather not at this point, I'm still hoping we get to CirceCI in a reasonable amount of time. sweat_smile

What I mean was not about the CI, because the CI might not run on specific branches. Our problem we had was that we wanted to test something before merging it to main (merging it to the stable branch). If we want our CI to produce images, then we also need to enable it for other branches than main

istankovic: > Your version interprets tag as the remote name

Gah, good catch, that's a typo. :+1:

docs/source/releasing.rst
fredo: `The previous step will have created a Docker image, which you can use for final testing before the release.`
 
 What are you referring to here? I thought you are talking about an docker image for the agent which you can use for final testing. Is that right?

fredo: I would just like to mention that it is the docker image of the agent. I'm saying that because in our mono repo lots of different modules are there (currently frontend, contracts, agent). Being a little bit more explicit would be nice I think

istankovic: > I would just like to mention that it is the docker image of the agent. I'm saying that because in our mono repo lots of different modules are there (currently frontend, contracts, agent). Being a little bit more explicit would be nice I think

Ah, I understand now, yeah, that's a good idea. :+1:

docs/source/releasing.rst
istankovic: > What I mean was not about the CI, because the CI (or rather building an image) might not run on specific branches. Our problem we had was that we wanted to test something which was not supposed to be merged on main (merging it to the stable branch). If we want our CI to produce images, then we also need to enable it for other branches than main

I guess we will anyway want to have this in the CI for `main` and all the "stable" branches (e.g. `0.x`, `1.x`...). In that case it won't be an issue.

Making a Docker image is not really a big deal, however we need to push the image to ghcr in order to make it available to our DO test setup. So I would say we have this

# releasable branches (`main`, `0.x`, `1.x` etc.): CI will create images and make them available in the registry
# non-releasable branches: CI will not create images; instead, if one wants to test, one should go to DO and manually make an image there and use that image for tests

The latter option is supported since a long time ago, I tested it many times and it should still work. But in any case, for _a release_ the CI should make sure there is an image and the release docs (this item) don't need to mention anything special.

fredo:
looks good!

Issue #1086, opened 2022-09-28
Query L1 gas price at most once per L1 block


istankovic:
Once an `InitiateL1Resolution` event is created and inserted into the event loop, it will cause a gas price RPC call approx. every second, which is not desirable nor actually needed due to the fact that block time is ~15s on Ethereum mainnet. The situation gets worse as the number of concurrent challenge games increases. To mitigate this, we should cache gas price by using `web3.middlewares.latest_block_based_cache_middleware`.

- [x] fix the bug introduced in c22aa1483ca4403fc8669ca0996f0d1336837218 by adding a custom middleware for caching `eth_getBlockByNumber` calls
- [ ] use `web3.middlewares.latest_block_based_cache_middleware` for the L1 `web3.Web3` instance [`0.x` branch]
- [ ] apply the same change to the `main` branch

istankovic:
Seems like `web3.middlewares.latest_block_based_cache_middleware` does not really work, possibly due to https://github.com/ethereum/web3.py/pull/2346 (which is not in 5.29.1, which we are using).

istankovic:
If we periodically query the gas price and do nothing else, it seems the middleware will always return the cached result instead of issuing a proper request as needed. This happens because the middleware make a call `web3.eth.get_block('latest')` which returns _a cached_ block, not the latest block, i.e. it seems that latest block is not properly updated.

istankovic:
Ok, I found the explanation for the wrong behaviour: we added caching of block data in c22aa1483ca4403fc8669ca0996f0d1336837218.
Where we also cache the eth_getBlockByNumber method. 🤦🏼‍♂️
I verified the latest block based cache middleware works as expected with the "simple cache" middleware removed.

So the solution in c22aa1483ca4403fc8669ca0996f0d1336837218 was overly simple and missed the fact that the query for the latest block number goes via that same call too. (One wonders how did anything work at all...)
So that's good at least -- no bug in web3py, rather a bug in our agent.

Since it would nice to retain the benefits of the simple cache middleware, a solution would be to implement our own middleware for eth_getBlockByNumber that forwards the request in case of the latest call and remove the addition of eth_getBlockByNumber to the simple cache RPC whitelist.

istankovic:
Reopening since the merged PR addresses only the first subtask -- we still need to implement the others.

istankovic:
@fredo please review @1140 and #1141.

istankovic:
This is now completed.

PR #1117, opened 2022-10-07
Fix simple cache

istankovic:
Fixes the problem described in https://github.com/beamer-bridge/beamer/issues/1086#issuecomment-1270735862.
Part of #1086.

### DIFF
--- a/beamer/middleware.py
+++ b/beamer/middleware.py
@@ -1,0 +1,43 @@
+from typing import Any, Callable
+
+import lru
+from web3 import Web3
+from web3.types import RPCEndpoint, RPCResponse
+
+
+def _result_ok(response: RPCResponse) -> bool:
+    if "error" in response:
+        return False
+    return response.get("result") is not None
+
+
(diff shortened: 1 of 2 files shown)


istankovic:
@fredo could you please review?

istankovic:
>  But this does not solve the original problem that we still get ~12 requests per block on L1 (84000 times per day), does it? This was, what we wanted to resolve with another cache, right?

Right, see #1086 for details -- this PR addresses only the first subtask, which is to make the usage of other cache even possible. That other cache usage will come as a separate PR.

fredo:
> > But this does not solve the original problem that we still get ~12 requests per block on L1 (84000 times per day), does it? This was, what we wanted to resolve with another cache, right?
> 
> Right, see #1086 for details -- this PR addresses only the first subtask, which is to make the usage of other cache even possible. That other cache usage will come as a separate PR.

right, understand. Then looks good!

fredo (approved):

PR #1119, opened 2022-10-10
frontend: fix new network native currency display

### LINKED ISSUE #1109: Frontend: Add new network - invalid currency "UNKNOWN"
When a user selects a network on our dApp which is not yet found inside his MetaMask wallet he is prompted to add the network inside his MetaMask app like so:

image

On the image provided, its clear that the "Currency symbol" field is not properly set (defaults to "UNKNOWN"). 
We need to set it to "ETH" for networks that use "ETH" as a base currency.

Implementation reference: https://docs.metamask.io/guide/rpc-api.html#wallet-addethereumchain 
# Acceptance criteria
- [ ] The currency is always displayed properly when prompted to add a new network

GabrielBuragev:
Since we are positioned as an ETH rollup to rollup bridge for the beginning i implemented it using static code. 
Could be improved by defining `nativeCurrency` for each chain in the config file but that requires a decent amount of work and also not needed atm.
 
Fixes #1109

### DIFF
--- a/frontend/src/services/web3-provider/ethereum-provider.ts
+++ b/frontend/src/services/web3-provider/ethereum-provider.ts
@@ -70,6 +70,11 @@ export abstract class EthereumProvider extends EventEmitter implements IEthereum
         chainId: chainIdHexValue,
         chainName: name,
         rpcUrls: [rpcUrl],
+        nativeCurrency: {
+          name: 'Ethereum',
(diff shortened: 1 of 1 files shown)


GabrielBuragev:
Due to time constraints a quick solution was required.
Further improvement of this can be tracked here: https://github.com/beamer-bridge/beamer/issues/1121

PR #1130, opened 2022-10-12
docs: Fix a typo

compojoom:


docs/source/faq.rst
```diff
@@ -55,7 +55,7 @@ FAQ
 
 * What is a rollup?
 
-  A rollup performs transaction executioon oustide Layer 1 nad then the data is
+  A rollup performs transaction executioon oustide Layer 1 and then the data is
```
istankovic, line 58: My goodness... there are additional typos in here...

docs/source/faq.rst
```diff
@@ -55,7 +55,7 @@ FAQ
...
   posted to layer 1 where consensus is reached. There are two types of rollups,
   Zero-Knowledge (ZK) and Optimistic Rollups.
```
istankovic, line 60: It would be nice to also fix consistency when referencing "layer 1", while we're here.

docs/source/faq.rst
compojoom: wow, and I missed those...

docs/source/faq.rst
compojoom: What do you mean with fix consistency when referencing "layer 1"? That we go for either "Layer 1" or "L1"???

istankovic: We have "Layer 1" and then immediately below "layer 1" -- note the case change.

istankovic (approved):
Thanks!

PR #1136, opened 2022-10-17
subgraph: initial subgraph for l2 contracts

### LINKED ISSUE #889: Build a Subgraph for Beamer Contracts
Build a Subgraph and upload it to the hosting service of the graph

We have an already existing implementation in the company which is very close to what we need in the beginning. 

### Scope
The subgraph will become a powerful tool which lots of projects in the Beamer ecosystem will be able to use. For now which will keep the scope as tiny as we need. 
We will need the subgraph for #890 which is primarily there to monitor and alert upon any critical behavior regarding claims

- false claims
- during challenges the false outcome is winning

### Events to capture
This leads to two subgraph implementations which will only handle ClaimMade events and RequestFilled events respectively

### Setting up a Subgraph
- subgraph.yaml
- handlers
- schemas

We can almost clone our internal repo and replace it with the things we need.

fredo:
Closes #889 

This PR introduces the initial subgraph for the Beamer contracts on L2


subgraph/l2/schema.graphql
```diff
@@ -0,0 +1,84 @@
...
+  id: ID!
+  lp: Bytes! # address
+}
+
+type LPRemoved @entity {
+  id: ID!
+  lp: Bytes! # address
+}
+
+type FillManagerOwnershipTransferred @entity {
```
GabrielBuragev, line 24: Same here as above, is this the correct event name or should it be `OwnershipTransferred`?

subgraph/l2/networks.json
```diff
@@ -0,0 +1,10 @@
+{
+  "optimism": {
+    "RequestManager": {
+      "address": "0xe236d136aD9f6dB125e9ba95eDc1d1A19423Ff0F"
+    },
+    "FillManager": {
+      "address": "0x85267688CE7677D69d75F288Ff3064431030aF9F"
+    }
```
GabrielBuragev, line 8: Would it make sense to have these addresses as an environment variable inside `.env` or something similar? I think it would spare us updates in the repo. Instead we would need to only update the .env file on the hosting side. Do you think that is possible?

GabrielBuragev (changes requested):
Everything looks good considering its the initial implementation of a subgraph. 

I have a question regarding the naming of one of the events and would like to know if we can possibly improve the configuration process by adding variable contract addresses managed inside an `.env` file.

subgraph/l2/subgraph.yaml
```diff
@@ -0,0 +1,68 @@
...
+        - name: FillManager
+          file: ./abis/FillManager.json
+      eventHandlers:
+        - event: HashInvalidated(indexed bytes32,indexed bytes32,indexed bytes32)
+          handler: handleHashInvalidated
+        - event: LPAdded(address)
+          handler: handleLPAdded
+        - event: LPRemoved(address)
+          handler: handleLPRemoved
+        - event: FillManagerOwnershipTransferred(indexed address,indexed address)
```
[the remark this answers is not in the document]
fredo: Subgraph does this automatically if you have the same event in multiple contracts. It modifies the abi file so that the events are distinguishable from each other and are stored in different entitites.
I would say in the future we can clean up and don't store this event at all because it's not really relevant to us nor the users.

subgraph/l2/networks.json
fredo: Yes, the addresses will be stored in the network file, I am currently editing these.

fredo:
You can actually run different datasources on different networks per instance of a graphnode. So I will move the content of the folder l2 one level up.

subgraph/l2/subgraph.yaml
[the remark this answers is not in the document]
GabrielBuragev: Aah, now it makes sense. Thanks for clarifying!

subgraph/l2/networks.json
GabrielBuragev: What do you mean? Will you edit them to use environment variables or should we leave them like this for the time being?

subgraph/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      - '8040:8040'
+    depends_on:
+      - ipfs
+      - postgres
+    extra_hosts:
+      - host.docker.internal:host-gateway
+    environment:
+      postgres_host: postgres
+      postgres_user: graph-node
+      postgres_pass: let-me-in
```
GabrielBuragev, line 19: Can we use an environment variable for the pg password instead?

subgraph/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+    ports:
+      - '5432:5432'
+    command:
+      [
+        "postgres",
+        "-cshared_preload_libraries=pg_stat_statements"
+      ]
+    environment:
+      POSTGRES_USER: graph-node
+      POSTGRES_PASSWORD: let-me-in
```
GabrielBuragev, line 41: Same as above

GabrielBuragev (approved):

Issue #1112, opened 2022-10-06
Prohibit invalidating FillIds of current block


fredo:
The protocol makes the assumption that future fill ids can only be guessed. This is achieved by the fill id being the blockhash of the last block (assuming an honest sequencer).

However, the fillId of the current block is known during execution. It is theoretically possible that both, invalidation and fill are in the same block. Since future fill ids must be guessed and all fill ids from past blocks cannot be invalidated anymore if they are valid (we check against fills), we only run into this problem if the invalidation is in the same block but executed before the fill. 

We can easily prevent that and revert if someone wants to invalidate the fill id  of the current block.

istankovic:
> We can easily prevent that and revert if someone wants to invalidate the fill id of the current block.

Ah, this is actually a very nice solution! Very nice and simple. :+1:

istankovic:
Do you know how we can test this? Because we would definitely like to have a test for this.

fredo:
Testing should be fairly simple. We just need to access the block hash of the latest block and use it as a fill id for an invalidation

PR #1138, opened 2022-10-17
Prohibit invalidation of previous block hash

fredo:
fixes: #1112

### DIFF
--- a/beamer/tests/contracts/test_fill_manager.py
+++ b/beamer/tests/contracts/test_fill_manager.py
@@ -139,3 +139,15 @@ def test_unset_resolver(deployer, token):
     )
 
     new_fill_manager.invalidateFill(request_id, 1, chain_id)
+
+
+def test_invalidate_previous_block_hash(fill_manager):
+    chain_id = brownie.web3.eth.chain_id
+    current_block = brownie.web3.eth.get_block("latest")
+    with brownie.reverts("Cannot invalidate fills of current block"):
+        fill_manager.invalidateFill(1, current_block["hash"], chain_id)
+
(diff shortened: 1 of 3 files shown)


contracts/contracts/FillManager.sol
```diff
@@ -142,6 +142,11 @@ contract FillManager is Ownable, LpWhitelist {
     ) external {
         require(l1Resolver != address(0), "Resolver address not set");
         require(fills[requestId] != fillId, "Fill valid");
+        require(
+            fillId != generateFillId(),
+            "Cannot invalidate block hash of previous block"
```
istankovic, line 147: Nitpick: since we talk about invalidating fills, not block hashes, in order to be more consistent I'd reword the message to be more like
```
         "Cannot invalidate fill within same block"
```

beamer/tests/contracts/test_fill_manager.py
```diff
@@ -139,3 +139,13 @@ def test_unset_resolver(deployer, token):
...
     new_fill_manager.invalidateFill(request_id, 1, chain_id)
+
+
+def test_invalidate_previous_block_hash(fill_manager):
+    chain_id = brownie.web3.eth.chain_id
+    current_block = brownie.web3.eth.get_block("latest")
+    with brownie.reverts("Cannot invalidate block hash of previous block"):
+        fill_manager.invalidateFill(1, current_block["hash"], chain_id)
+
+    # We proceeded one block so invalidating the fill id should work now
```
istankovic, line 150: Nitpick: can we assert that here, e.g.
```python
    assert current_block["hash"] != brownie.web3.eth.get_block("latest")["hash"]
```
?

istankovic (approved):
Looks good, with a couple of nitpicks.

PR #1140, opened 2022-10-18
beamer: agent: use web3.middlewares.latest_block_based_cache_middlewa…

istankovic:
…re for L1

This should help reduce the number of calls to L1 RPC nodes, and in particular reduce the number of eth_gasPrice calls (see https://github.com/beamer-bridge/beamer/issues/1086).

### DIFF
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -4,6 +4,7 @@ from concurrent.futures import ThreadPoolExecutor
 import structlog
 from eth_typing import Address
 from web3 import Web3
+from web3.middleware import latest_block_based_cache_middleware
 
 import beamer.metrics
 from beamer.chain import POLL_PERIOD_MAINNET, POLL_PERIOD_TESTNET, EventMonitor, EventProcessor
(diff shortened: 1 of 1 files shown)


fredo (approved):
looks good. 

One question which popped into my head just now. Since we are applying this cache only to web3_l1, couldn't we have applied the other cache which we used to use, only for the l2s? Or was the older cache fundamentally broken, so we had to write our own?

istankovic:
> One question which popped into my head just now. Since we are applying this cache only to web3_l1, couldn't we have applied the other cache which we used to use, only for the l2s? Or was the older cache fundamentally broken, so we had to write our own?

There are two caches at play here. The `cache_get_block_by_number` which is used by all the chains, but that one is actually doing the correct thing now (previously that part was broken). And there is the latest_block_based_cache middleware, which is only used by L1 `Web3` instance.

I considered using the latter middleware (actually, manually constructed variants of it) for the other chains, but decided against it in the end because we don't really have the info on various other chains' block times, which we could use to make the cache algorithm settle more quickly and I didn't think it was worth it. Maybe we can consider it in the future, but let's see.

Issue #1131, opened 2022-10-13
Add matomo tracking


andfletcher:
# Rationale
We want to use matomo to get more user behavior insights to improve the website and app.

# Implementation
 - [x] Add the following to our websites:

Make sure this code is on every page of your website. We recommend pasting it immediately before the closing </head> tag

```
<!-- Matomo -->
<script>
  var _paq = window._paq = window._paq || [];
  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
  _paq.push(['trackPageView']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="https://beamerbridge.matomo.cloud/";
    _paq.push(['setTrackerUrl', u+'matomo.php']);
    _paq.push(['setSiteId', '1']);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.async=true; g.src='//cdn.matomo.cloud/beamerbridge.matomo.cloud/matomo.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
<!-- End Matomo Code -->
```

- [x] Add a disclaimer to our privacy policy that we are using matomo

```
Web analysis tool Matomo
We analyze the use of our website with the open-source web analysis tool Matomo (formerly Piwik; data protection declaration can be found at https://matomo.org/privacy-policy/). For this purpose, the automatically collected IP addresses are anonymized before the evaluation. The analysis is therefore based on anonymized data records, and there is no personal evaluation. The web analysis is used exclusively to optimize the website in terms of user-friendliness and to provide useful information about our services. This data is not merged with other personal data sources or passed on to third parties.
```

In addition to the data mentioned, Matomo also uses so-called “cookies” (see below).

- [x] Allow opt out from the tracking via privacy policy by adding matomo opt in / out option bellow added privacy text
```
<div id="matomo-opt-out"></div>
<script src="https://beamerbridge.matomo.cloud/index.php?module=CoreAdminHome&action=optOutJS&divId=matomo-opt-out&language=auto&showIntro=1"></script>
```

DBecane:
@GabrielBuragev can you give me more info about if and for when this issue was prioritized? Just want to get an estimation about when this can be done. I assume this was taken to Jacob already as he was aware of it when we had our marcomms sprint planning.

GabrielBuragev:
@DBecane i am already working on it in parallel to other critical issues. Should be live by end of tommorow.

GabrielBuragev:
[...]

PR #1142, opened 2022-10-18
frontend: include matomo tracking script

GabrielBuragev:
Solves a small part of the issue: #1131 

Seems like [Piwik is not tracking](https://stackoverflow.com/questions/13688749/piwik-localhost-tracking) `localhost` activity by default therefore there was no need to avoid script loading inside dev environment.


compojoom (approved):
LGTM

PR #1148, opened 2022-10-19
pyproject.toml: raise minimum Python version to 3.10

istankovic:
And also update all dependencies.


fredo:
you also need to update the python image in .circle I/config.yaml

fredo (approved):

PR #1150, opened 2022-10-21
frontend: fix "use max token balance" feature

### LINKED ISSUE #1127: Frontend: "Use max" token amount feature should include fees in final amount
## Description of Deliverables(s)

Community member: 
When I have 326usdc in my wallet, choose max, it will prompt that the amount is not enough.
It took a long time to understand this setting because a portion of the amount is left to pay for processing fees
Personally, I think this amount can be set as dynamic, if the remaining amount is not enough to pay the commission, it will be deducted from the selected amount; if the remaining amount is enough to pay the commission, it will be deducted from the remaining amount.
Or when the user clicks max, the amount entered in the input box is automatically subtracted by 1. 

## To Do 

I guess we can make the 'max' input dynamic so it takes the max possible amount deducted from the fees? 

@czepluch I think your input here is needed to make a decision. What's the best way to approach thin in your opinion? Given that we want to give people the choice to deduct or top up the fee amount.

GabrielBuragev:
Fixes #1127 

I adjusted the `UInt256` abstraction to not support creation of negative integers since that was contradictory to what this class is supposed/expected to do (unsigned should stand for positive only). It was necessary to adjust it as so since i introduced arithmetic operations over it.

Also includes a solution to [Manuel's suggestion](https://github.com/beamer-bridge/beamer/issues/1127#issuecomment-1278048533)

### DIFF
--- a/frontend/src/services/transactions/request-manager.ts
+++ b/frontend/src/services/transactions/request-manager.ts
@@ -9,10 +9,10 @@ import type { Cancelable } from '@/types/async';
 import type { EthereumAddress } from '@/types/data';
 import { UInt256 } from '@/types/uint-256';
 
-export async function deriveBaseTransferableAmountFromTotalAmount(
+export async function getAmountBeforeFees(
+  totalAmountWei: UInt256,
   rpcUrl: string,
   requestManagerAddress: string,
-  totalAmountWei: UInt256,
 ): Promise<UInt256> {
   const PARTS_IN_MILLION = new UInt256('1000000');
   const contract = getContract(rpcUrl, requestManagerAddress, RequestManager.abi);
--- a/frontend/src/services/transactions/request-manager.ts
+++ b/frontend/src/services/transactions/request-manager.ts
@@ -4,20 +4,47 @@ import type { BigNumberish } from 'ethers';
 import { BigNumber, Contract } from 'ethers';
 
 import RequestManager from '@/assets/RequestManager.json';
+import { getContract } from '@/services/transactions/utils';
 import type { Cancelable } from '@/types/async';
 import type { EthereumAddress } from '@/types/data';
 import { UInt256 } from '@/types/uint-256';
 
-function getContract(rpcUrl: string, address: EthereumAddress): Contract {
-  const provider = new JsonRpcProvider(rpcUrl);
-  return new Contract(address, RequestManager.abi, provider);
+export async function deriveBaseTransferableAmountFromTotalAmount(
+  rpcUrl: string,
+  requestManagerAddress: string,
+  totalAmountWei: UInt256,
+): Promise<UInt256> {
+  const PARTS_IN_MILLION = new UInt256('1000000');
+  const contract = getContract(rpcUrl, requestManagerAddress, RequestManager.abi);
+  const minLpFeeWei: UInt256 = new UInt256(await contract.minLpFee());
+  const lpFeePartsPerMillion: UInt256 = new UInt256(await contract.lpFeePPM());
+  const protocolFeePartsPerMillion: UInt256 = new UInt256(await contract.protocolFeePPM());
+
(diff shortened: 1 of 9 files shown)


frontend/tests/unit/services/transactions/request-manager.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+          rpcUrl,
+          requestManagerAddress,
+          totalAmountWei,
+        );
+
+        const decimalResult =
+          (totalAmountDecimal * PARTS_IN_PERCENT) /
+          (PARTS_IN_PERCENT + lpFeePercent + protocolFeePercent);
+
+        expect(baseAmount.format(DECIMALS)).toBeCloseTo(decimalResult, DECIMALS - 1);
```
compojoom, line 69: I would actually convert the decimal amount to wei using ethersjs and then make sure that the expected wei amounts match

frontend/tests/unit/services/transactions/request-manager.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+          rpcUrl,
+          requestManagerAddress,
+          totalAmountWei,
+        );
+
+        const decimalResult =
+          (totalAmountDecimal * PARTS_IN_PERCENT) / (PARTS_IN_PERCENT + protocolFeePercent) -
+          minLpFee;
+
+        expect(baseAmount.format(DECIMALS)).toBeCloseTo(decimalResult, DECIMALS - 1);
```
compojoom, line 128: same as above.

frontend/src/services/transactions/request-manager.ts
```diff
@@ -4,20 +4,47 @@ import type { BigNumberish } from 'ethers';
...
 import RequestManager from '@/assets/RequestManager.json';
+import { getContract } from '@/services/transactions/utils';
 import type { Cancelable } from '@/types/async';
 import type { EthereumAddress } from '@/types/data';
 import { UInt256 } from '@/types/uint-256';
 
-function getContract(rpcUrl: string, address: EthereumAddress): Contract {
-  const provider = new JsonRpcProvider(rpcUrl);
-  return new Contract(address, RequestManager.abi, provider);
+export async function deriveBaseTransferableAmountFromTotalAmount(
```
compojoom, line 12: maybe: getAmountBeforeFees(totalAmountWei...)

compojoom (changes requested):
Just the few changes as we discussed in the call.

frontend/src/services/transactions/request-manager.ts
GabrielBuragev: :heavy_check_mark:

GabrielBuragev:
> I would actually convert the decimal amount to wei using ethersjs and then make sure that the expected wei amounts match

So, for this part, i solved it by [writing my own function](https://github.com/beamer-bridge/beamer/pull/1150/commits/7b93024202fa3542a1248854e8746a95a213c899#diff-80b936d3f9979a9815e5987d0f4e5e45a775ffffdedbbbc1a03f09c99f321dc9R31-R39) that transforms the decimal number to wei since using ethersjs in the tests leads to:
- Undesirable extra dependency 
- `"fractional component exceeds decimals"` error when trying to parse a JS floating number with `N` decimals to a Wei representation with `N-M` (M > 0) decimals. IOW, we have to pass this function a JS floating number with exact same decimals that we are trying to represent the Wei amount to. That will almost never be possible since JS cannot be restricted to using `N` number of decimals when doing division. 
We could use JS `floatNumber.toFixed(decimalCount)` but that leads to another undesired behavior which is rounding the decimals.

Considering all of the above, i decided to write a simple but limited function to do the job for our specific test cases.

compojoom (approved):
LGTM

GabrielBuragev:
Update on the final implementation:
We decided to simplify the tests by checking the response from the function against static values (expected results) instead of having the decimal calculation logic inside the tests.

PR #1167, opened 2022-11-01
frontend: increase request expiry period

GabrielBuragev:


compojoom (approved):
You can event have this as env variable for the website... But I guess this is fine as well.

GabrielBuragev:
Good suggestion :+1:

PR #1177, opened 2022-11-07
Frontend: Disallow entering non-numerical characters in token amount input

### LINKED ISSUE #1165: Frontend: Disallow entering non-numerical characters in token amount input
Rationale
---
To keep the UI/UX simple & less confusing for the user, as well as the code cleaner & more robust, we need to limit the allowed characters inside the token amount input. The token amount input should only accept numerals and a "." as a decimal separator.
It should work exactly like the token amount input's on PoWDump, Uniswap etc.
Having a free-style input showed us that many things can go wrong at the form validation & contract interaction level when the value for the token amount is invalid in the first place.

Implementation
---
After [my previous failed attempt](https://github.com/beamer-bridge/beamer/pull/918#issuecomment-1225298052) to solve this, i was challenged again to find a solution to this issue by using [this implementation](https://github.com/brainbot-com/PoWDump/blob/main/packages/website/components/input-row/custom-decimal-input.tsx?rgh-link-date=2022-10-27T09%3A01%3A56Z) as reference.
By playing around with this code in Vue i learned this lessons:
- Vue triggers the element's `input` event too late (only after the elements inner value has changed) therefore preventing (blocking) the input is impossible for these events.
- The `change` event is triggered when the input loses focus apparently, so its quite different than the onChange event in React, therefore it is useless for our case.
- The only events you can play with are the keydown, keypress & keyup events however these are emitting a KeyboardEvent which only contains info about the keyCode that was pressed. This explains the background of my attempt in the past to fix it by detecting the keyCode which rendered useless in the end because of the cross-browser/device incompatibility.

However, by extracting the regexp from the reference code & playing around with the input event together with the `HTMLInputElement` API i was able to come to a solution that works.

The solution is to use `HTMLInputElement.value` to overwrite the value (with the previous input value) once it goes from a valid state to an invalid state (validated by the regexp).

GabrielBuragev:
closes #1165

Slight refactoring to the `Input` component led to splitting the "focus on mount" behavior to its own reusable directive.


frontend/tests/unit/components/inputs/NumericInput.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+function createWrapper(options?: {
+  modelValue?: string;
+  focusOnMount?: boolean;
+  attachToBody?: boolean;
+  valid?: boolean;
+  alignRight?: boolean;
+}) {
+  return mount(NumericInput, {
+    shallow: false,
```
manuelwedler, line 15: I would always prefer a shallow mount for unit tests. This way it can be ensured that we only test the one component that should be tested and we avoid possible side effects by a child component.

frontend/tests/unit/components/inputs/BasicInput.spec.ts
```diff
@@ -1,20 +1,18 @@
 import { mount } from '@vue/test-utils';
 
-import Input from '@/components/inputs/Input.vue';
+import BasicInput from '@/components/inputs/BasicInput.vue';
```
manuelwedler, line 3: I'm also wondering if it would make sense to stub the vFocusOnMount directive here for the same reason as above. Not sure though as it would require additional work wherever we use the directive.

manuelwedler (changes requested):
Looks good, just some minor improvements could be taken

frontend/src/directives/vFocusOnMount.ts
```diff
@@ -0,0 +1,11 @@
+import type { ObjectDirective } from 'vue';
+
+export type Param = boolean;
+
+export const vFocusOnMount: ObjectDirective<HTMLElement, Param> = {
+  mounted: async (el: HTMLElement, binding) => {
```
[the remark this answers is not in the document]
GabrielBuragev: You are right! I overlooked it.

frontend/tests/unit/components/inputs/BasicInput.spec.ts
GabrielBuragev: In order to make this work, we will need to bump `@vue/test-utils` to `2.2.0` since stubbing directives (using `<script setup>`) is not supported in the versions below that.
https://github.com/vuejs/test-utils/releases/tag/v2.2.0

Working on it, lets see how much will break 😄

frontend/src/directives/vFocusOnMount.ts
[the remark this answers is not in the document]
GabrielBuragev: ✅

frontend/tests/unit/components/inputs/NumericInput.spec.ts
GabrielBuragev: ✅

frontend/tests/unit/components/inputs/BasicInput.spec.ts
GabrielBuragev: ✅

manuelwedler (approved):
Looks good now :)

GabrielBuragev:
Thanks for the review!

PR #1180, opened 2022-11-08
.github: go back to prettier-solidity-plugin 1.0.0-beta.19

istankovic:
The newer 1.0.0-rc.1 causes unnecessary formatting changes and this is intended to fix our CI until we decide how to handle those changes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
     steps:
       - attach_workspace:
           at: "~"
-      - run: npm install prettier prettier-plugin-solidity --location=global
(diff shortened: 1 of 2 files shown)


compojoom (approved):

GabrielBuragev:
Do we maybe need to update here as well:
- https://github.com/beamer-bridge/beamer/blob/main/.circleci/config.yml#L72

istankovic:
> Do we maybe need to update here as well:
> 
>     * https://github.com/beamer-bridge/beamer/blob/main/.circleci/config.yml#L72

Definitely, thanks for catching this!

PR #1184, opened 2022-11-09
frontend: Add explorer URL to wallet

### LINKED ISSUE #1124: Add Block Explorer when proposing the network for Metamask
#### Expected behavior
I would expect to be able to click on a transaction in metamask and land on the respective chain explorer

#### Actual behavior
If optimism is added through beamerbridge then transactions made don't have a link to the optimism blockexplorer:
image

When we propose the optimism network to Metamask, we can add a blockExplorer prop and specify https://optimistic.etherscan.io for block explorer. This way the TXs in Metamask will have a correct link.

manuelwedler:
Closes #1124 

When adding a chain to the connected wallet, the explorer URL is now added. This also changes our config files by renaming `explorerTransactionUrl` to `explorerUrl` and removing the `/tx` path from the urls. This can be safely done as explorer routes are standardized in EIP-3091. Furthermore, the `addChain` was changed in that way that it now consumes the usual `Chain` type instead of the `ChainData` type. The `ChainData` type was added when we did not have the split types of the configuration and is redundant now. Same goes for `Token` and `TokenData`.

### DIFF
--- a/frontend/config/chains/10.json
+++ b/frontend/config/chains/10.json
@@ -1,6 +1,6 @@
 {
   "identifier": 10,
-  "explorerTransactionUrl": "https://optimistic.etherscan.io/tx/",
+  "explorerUrl": "https://optimistic.etherscan.io",
   "rpcUrl": "https://opt-mainnet.g.alchemy.com/v2/umLO_ve2OzJ_7rAVebkVbUsuM37cCFoN",
   "name": "Optimism Mainnet",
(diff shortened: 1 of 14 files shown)


GabrielBuragev (approved):
Everything looks good. ✅ 
I like the removing of `ChainData` & `TokenData` from the `src` part of the codebase. They were causing a lot of confusion.

PR #1187, opened 2022-11-10
Finish move to Circle CI

istankovic:
This PR essentially rewrites the current CircleCI config so that
- `config.yml` detects changes that have been made and sets up several pipeline parameters accordingly
- `workflows.yml` uses pipeline parameters to determine which workflows to run
- caches are used aggressively: Python runtime, Python  packages, npm runtime, npm packages, Solidity
- docker layer caching is used to speed up docker image builds (most notably e2e tests)
- most jobs use a VM executor (`vm`) due to networking setup limitations which we need to avoid for our e2e tests
- using a VM adds overhead to each job, but that could be mitigated by merging all jobs of a single workflow, which would also avoid needless workspace operations (not done in this PR, something to consider in the future)

The first part of this PR contains changes to update all components to node version 18, in order to simplify life.


istankovic:
> [This line](https://github.com/beamer-bridge/beamer/blob/301b52908c78f1ca9bbb89395cc005d25b2e92b3/beamer/l1_resolution.py#L13) also needs to be updated

:eyes: 
Thanks!

.circleci/workflows.yml
```diff
@@ -110,23 +110,28 @@ jobs:
...
-            for ver in $(ls /opt/circleci/.nvm/versions/node/); do nvm uninstall ${ver}; done
             nvm install --default v18
             nvm version default > ~/.nvmrc
       - run: npm install prettier prettier-plugin-solidity@1.0.0-beta.19 --location=global
       - run: npm install solhint --location=global
       - run: npm install ganache --location=global
+      - run: echo 'nvm use v18' >> ~/.bashrc
       - save_cache:
-          key: node18-packages-v3
+          key: node18-packages-v4
```
fredo, line 130: do we store the executables of the above installed packages here?

istankovic: > do we store the executables of the above installed packages here?

Yes.

fredo (approved):
very nice! 🍷

PR #1188, opened 2022-11-11
frontend: add TypeChain library & yarn hooks

### LINKED ISSUE #549: frontend: add TypeChain to generate contract types
Using [TypeChain](https://github.com/dethcrypto/TypeChain) brings a huge typing advantage. We just need to add a hook script that generates the `*.d.ts` files from the already existing `.json` files with the compiled ABI. Afterwards we have proper typing of the contracts with all functions, events and their parameters.

GabrielBuragev:
Closes #549


manuelwedler (approved):
Looks good. Very happy to have this now!

PR #1220, opened 2022-11-23
contracts: allow claim until validUntil + claimRequestExtension

fredo:
Allows claiming after request expiry. Time is bounded by claimRequestExtension.

I went for the strict approach to introduce a new variable `claimRequestExtension`. Same as we did with claimPeriod and challengePeriodExtension which follow the same principle: Give an actor enough time to react.
In the future we can think about merging these values together, but for now I wanted to keep the pattern and create an explicit value.

### DIFF
--- a/beamer/tests/agent/unit/test_handlers.py
+++ b/beamer/tests/agent/unit/test_handlers.py
@@ -27,6 +27,29 @@ from beamer.tests.constants import FILL_ID
 from beamer.typing import FillId, Termination
 
 
+@pytest.mark.parametrize("claimable", [True, False])
+def test_claim_after_expiry(claim_request_extension, claimable):
+    context, config = make_context()
+    valid_until = TIMESTAMP - claim_request_extension
+    assert valid_until > 0
+    
+    if claimable:
+        valid_until += 1
+
+    request = make_request(valid_until)
+
+    context.requests.add(request.id, request)
+    request.filler = config.account.address
+    request.try_to_fill()
+
+    assert request.is_filled  # pylint:disable=no-member
+    claim_request(request, context)
+    if claimable:
+        assert request.is_claimed  # pylint:disable=no-member
+    else:
+        assert request.is_ignored  # pylint:disable=no-member
+
+
 def test_skip_not_self_filled():
     context, _ = make_context()
     request = make_request()
@@ -38,10 +61,11 @@ def test_skip_not_self_filled():
     assert request.is_pending  # pylint:disable=no-member
 
 
-def test_ignore_expired():
+def test_ignore_expired(claim_request_extension):
     context, config = make_context()
-
-    request = make_request()
+    valid_until = TIMESTAMP - claim_request_extension
+    assert valid_until > 0
+    request = make_request(valid_until=valid_until)
     request.filler = config.account.address
     context.requests.add(request.id, request)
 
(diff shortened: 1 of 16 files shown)


beamer/tests/agent/unit/test_handlers.py
```diff
@@ -27,6 +27,27 @@
 from beamer.typing import FillId, Termination
 
 
+@pytest.mark.parametrize("claimable", [True, False])
+def test_claim_after_expiry(claim_request_extension, claimable):
+    context, config = make_context()
+    valid_until = TIMESTAMP - claim_request_extension
```
istankovic, line 33: Nitpick: we should add
```
    assert valid_until > 0
```
here, similar to what we have in `test_ignore_expired`.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -296,6 +296,31 @@ def test_claim_nonexistent_request(request_manager):
...
+    token.mint(requester, 1, {"from": requester})
+    request_id = make_request(request_manager, token, requester, requester, 1)
+
+    valid_until = request_manager.requests(request_id)[RM_R_FIELD_VALID_UNTIL]
+    claim_request_extension = request_manager.claimRequestExtension()
+    # test that request expiration does not prevent claiming
+    timestamps = [valid_until - 1, valid_until, valid_until + claim_request_extension - 1]
+
+    for timestamp in timestamps:
+        chain.sleep(timestamp - chain.time())
```
istankovic, line 315: I like the fact we are testing all the "interesting" timestamps.
Is there a way to ensure that `sleep` did the trick, e.g. something like
```
assert chain.time() == timestamp
```
so that we also ensure this does not silently break in the future?

docs/source/protocol.rst
```diff
@@ -62,6 +63,23 @@ Proposed Values
 We propose a value of ``k = 1.3`` so that honest challengers are rewarded with a minimum of 30% of the challenge gas
 costs. The other parameters should be determined at time of deployment.
 
+claimRequestExtension
+-----------
```
istankovic, line 67: Incomplete section marker (too short).

docs/source/protocol.rst
```diff
@@ -135,6 +153,9 @@ expiration times. While setting a very low expiration time most likely leads to
 by any LP, an upper boundary ensures that funds can eventually be withdrawn. With the current setup
 of fixed fees and a race between LPs, we introduce a safety net for LPs to ensure that there is
 enough time to register a claim of a filled request *before* it expires.
+An LP is able to claim a request after its expiry date. The period in which an LP can do that is
+defined by ``claimRequestExtension``. Note that during this time, the user can also withdraw the funds if there is
+no claim active.
```
istankovic, line 158: A very minor nitpick: I'd prefer "if there are no active claims".

istankovic (approved):
Looks good, with a few comments.

beamer/tests/contracts/test_request_manager.py
fredo: Unfortunately I don't think so and with new versions of either brownie or ganache, this can break again.
The reason for this whole mess was a major change in ganache v6 to v7 where the next transaction mined would increment the timestamp. In v7 this is not the case anymore and you need to explicitly increment the timestamp for a new block.

Brownie then, decided to increment the timestamp internally before setting the time in ganache upon `chain.mine()`. They did this to be consistent with v6. So now, if one calls `chain.mine(timestamp=x)`, brownie internally increments the timestamp before calling ganache which is super counter intuitive to me. 

At least they mention in there docs that for timestamp testing one should use chain.sleep. I personally like the fact that I can control the timestamp for the next block in ganache, especially for this kind of testing.

Issue #1170, opened 2022-11-02
Last Contract optimizations by audit


fredo:
```
RequestManager, line 41. Optimization, the Request.withdrawClaimId field could fit into a smaller type.
RequestManager, line 187. Note, consider setting transferLimit on a per token basis.
RequestManager, line 462. Optimization, the challengeClaim() could pointlessly update claim.lastChallenger with it's own value.
```

* WithdrawClaimId could be put into `uint96` and would then fit into one storage slot with filler (address type takes 20 bytes -> 160 bits).
* TransferLimit on a per token basis will be covered in a seperate issue (together with fees per token)
* In order to fix the 3rd issue  we need to fit lastChallenger and challengerStakeTotal into one slot. In this way reading the lastChallenger does not cost an extra sRead bytecode. In the same way we could also store claimer and claimerStake into one slot. This would result in uint96 for stakes. We should think about if we want to limit the maximum stake to avoid overflows. Realistically uint96 is more than enough considered the 120 million ETH in existence. (Currently uint88 is already enough)

Issue #941, opened 2022-08-18
Decimals of fee value


fredo:
We discovered an issue with the fixed fee value (reimbursement value). We currently have one value for all tokens. The problem is that different tokens might have different size of decimals (i. e. USDC only having 6, DAI having 18). 

Starting with only one token it's not a problem. As soon as we have multiple tokens we might run into a problem. 
Possible solutions require a redeployment which should be done soon but after launch. 
This is also a good test to how redeployments work since it is a critical process in the product liveness.

Solutions: 

Option 1: Have a constant decimal fee value
We could define a constant decimal fee value. Upon request we pad the value with the number of zeros required for this token dynamically. The number of decimals can be fetched from the contract.
- This leads to still allowing all tokens to be transferred
- Only works for tokens with equal USD value. (5 USD != 5 ETH)

Option 2: Have a mapping per token
We can have a fee per token with the correct value. This leads to the following situation. Either we will have a whitelist of tokens or we will have a default fee value (i.e. "0") for tokens which are not in the list.
Here we could adjust for different tokens which are not stables. 

## Outcome
Agreed on the following:

We are gonna have a mapping on a per token basis:

values: protocolFee, lpFee, minLpFee, transferLimit
function: addToken, removeToken (onlyOwner)
events: TokenAdded, TokenRemoved, TokenChanged

Open question/partially open:
I would also like to add a tokenChanged event. This should reflect whenever fees were changed. I had a discussion the other day with an agent how to reflect the earned fees, and we actually never know when fees were changed. By having this event, we always know the current fees when a request was created by processing the tokenChanged event first.
(Could also be done by emitting another TokenAdded event, I don't have strong opinions on that).

fredo:
This should also be implemented before the audit

istankovic:
Before we decide on an option here, I think it would be useful to know about our strategy for supporting tokens.
@czepluch will there be a whitelist of tokens?

czepluch:
It's a good question. Maintaining a whitelist definitely makes it easier for us to make sure that our frontend reflects the supported tokens and we would also have some level of quality assurance. It would however introduce additional centralisation too. And I would say that ultimately we should not choose the tokens, but rather tokens should be added from the agents where they see that there is a demand. 

I know this doesn't really answer your questions yet, but I think it starts a discussion needed to come to the best solution.

istankovic:
> Option 2: Have a mapping per token
[...]

PR #1236, opened 2022-11-30
Token Mapping

fredo:
implements: #941 and part of #1170 

Some findings I made along the way which influenced my design decisions:

* We already had a mapping per token address, namely `collectedProtocolFees`. I decided to have an overarching token mapping which includes collected protocol fees
* There was an open question how we define whether a token is "whitelisted" or not. Turns out, this check is already in place by checking the `transferLimit`. With the transferLimit we can control the tokens which are allowed to be transferred via the protocol. With this in place we could even allow transfers free of any charge.
* I implemented the rest of the suggested contract optimiziations as half of it is already part of this issue and it was more or less nitpicks
* Design choice for types:
collectedProtocolFees & minLpFee: uint96 -> 7.9e28 -> results in a max decimal value of 79 billion (assuming the token has 18 decimals). Additionally, in the requests object lpFee also has uint96, and we can assume that the owner of the fees will withdraw before 79 billion (in decimals) is reached.
protocolFeePPM, lpFeePPM: Assuming 999999 is the highest value one would need uint20, rounded up to uint32.
The fee values often are read and written together, so it makes sense to store them in one uint256 slot if possible.
I left `transferLimit` uint256 as it limits amount. So any choice here would have to be reflected in the choice of the type of amount. I'll leave that for future optimizations.
* **Note: Contract Interface Changed!** the functions `lpFee`, `protocolFee` and `totalFee` additionally have the tokenAddress as a parameter. Probably Frontend related

### Token Events
During the implemenation of adding the events as discussed in #941, I realized that we were about to overcomplicate things in the contract. I took a different approach by only having one event. Here are my thoughts on this.

First I thought about the concrete requirements we have for this event to express. I talked to @GabrielBuragev what the frontend needs. And it boiled down to the fact that the frontend needs to know which tokens are transferrable and which aren't. Basically which token is whitelisted.
If we don't want to go the route and preconfigure, we can achieve this by fetching the events. What we need to get from the events is, whether a token is whitelisted. Effectively the whitelist is expressed in transferLimit (transferLimit = 0 not whitelisted, else whitelisted). That's handy.
Secondly, we need to fetch the last event for every token, which changed the token being (de-)whitelisted. So explicit events for adding, changing and removing is only more human readable but a single event expresses the same amount of information (Note, the interesting part is the transfer limit which serves as a whitelist).
The drawback in my opinion for three distinct events lies in the complexity of the implementation in the contracts. We would need to make sure that we always have an order of events (1x tokenAdded -> N times token changed -> 1x Removed -> 1x tokenAdded ...). This would have needed to be handled in either separate functions or with a bunch of conditions. I don't see any use in having these becuase if a token was added, changed or removed can also be constructed from a generalized event. 

Therefore I created 
```
    event TokenUpdated(
        address tokenAddress,
        uint256 transferLimit,
        uint96 minLpFee,
        uint32 lpFeePPM,
        uint32 protocolFeePPM
    );
```
which gives us all the information we need (transferLimit as the whitelist and the current fees).

### Deployment Token information 
We are adding a new constraint to our deployment infrastructure. The config needs to have token information within the config if we want to immediately add tokens to our deployments. However, with the option `--deploy-mintable-token` we get the address of the token only upon deployment, where the config is already consumed.

I realized that our configs for deploying beamer have two different schemas.

* A real deployment config file: It is possible to add multiple L2s (even preferred)
* A test deployment config file: They are only capabale of having one L2 in their config. It is not possible to have multiple L2s.

This is due to the fact that our deployment scripts are rollup specific. That's not a problem at all, but I see that in the future with cross rollup e2e tests we have to touch this part anyway. Additionally we will rework the deployment infrastructure anyways.

Because of this reason I chose a simple digestable approach. For real deployments (mainnet, testnet) which are typically one time events, we dont use the `--deploy-mintable-token` option. We use existing tokens.
So this problem only cooncerns our test setups.
For those cases (local setups, e2e tests) I create a placeholder in the token dict for the value of the key `token_address` indicating that these data are to be configured in the deployed mintable token contract. This means whenever we use the option `--deploy-mintable-token` the config should have a token configured with `token_address: "deploy_mintable_token"`.  I think this is fine as we have a pretty strict and never changing set of such configurations `optimism-local`, `arbitrum-local`, etc. .

Again, I think we will rework the deployment infrastructure and the e2e test setup in the near future so this change will be reworked as well. I can imagine that it would be a good idea to remove the `deploy-mintable-token` option completely and explicitely deploy a token when it is needed. In this way a future deployment script would be equal for test and real deployments.


contracts/contracts/RequestManager.sol
```diff
@@ -120,6 +120,17 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
 
     event FinalityPeriodUpdated(uint256 targetChainId, uint256 finalityPeriod);
 
+    /// Emitted when token object of a token address is updated.@author
```
istankovic: The `@author` seems to be misplaced here.

beamer/tests/util.py
```diff
@@ -163,20 +164,33 @@ def calculate_gas_spending():
...
-        request_manager.minLpFee(),
-    )
+# The function updates the token values partially whatever variable is in the params dictionary
+# This function works similar to adding a parameters dict to a transaction in web3.py
+def update_token(request_manager, token, params, *args, **kwargs):
+    token_data = list(request_manager.tokens(token.address))
+    for i, key in enumerate(["transfer_limit", "min_lp_fee", "lp_fee_ppm", "protocol_fee_ppm"]):
+        token_data[i] = params.get(key, token_data[i])
+    # the last element of token data is collectedProtocolFees which is not part of the update
+    request_manager.updateToken(token.address, *token_data[:-1], *args, **kwargs)
```
istankovic, line 174: Nitpick: I think we could avoid the `[:-1]` kludge by doing e.g.

```python
fields = "transfer_limit", "min_lp_fee", "lp_fee_ppm", "protocol_fee_ppm"
token_data = tuple(params.get(key, token_data[i]) for i, key in enumerate(fields))
request_manager.updateToken(token.address, *token_data, *args, **kwargs)
```

beamer/tests/contracts/test_request_manager_fees.py
```diff
@@ -244,3 +244,18 @@ def test_different_fees(request_manager, token, claim_period, claim_stake):
...
     assert token.balanceOf(claimer) == amount * 2 + lp_fee_1 + lp_fee_2
+
+
+def test_ppm_out_of_bound(request_manager, token):
+    ppm = 999999
+
+    request_manager.updateToken(token.address, 0, 0, ppm, 0)
+    request_manager.updateToken(token.address, 0, 0, 0, ppm)
+
+    ppm = 1000000
```
istankovic, line 255: Nitpick: I would use `1_000...` syntax here.

beamer/tests/contracts/test_request_manager_fees.py
```diff
@@ -244,3 +244,18 @@ def test_different_fees(request_manager, token, claim_period, claim_stake):
     )
     assert token.balanceOf(request_manager) == protocol_fee_1 + protocol_fee_2
     assert token.balanceOf(claimer) == amount * 2 + lp_fee_1 + lp_fee_2
+
+
+def test_ppm_out_of_bound(request_manager, token):
+    ppm = 999999
```
istankovic, line 250: Nitpick: I would use `999_999` syntax here.

scripts/deployment/arbitrum-local-template.json
```diff
@@ -17,7 +17,17 @@
                 "claim_request_extension": 3600,
                 "claim_period": 3600,
                 "challenge_period_extension": 3600
-            }
+            },
+            "token":
```
istankovic, line 21: Shouldn't this be called `"tokens"` (note the plural), since it's a list of tokens?

scripts/deployment/main.py
```diff
@@ -158,6 +159,30 @@ def deploy_beamer(
...
+    for token_arguments in tokens:
+        token_address = token_arguments["token_address"]
+        if token_address == "deploy_mintable_token":
+            if token is None:
+                raise ValueError(
+                    "Expecting mintable token to be deployed. (option --deploy-mintable-token)"
+                )
+            token_address = token.address
+        decimals = (
+            web3.eth.contract(address=token_address, abi=CONTRACTS["MintableToken"][0])
```
istankovic, line 173: We could move `beamer.chain._load_ERC20_abi` to `beamer.util`, make it public and use it here.

scripts/deployment/ganache-local.json
```diff
@@ -17,7 +17,18 @@
                 "claim_request_extension": 3600,
                 "claim_period": 3600,
                 "challenge_period_extension": 3600
-            }
+            },
+            "token":
+                [
+                    {
+                        "token_address": "deploy_mintable_token",
```
istankovic, line 24: Nitpick: I feel like just `"mintable_token"` would be slightly better.

scripts/call_contracts.py
```diff
@@ -345,40 +345,35 @@ def whitelist(
...
 
-    tx_hash = request_manager.functions.updateFeeData(
-        protocol_fee_ppm, lp_fee_ppm, min_lp_fee
-    ).transact()
-    web3.eth.wait_for_transaction_receipt(tx_hash, poll_latency=1.0)
+    token_data = list(request_manager.functions.tokens(token_address).call())
+    for i, value in enumerate([transfer_limit, min_lp_fee, lp_fee_ppm, protocol_fee_ppm]):
+        token_data[i] = value or token_data[i]
+    # the last element of token data is collectedProtocolFees which is not part of the update
+    tx_receipt = transact(request_manager.functions.updateToken(token_address, *token_data[:-1]))
```
istankovic, line 374: Nitpick: similarly to a previous instance, I think we could avoid the `[:-1]` slice.

scripts/call_contracts.py
```diff
@@ -345,40 +345,35 @@ def whitelist(
...
 
-    print(f"Transaction sent, tx_hash: {tx_hash.hex()}")
+    print(f"Transaction sent, tx_hash: {tx_receipt['hash']}")
```
istankovic, line 376: What is the difference between the two? Isn't `transactionHash` part of the receipt, not `hash`?

istankovic (approved):
Looks great! Thanks for the helpful sequencing of commits, really appreciated. :heart: 
There are some minor nitpicks and a question, but overall great work.

beamer/tests/constants.py
```diff
@@ -15,4 +15,6 @@
 RM_C_FIELD_TERMINATION = 6
 
 # token fields
+RM_T_FIELD_TRANSFER_LIMIT = 0
+RM_T_FIELD_PROTOCOL_FEE_PPM = 3
 RM_T_FIELD_COLLECTED_PROTOCOL_FEES = 4
```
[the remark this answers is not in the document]
fredo: I was about to do the same... :)
I'm really annoyed that this is not accessible easily.

beamer/tests/util.py
fredo: nice!

scripts/deployment/main.py
fredo: good idea

scripts/call_contracts.py
fredo: nothing in particular, but in contrast to the other commands, I reused the `transact()` wrapper which we use in the agent to encapsulate everything from sending the transactions to waiting for the transaction receipt.
For consistency, we should go with one approach in the whole script though.

fredo: I like using `transact()`better because it comes with checks for failing transactions and once we check for the status of the tx_receipt it even automatically checks for transaction success or reverts.

fredo:
Thanks for the detailed feedback. I liked the suggestion to get rid of `[:-1]` very much!

Issue #1234, opened 2022-11-30
CLI script for calling functions of deployed contracts is outdated 


GabrielBuragev:
By following the [contributing guideline](https://docs.beamerbridge.com/contributing.html) on how to setup your local environment i noticed that our `call_contracts.py` CLI script lacks some functionality. 
It is important to fix these issues so we can have a smoother start-up process for newcomers as well as the dev team.
 
#### Expected behavior
Following the steps in the guideline should lead to a fully functional up & running system.

#### Actual behavior

When running the step:

> Whitelist the agent’s address: 
```
python scripts/call_contracts.py --keystore-file 0x1CEE82EEd89Bd5Be5bf2507a92a755dcF1D8e8dc.json \
                                 --password '' \
                                 --eth-rpc http://localhost:8545 \
                                 --deployment-dir deployments/ganache-local \
                                 whitelist 0x1CEE82EEd89Bd5Be5bf2507a92a755dcF1D8e8dc
```
At first look everything seems to work. But when i ran the next command, for starting the beamer agent, i got this error:
![image](https://user-images.githubusercontent.com/18466333/204752677-84784299-28ea-4ad7-b562-0775896fc9a4.png)

#### Steps to reproduce the behavior
- Checkout on branch `main`
- Follow the steps inside the [contributing guideline](https://docs.beamerbridge.com/contributing.html)
- Command "Start beamer-agent:" expected to fail

## Solution
Add the missing logic for whitelisting on both `RequestManager` & `FillManager` contracts inside `call_contracts.py`: https://github.com/beamer-bridge/beamer/blob/main/scripts/call_contracts.py#L325-L343

# Acceptance Criteria
- [ ] Missing logic for whitelisting is added inside `whitelist` CLI command

PR #1239, opened 2022-11-30
Misc fixes

### LINKED ISSUE #1230: Update contributing docs & improve "getting started" steps
There is a missing step to build the relayers, otherwise the agent won't run.
Additionally, we have to either fix the make target for relayers that the binaries are stored at `data/relayers` or add an additional step in the docs that one has to do that manually. I prefer the first option.

- [x] Add `make relayers` to contributing docs
- [x] change relayers target in make file so that binaries are stored in `data/relayers` 
- [x] add .gitignore for `beamer/data/relayers`
- [x] the relayers step should not add pkg package - it should be part of the package.json

Currently we can't know whether the docs under the contributing section https://docs.beamerbridge.com/contributing.html are still correct. I propose that we pack all the steps that are necessary to get started in a make "setup-local-dev-environment" and have a circleCI step that executes it in order to make sure that it is still correct.

- [ ] make a "setup-local-dev-environment" step in the make file
- [ ] setup circleCI to execute the above step
- [ ] Think on ways to keep in sync the steps in the make file with the docs contributing page

Other getting started issues. 
If you just run `make`  one would most probably get an error like this:
image

- [ ] Most people don't have prettier and prettier-plugin-solidity installed globally. Since our project relies on them, we should make sure that they are set as locally required dependencies.

- [ ] the request library https://pypi.org/project/requests/ is being imported directly in the code, but it is nowhere defined in the pyproject.toml file. Currently that works because web3 requires it. 

- [x] fix the typos in the 'beamer-agent' command

- [ ] the docs are written in a way that suggest that you can execute them from top to bottom in the same shell. The problem is that We first for example define a variable `TOKEN_ADDRESS`, but afterwards starts the agent - the agent output is in the current shell, so we have to open a new one. When you open a new one the `TOKEN_ADDRESS` that was defined in the other shell is not there. So one need to re-execute this command. 
- [ ] For MacOS users, `coreutils` is a requirement (`brew install coreutils`). It covers commands like `realpath` which is not found on MacOS by default and yet used by our `arbitrum.sh` script.

istankovic:
Fixes some issues mentioned in #1230.

The idea is not to resolve #1230 entirely, but at least to make the onboarding the new dev less painful.

Closes: #1234.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,6 @@ venv/
 
 vetur.config.js
 
+beamer/data/relayers
+relayer/relayer-*-x64
 addresses.json
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,8 @@ relayers:
 	yarn --cwd relayer add pkg
 	yarn --cwd relayer run pkg -t node18-linux-x64 -o relayer-node18-linux-x64 build/src/service.js
 	yarn --cwd relayer run pkg -t node18-macos-x64 -o relayer-node18-macos-x64 build/src/service.js
+	mkdir -p beamer/data/relayers
+	cp relayer/relayer-*-x64 beamer/data/relayers
 
 docs:
 	make -C docs html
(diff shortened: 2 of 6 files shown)


GabrielBuragev (approved):
LGTM

PR #1243, opened 2022-12-02
Use uint96 for claim id

fredo:
Optimization for storage

WithdrawClaimId and filler now fit into one slot


fredo:
wrong branch based on

PR #1244, opened 2022-12-02
contracts: use uint96 for claim id and nonce

fredo:
Optimization for storage

WithdrawClaimId and filler now fit into one slot


istankovic:
> WithdrawClaimId and filler now fit into one slot

I am curious: how can one easily verify that?
I know there's `--storage-layout` option to `solc`, but it's tedious to invoke it manually because of the paths...
Would be nice to have `Makefile` target that outputs storage information so we could also use this in PRs, like we use the gas profiles...

fredo:
> > WithdrawClaimId and filler now fit into one slot
> 
> I am curious: how can one easily verify that? I know there's `--storage-layout` option to `solc`, but it's tedious to invoke it manually because of the paths... Would be nice to have `Makefile` target that outputs storage information so we could also use this in PRs, like we use the gas profiles...

I looked briefly for tools, found [this one](https://github.com/Taka-Security/sslc), but looks outdated. Indeed I tried to run it, but it failed on the RequestManager. It's basically doing a static analysis on the source code.
I consider this out of scope for now, but sounds like a good idea to have it in the future. Please create an issue for it.

istankovic:
This PR is part of #1170.

istankovic:
We seem to have went from 288 down to 256 bytes for `Request`:
```
-      "numberOfBytes": "288"
+      "numberOfBytes": "256"
```

Nice!

istankovic (approved):
Looks good!

A side note: we might consider using Solidity user-defined value types, which are a zero cost abstraction, to make the types more readable and explicit, e.g.

```
type NonceType is uint96;
type ClaimId is uin96;
....
function foo(ClaimId claimId, ...)
```

PR #1245, opened 2022-12-02
Stakes to be uint96

fredo:
use uint96 for stakes so that we can access claim states with less read operations

### DIFF
--- a/contracts/contracts/RequestManager.sol
+++ b/contracts/contracts/RequestManager.sol
@@ -48,10 +48,10 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
     struct Claim {
         bytes32 requestId;
         address claimer;
-        uint256 claimerStake;
-        mapping(address => uint256) challengersStakes;
+        uint96 claimerStake;
+        mapping(address => uint96) challengersStakes;
         address lastChallenger;
-        uint256 challengerStakeTotal;
+        uint96 challengerStakeTotal;
         uint256 withdrawnAmount;
         uint256 termination;
         bytes32 fillId;
@@ -100,9 +100,9 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
         bytes32 indexed requestId,
         uint96 claimId,
         address claimer,
-        uint256 claimerStake,
+        uint96 claimerStake,
         address lastChallenger,
-        uint256 challengerStakeTotal,
+        uint96 challengerStakeTotal,
         uint256 termination,
         bytes32 fillId
     );
@@ -146,7 +146,7 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
 
     /// The minimum amount of source chain's native token that the claimer needs to
     /// provide when making a claim, as well in each round of the challenge game.
-    uint256 public immutable claimStake;
+    uint96 public immutable claimStake;
 
     /// The additional time given to claim a request. This value is added to the
     /// validity period of a request.
@@ -251,7 +251,7 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
     /// @param _claimPeriod Claim period, in seconds.
     /// @param _challengePeriodExtension Challenge period extension, in seconds.
     constructor(
-        uint256 _claimStake,
+        uint96 _claimStake,
         uint256 _claimRequestExtension,
         uint256 _claimPeriod,
         uint256 _challengePeriodExtension
@@ -442,7 +442,7 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
         Claim storage claim = claims[nonce];
         claim.requestId = requestId;
         claim.claimer = claimer;
-        claim.claimerStake = msg.value;
+        claim.claimerStake = uint96(msg.value);
         claim.termination = termination;
         claim.fillId = fillId;
 
@@ -450,7 +450,7 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
             requestId,
             nonce,
             claimer,
-            msg.value,
+            uint96(msg.value),
             address(0),
             0,
             termination,
@@ -495,22 +495,26 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
         Claim storage claim = claims[claimId];
         bytes32 requestId = claim.requestId;
         uint256 termination = claim.termination;
-        Request storage request = requests[requestId];
         require(block.timestamp < termination, "Claim expired");
-        require(request.filler == address(0), "Request already resolved");
         require(
-            !request.invalidFillIds[claim.fillId],
+            requests[requestId].filler == address(0),
+            "Request already resolved"
+        );
+        require(
+            !requests[requestId].invalidFillIds[claim.fillId],
             "Fill already invalidated"
         );
 
         uint256 periodExtension = challengePeriodExtension;
         address claimer = claim.claimer;
-        uint256 claimerStake = claim.claimerStake;
-        uint256 challengerStakeTotal = claim.challengerStakeTotal;
+        uint96 claimerStake = claim.claimerStake;
+        uint96 challengerStakeTotal = claim.challengerStakeTotal;
 
         if (claimerStake > challengerStakeTotal) {
             if (challengerStakeTotal == 0) {
-                periodExtension += finalityPeriods[request.targetChainId];
+                periodExtension += finalityPeriods[
+                    requests[requestId].targetChainId
+                ];
             }
             require(msg.sender != claimer, "Cannot challenge own claim");
             require(
@@ -526,12 +530,12 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
         }
 
         if (msg.sender == claimer) {
-            claimerStake += msg.value;
+            claimerStake += uint96(msg.value);
             claim.claimerStake = claimerStake;
         } else {
             claim.lastChallenger = msg.sender;
-            claim.challengersStakes[msg.sender] += msg.value;
-            challengerStakeTotal += msg.value;
+            claim.challengersStakes[msg.sender] += uint96(msg.value);
+            challengerStakeTotal += uint96(msg.value);
             claim.challengerStakeTotal = challengerStakeTotal;
         }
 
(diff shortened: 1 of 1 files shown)


istankovic:
> use uint96 for stakes so that we can access claim states with less read operations

Are you sure this is really the case?

From a cursory comparison of the generated optimised IR of
```
struct Claim1 {                                          
    uint96 x;                                            
}                                                        
                                                         
struct Claim2 {                                          
    uint256 y;                                           
}                                                        
                                                         
contract Foo {                                           
    mapping (uint256 => Claim1) claims1;                 
    mapping (uint256 => Claim2) claims2;                 
                                                         
   function bar(uint256 i) public returns (uint96) {     
        uint96 b = claims1[i].x;                         
        return b;                                        
    }                                                    
                                                         
    function baz(uint256 i) public returns (uint256) {   
        uint256 z = claims2[i].y;                        
        return z;                                        
    }                                                    
```

I can see that it is in fact the `uint96` version that does more work:
```
function shift_right_0_unsigned(value) -> newValue                                                                         
{ newValue := shr(0, value) }                                                                                              
function cleanup_from_storage_t_uint96(value) -> cleaned                                                                   
{                                                                                                                          
    cleaned := and(value, 0xffffffffffffffffffffffff)                                                                      
}                                                                                                                          
function extract_from_storage_value_offset_0t_uint96(slot_value) -> value                                                  
{                                                                                                                          
    value := cleanup_from_storage_t_uint96(shift_right_0_unsigned(slot_value))                                             
}                                                                                                                          
function read_from_storage_split_offset_0_t_uint96(slot) -> value                                                          
{                                                                                                                          
    value := extract_from_storage_value_offset_0t_uint96(sload(slot))                                                      
}                                                                                                                          
/// @ast-id 33 @src 0:168:274  "function bar(uint256 i) public returns (uint96) {..."                                      
function fun_bar_33(var_i_18) -> var__21                                                                                   
{                                                                                                                          
    /// @src 0:208:214  "uint96"                                                                                           
    let zero_t_uint96_1 := zero_value_for_split_t_uint96()                                                                 
    var__21 := zero_t_uint96_1                                                                                             
    /// @src 0:237:244  "claims1"                                                                                          
    let _2_slot := 0x00                                                                                                    
    let expr_25_slot := _2_slot                                                                                            
    /// @src 0:245:246  "i"                                                                                                
    let _3 := var_i_18                                                                                                     
    let expr_26 := _3                                                                                                      
    /// @src 0:237:247  "claims1[i]"                                                                                       
    let _4 := mapping_index_access_t_mapping$_t_uint256_$_t_struct$_Claim1_$3_storage_$_of_t_uint256(expr_25_slot, expr_26)
    let _5_slot := _4                                                                                                      
    let expr_27_slot := _5_slot                                                                                            
    /// @src 0:237:249  "claims1[i].x"                                                                                     
    let _6 := add(expr_27_slot, 0)                                                                                         
    let _7 := read_from_storage_split_offset_0_t_uint96(_6)                                                                
    let expr_28 := _7                                                                                                      
    /// @src 0:226:249  "uint96 b = claims1[i].x"                                                                          
    let var_b_24 := expr_28                                                                                                
    /// @src 0:266:267  "b"                                                                                                
    let _8 := var_b_24                                                                                                     
    let expr_30 := _8                                                                                                      
    /// @src 0:259:267  "return b"                                                                                         
    var__21 := expr_30                                                                                                     
    leave                                                                                                                  
}                                                                                                                          
/// @src 0:67:477  "contract Foo {..."                                                                                     
```

Compare this to
```
function cleanup_from_storage_t_uint256(value) -> cleaned                                                                   
{ cleaned := value }                                                                                                        
function extract_from_storage_value_offset_0t_uint256(slot_value) -> value                                                  
{                                                                                                                           
    value := cleanup_from_storage_t_uint256(shift_right_0_unsigned(slot_value))                                             
}                                                                                                                           
function read_from_storage_split_offset_0_t_uint256(slot) -> value                                                          
{                                                                                                                           
    value := extract_from_storage_value_offset_0t_uint256(sload(slot))                                                      
}                                                                                                                           
/// @ast-id 50 @src 0:280:388  "function baz(uint256 i) public returns (uint256) {..."                                      
function fun_baz_50(var_i_35) -> var__38                                                                                    
{                                                                                                                           
    /// @src 0:320:327  "uint256"                                                                                           
    let zero_t_uint256_9 := zero_value_for_split_t_uint256()                                                                
    var__38 := zero_t_uint256_9                                                                                             
    /// @src 0:351:358  "claims2"                                                                                           
    let _10_slot := 0x01                                                                                                    
    let expr_42_slot := _10_slot                                                                                            
    /// @src 0:359:360  "i"                                                                                                 
    let _11 := var_i_35                                                                                                     
    let expr_43 := _11                                                                                                      
    /// @src 0:351:361  "claims2[i]"                                                                                        
    let _12 := mapping_index_access_t_mapping$_t_uint256_$_t_struct$_Claim2_$6_storage_$_of_t_uint256(expr_42_slot, expr_43)
    let _13_slot := _12                                                                                                     
    let expr_44_slot := _13_slot                                                                                            
    /// @src 0:351:363  "claims2[i].y"                                                                                      
    let _14 := add(expr_44_slot, 0)                                                                                         
    let _15 := read_from_storage_split_offset_0_t_uint256(_14)                                                              
    let expr_45 := _15                                                                                                      
    /// @src 0:339:363  "uint256 z = claims2[i].y"                                                                          
    let var_z_41 := expr_45                                                                                                 
    /// @src 0:380:381  "z"                                                                                                 
    let _16 := var_z_41                                                                                                     
    let expr_47 := _16                                                                                                      
    /// @src 0:373:381  "return z"                                                                                          
    var__38 := expr_47                                                                                                      
    leave                                                                                                                   
}                                                                                                                           
```

Behind the scenes both version use `sload` naturally, however the `uint96` version then has to do a mask operation.
At the same time, `sload` has static gas cost of 0, and dynamic gas cost either 100 (warm case) or 2100 (cold case).
So actually I do not see the benefit you have mentioned here.

Moreover, consider the case of
```
  uint96(msg.value)
```
which will need to do additional operations as well. Here the question would be, do we maybe want to actually check and make sure that the result fits in the 96 bits?

Looking at all of this, unless there is something I'm missing, I would not do this change.

fredo:
Thanks for taking the effort for this test! 

Unfortunately, your test is not representable for what the PR is trying to achieve.

Let me try to give more clarity here:

### A better representative example

Your example stores only one value per struct. Per default, Solidity's store slot size ist 256 bits. So any value of a struct will be stored in at least 256 bits slots. Looking at your example, it is clear why uint96 creates more work. Both values are stored in a 256 bit slot but the uint96 value has to be masked.

However, Solidity tries to be smart and pack values into one storage slot if possible. So if there were more attributes than one in your structs AND those values can be packed together into one slot, this would be the correct representation for a test.
Example:

```
struct Claim1 {                                          
    uint96 x;
    address y;                                            
} 
```
```
struct Claim2 {                                          
    uint256 x;
    address y;                                            
} 
```

Note that address takes 160 bits, so Claim1 should be possible to be stored in one 256 bits slot whereas in Claim 2, two 256 bits slots would be needed.

`sload` reads 256 bits slots, so the idea behind the change is that one sload (with maybe some additional masking work) can read two parameters of the struct (x and y).

### The actual desired optimization
It is also worth mentioning the actual desired optimization. As described in #1170, in `challengeClaim()` `lastChallenger` is always updated (results in a `SSTORE` operation) even if it is the same value as before. Explicitly this happens, when the same agent counter challenges an ongoing challenge (he was the `lastChallenger` and is it now again). 
In order to conditionally update lastChallenger, we have to read the value first (`SREAD`) and than compare it to the new value (as you can see in the PR).
The reason for trying to squeeze stakes and addresses together into one slot is that we now don't need an additional `SREAD`operation, because `lastChallenger`and `challengerStakesTotal` is now stored in the same slot and we have to read `challengerStakesTotal` anyway. We basically get the read "for free".

### Uint96(msg.value)
This is pretty safe, as there are 120 million ETH in existence which can be stored already in a uint88. If I remember correctly with uint96 we can represent 79e18 ETH which is ~10 orders of magnitude greater than the total number of ETH in existence. It is principally impossible to reach this boundary.


As all of this was a suggestion by the audit company and I did a theoretical research on this to approve, I'd appreciate if you redo the test with a more representative example and have in mind what the original goal of the test is. Also the measure of interest is gas cost (not only number of instructions)

If you have any further questions please reread the issue or just ask me. I'm curious to see what you will find out.

contracts/contracts/RequestManager.sol
```diff
@@ -444,22 +444,27 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+            !requests[requestId].invalidFillIds[claim.fillId],
             "Fill already invalidated"
         );
 
         uint256 periodExtension = challengePeriodExtension;
         address claimer = claim.claimer;
-        uint256 claimerStake = claim.claimerStake;
-        uint256 challengerStakeTotal = claim.challengerStakeTotal;
+        uint96 claimerStake = claim.claimerStake;
+        address lastChallenger = claim.lastChallenger;
```
istankovic, line 460: We should remove this read of `lastChallenger`.

contracts/contracts/RequestManager.sol
```diff
@@ -475,12 +480,14 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
-            claimerStake += msg.value;
+            claimerStake += uint96(msg.value);
             claim.claimerStake = claimerStake;
         } else {
-            claim.lastChallenger = msg.sender;
-            claim.challengersStakes[msg.sender] += msg.value;
-            challengerStakeTotal += msg.value;
+            if (msg.sender != lastChallenger) {
+                claim.lastChallenger = msg.sender;
+            }
```
istankovic, line 488: We should remove this `if` and update the `lastChallenger` unconditionally, as before.

istankovic (changes requested):
Per our discussion, we decided to do the change, but not in order to reduce the number of read operations (current solc does not do that optimization), rather to reduce the `Claim` storage requirements.

Please also make sure to update the commit message to reflect the new reasoning behind the change.

fredo:
I guess #1267 has to be merged first?

GabrielBuragev:
#1267 is merged 👍

istankovic:
@fredo could you also please update the commit message?

istankovic (approved):
Looks good!

PR #1246, opened 2022-12-05
Docs: Update contributing section

bilbeyt:


.gitignore
```diff
@@ -14,3 +14,6 @@ vetur.config.js
 beamer/data/relayers
 relayer/relayer-*-x64
 addresses.json
+0x1CEE82EEd89Bd5Be5bf2507a92a755dcF1D8e8dc.json
+agent-ganache-local.conf
+deployments/ganache-local
```
istankovic, line 19: I don't think we should add these: these are just examples from the docs, not something that is always there. Those paths could be named arbitrarily.

docs/source/contributing.rst
```diff
@@ -8,6 +8,8 @@ Requirements
 - `poetry <https://python-poetry.org>`_
 - `ganache <https://trufflesuite.com/ganache>`_ 7
 - `prettier-plugin-solidity <https://github.com/prettier-solidity/prettier-plugin-solidity>`_
+- `jq <https://stedolan.github.io/jq/>`
+- `realpath <https://www.gnu.org/software/coreutils/manual/html_node/realpath-invocation.html#realpath-invocation>`
```
istankovic, line 12: I think we should mention `realpath` only for macOS case since otherwise it's pretty much there.
And I remember that @GabrielBuragev said that can be installed via `coreutils`:
```
brew install coreutils
```
(please correct me if I'm wrong).

So instead of mentioning `realpath`, I would just mention to install `coreutils` if on mac.

docs/source/contributing.rst
```diff
@@ -90,12 +92,17 @@ Submit a request::
                                      --amount 1 \
                                      --target-address 0x1CEE82EEd89Bd5Be5bf2507a92a755dcF1D8e8dc \
                                      --target-chain-id 1337 \
-                                     --target-token-address $TOKEN_ADDR
+                                     --target-token-address $TOKEN_ADDR \
+                                     --validity-period 2700
```
istankovic, line 96: Why do we need `--validity-period` option here?

docs/source/contributing.rst
```diff
@@ -90,12 +92,17 @@ Submit a request::
...
 
 
 Working with a local Optimism instance
 --------------------------------------
 
+For MacOS add an alias to your bash for mktemp as follows::
+
+    alias mktemp=gmktemp
```
istankovic, line 104: Eh, it would be nice if our scripts did not require that. And I think @GabrielBuragev already had a fix that avoid the GNU-specific, option, is that right, Gabriel?

docs/source/contributing.rst
```diff
@@ -116,6 +123,9 @@ To list Optimism contracts' addresses, run::
     sh ./docker/optimism/optimism.sh addresses
 
 
+Note: Networks are connected as submodules. Consider this when you clone/pull.
+
+
```
istankovic, line 128: Could you please explain what this is about?

istankovic (changes requested):

docs/source/contributing.rst
GabrielBuragev: Exactly @istankovic.

This needs to be installed: https://formulae.brew.sh/formula/coreutils

docs/source/contributing.rst
GabrielBuragev: Yes, exactly. That fix can be applied on the code level. 
Should i submit it as a separate PR or do you want to include it in this PR?

The change is the following:

change [this](https://github.com/beamer-bridge/beamer/blob/main/docker/optimism/optimism.sh#L33) line to: 
`WAIT_FOR_SEQUENCER_SCRIPT=$(mktemp)-wait-for-sequencer.sh`

docs/source/contributing.rst
GabrielBuragev: I believe he is referring to the extra step required to pull the submodules after doing a `git clone`: 
`git submodule update --init` OR `git submodule --init`.

Anyways, i believe we should add the `git submodule update --init` as a required command before starting any local rollup instance.

docs/source/contributing.rst
istankovic: Yes, @GabrielBuragev, I prefer your fix here, please submit a PR and then we can drop this change.

docs/source/contributing.rst
istankovic: Then I would rather fix the script and change the default value to be valid, according to the current min/max validity in the contract. This way, no option would be needed. I suggest to just set it to the current minimum validity period, which is 30 min.

.gitignore
```diff
@@ -13,4 +13,4 @@ vetur.config.js
 
 beamer/data/relayers
 relayer/relayer-*-x64
-addresses.json
+addresses.json
```
istankovic, line 16: Nitpick: unnecessary newline.

docs/source/contributing.rst
```diff
@@ -96,6 +98,12 @@ Submit a request::
 Working with a local Optimism instance
 --------------------------------------
 
+Before starting any local rollup, use::
+
+    git submodule update --init
+
+to pull rollup.
```
istankovic, line 105: I would move this into a new section, at the top, e.g.

```
Setting up the repository
--------------------------------------

Clone the repository:
git clone  --recurse-submodules git@github.com:beamer-bridge/beamer.git

If you already have a repository, make sure the submodules are up to date:
git submodule update
```

istankovic (approved):
Thanks!

docs/source/contributing.rst
GabrielBuragev: So, i see one problem with the first command here. There are people that don't use SSH keys for connecting to github (don't ask me why) and therefore it might fail due to unrecognized host/public key.
In order to be sure that this command will never fail, i suggest we use the HTTPS protocol here since that is supported by default when using git.

Also i would suggest to merge these two steps into one like so:
```
Setting up the repository
--------------------------------------

Clone the repository:
git clone https://github.com/beamer-bridge/beamer.git && 
cd beamer && 
git submodule update --init
```

What do you think about this?

GabrielBuragev (approved):
Just left one minor comment on the "Setting up the repository" section. Would like to hear your opinion on that one.
Other than that, everything looks fine. Thanks for improving the docs!

docs/source/contributing.rst
istankovic: > So, i see one problem with the first command here. There are people that don't use SSH keys for connecting to github (don't ask me why) and therefore it might fail due to unrecognized host/public key.
> In order to be sure that this command will never fail, i suggest we use the HTTPS protocol here since that is supported by default when using git.

Actually, that's a very good point and we should change it to use the https transport, not the git one.
Thanks for noticing, Gabriel.

I would leave the `--recurse-submodules` part though, because then you don't need to go and manually update the submodules -- the clone will do that for you.

GabrielBuragev: Alright, lets stick with the `--recurse-submodules`. 
So we can do it like:
```
Setting up the repository
--------------------------------------

Clone the repository:
git clone  --recurse-submodules https://github.com/beamer-bridge/beamer.git && cd beamer

If you already have a repository, make sure the submodules are up to date:
git submodule update

istankovic: Right, and you don't even need the `&& cd beamer` (that only made sense with the submodule update following it immediately).

GabrielBuragev: hmm, but the next steps after setting up the repository should be done inside the repo folder.

istankovic: > hmm, but the next steps after setting up the repository should be done inside the repo folder.

Yes, but the general assumption is whoever is trying to clone & use the code is capable of deducing that fact for themselves. :-)

GabrielBuragev: Well, if we go in that direction, then we should also consider the SSH -> HTTPS edit of the first command as superfluous :P

GabrielBuragev: IMO, its just a really nice feeling when things simply work by copying & pasting the commands one by one. We want anyone to be able to contribute & understand the setup, not just techies.

istankovic: > IMO, its just a really nice feeling when things simply work by copying & pasting the commands one by one. We want anyone to be able to contribute & understand the setup, not just techies.

Yeah, I see where you're coming from. On the other hand, the fact is that we simply list the requirements and expect people to be able to install them on their own. It's a tricky balance between giving people ready-to-use commands and not giving too much so that they do not see the forest for the trees.

In any case, I don't feel strongly about it so it's fine with me.

bilbeyt: I don't see so much difference so i will stick with one liner command and will change it to use https instead of ssh. The person who will work on this can easily use cd command imo.

PR #1248, opened 2022-12-05
docker: fix mktemp argument error on macOS

GabrielBuragev:
Fixes the cross platform issue when calling `mktemp`.

The argument `--suffix` is not a recognized argument for the `mktemp` tool inside Mac OS.


istankovic (approved):
Thanks!

PR #1249, opened 2022-12-05
Improve contributing docs & add missing info

GabrielBuragev:
While working with different versions of Node.js locally (by using `nvm`) i faced the issue where the different Node.js versions were using different versions of `prettier-plugin-solidity` installed globally. I realized that this pain can be avoided by not depending on this prettier plugin globally but instead move it to the dependency list of the project.

I managed to pull out the `prettier-plugin-solidity@1.0.0-beta.19` npm package from the required globally installed packages to a locally installed one.
It is now installable via a simple `npm install` command in the root of the project. 

_Note:_ I chose to use `npm install` instead of `yarn install` here as `npm` comes pre-installed as a part of the Node.js installation and therefore no additional commands are required.  

This PR contains the following changes:
- Removed `prettier-plugin-solidity` from the requirements
- Added Node.js 18 as a requirement. We kinda overlooked this critical dependency since most of us have it installed by "default" on our machines but that is not the case for many people.
- Adjusted the `.circleci` config file to use the newest method of installing the node dependencies

Question is, should we also have `yarn` as a requirement?
I see that it is being used inside the `Makefile` for building the relayer.


.circleci/workflows.yml
```diff
@@ -124,7 +124,7 @@ jobs:
       - run: npm install yarn --location=global
       - run: npm install solhint --location=global
       - run: npm install ganache --location=global
-      - run: npm install prettier prettier-plugin-solidity@1.0.0-beta.19 --location=global
```
istankovic, line 127: I wonder whether we should also do the same thing for yarn, solhint and ganache.
Then all the packages would be installed locally, under `node_modules`, right?

package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "devDependencies": {
+    "prettier": "^2.8.0",
+    "prettier-plugin-solidity": "1.0.0-beta.19"
+  }
+}
```
istankovic, line 6: I have a few questions regarding this:

- does this only work if `package.json` is on the top level?
- how will this interact with `package.json` from the frontend?
- if I say `npm install`, that will store all packages under `node_modules`, which means there would possibly be duplicates between the frontend `node_modules` and toplevel `node_modules`, right?

docs/source/contributing.rst
```diff
@@ -7,13 +7,16 @@ Requirements
...
-
+- `Node.js <https://nodejs.org/en/download/>`_ 18
 
 Testing
 -------
 
-First make sure beamer is installed::
+First make sure that the node dependencies are installed::
+
+    npm install
```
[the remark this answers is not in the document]
GabrielBuragev: Yes, you are right. I will enforce that.

.circleci/workflows.yml
GabrielBuragev: We could do that yes. 
The only difference, when having them locally installed, would be in the way they are being invoked.

For globally installed npm packages, most of the times they are being aliased therefore you can run them as any other OS installed package (for example by running the command `ganache`).

For locally installed npm packages, they are usually ran by using `npx` which is a package runner. `npx` comes as a part of `npm` so no need for extra commands/dependencies.

I have to see where they are being used, and how they are used and adjust that accordingly.

package.json
GabrielBuragev: - Yes, `package.json` needs to be in the root for this to work.
- There is no interaction between these 2 folders. Each folder that has a separate `package.json` file is considered its own "project" by npm
- Yes, there can be duplicates in the separate `node_modules` folder but that is not a big problem. 
A little explainer on how npm looks for the `package.json` file when you run `npm install`: 
When you run `npm install` npm goes bottom-to-top trying to find the first `package.json` it can use for the installation process. So if you are inside the `frontend` folder npm will use the `package.json` that is found inside that folder. when you are in `contracts` folder for example, `npm install` will use the `package.json` that is located in the top-most folder (the root). Same goes when you are in the root folder.

istankovic: I see, thanks for the explanation.

docs/source/contributing.rst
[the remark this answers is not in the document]
GabrielBuragev: ✅

.circleci/workflows.yml
```diff
@@ -122,9 +122,7 @@ jobs:
             # Remove other versions so they don't end up in the cache.
             for x in $(ls -1 ${NVM_DIR}/versions/node | grep -v $(nvm version default)); do nvm uninstall $x; done
       - run: npm install yarn --location=global
```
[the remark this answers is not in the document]
GabrielBuragev: Because yarn is yet another package manager, same like npm. 
It is preferred that it is installed globally because otherwise it would not make much sense to run it via another package manager like: `npx yarn ...`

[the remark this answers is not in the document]
GabrielBuragev: For that question, i believe we need to loop in @manuelwedler and get his input on why yarn was used in the first place. I have no strong opinion on using strictly yarn or npm, i am up for switching to npm as well.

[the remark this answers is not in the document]
manuelwedler: First of all, we had some bad experiences in Raiden with npm and yarn was always reliable. For example, I had flaky installations with npm. That's why we started to use yarn basically everywhere. 

I don't completely get what should be achieved here but I have a lot of questions..
- Why is yarn installed and then, npm is used for installing the dependencies (in this step)? Not sure in which directory this is run, but as far as I see we only have yarn.lock files (and not the npm lock files) in our repo. So we should also use yarn in the CI.
- Why don't we use an image that has node preinstalled? They usually come with the package managers installed already I think.
- We have this circleci/node orb declared on the top. Do I get it correctly that we only use it for the frontend workflow?

One important hint: Always use `yarn install --frozen-lockfile` or the corresponding `npm ci` for installing packages in CI environments. The mentioned orb does that automatically.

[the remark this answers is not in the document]
istankovic: > I don't completely get what should be achieved here but I have a lot of questions..

:-)

> Why is yarn installed and then, npm is used for installing the dependencies (in this step)? Not sure in which directory this is run, but as far as I see we only have yarn.lock files (and not the npm lock files) in our repo. So we should also use yarn in the CI.

I think the reason was purely historical as Paul came up with the original version of the relayer and I didn't really know to mess too much with it due to my lack of knowledge of the node ecosystem and best practices.
If yarn is the preferred way to install packages, then, let's switch to it, by all means.  :+1: 

> Why don't we use an image that has node preinstalled? They usually come with the package managers installed already I think.

Eh... this is where it gets a bit complicated. We can't use simple Docker containers for our tests because the networking setup will not work (long story). So we had to go with a VM. But in that VM we need both Python and node packages, which is the reason we cache the entire node runtime + packages.

> We have this circleci/node orb declared on the top. Do I get it correctly that we only use it for the frontend workflow?

Yes. The frontend is completely separate and there is no need to run that workflow in a VM, which itself incurs ~25s overhead, compared to a <1s for a Docker container.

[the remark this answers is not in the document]
manuelwedler: I'm all for just using yarn. You need to run the installation locally with `yarn install` to generate a `yarn.lock` file and replace the `package-lock.json` here with it.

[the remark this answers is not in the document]
GabrielBuragev: ✅

GabrielBuragev:
@manuelwedler @istankovic can you have another look please?

Switched to using `yarn` exclusively, adapted the commit messages, docs & circle CI config.

.gitignore
```diff
@@ -14,3 +14,5 @@ vetur.config.js
 beamer/data/relayers
 relayer/relayer-*-x64
 addresses.json
+
+node_modules
```
istankovic, line 18: I understand that `node_modules` is no longer used with yarn -- could you please clarify whether that is the case?
My impression is that yarn now stores cached packages under `.yarn`.

docs/source/contributing.rst
```diff
@@ -6,14 +6,17 @@ Requirements
 
 - `Python <https://www.python.org>`_ 3.10
 - `poetry <https://python-poetry.org>`_
-- `ganache <https://trufflesuite.com/ganache>`_ 7
-- `prettier-plugin-solidity <https://github.com/prettier-solidity/prettier-plugin-solidity>`_
-
+- `Node.js <https://nodejs.org/en/download/>`_ 18.x
+- `Yarn <https://classic.yarnpkg.com/en/>`_ 1.x
```
istankovic, line 10: Why don't we use version 2.x? yarn 2 was released more than 2 years ago.

.gitignore
GabrielBuragev: All the node package managers use `node_modules` and yarn is not an exception.

.circleci/workflows.yml
```diff
@@ -122,9 +122,7 @@ jobs:
...
-      - run: npm install solhint --location=global
-      - run: npm install ganache --location=global
-      - run: npm install prettier prettier-plugin-solidity@1.0.0-beta.19 --location=global
+      - run: yarn install --frozen-lockfile
       - save_cache:
           key: node18-packages-v10
```
istankovic, line 127: I think we should also update the `node18-packages-v10` to `v11`, to make sure CI is not using older cache for newer runs.

.gitignore
istankovic: Yeah, I just assumed we're going to use yarn 2, which does not use `node_modules` anymore.

istankovic (approved):
Looks fine, I would just update the node packages cache key.

.circleci/workflows.yml
GabrielBuragev: Good idea. Will do so 👍

.circleci/workflows.yml
```diff
@@ -26,7 +26,7 @@ commands:
       # Restore node runtime and packages at /opt/circleci/.nvm
       - restore_cache:
           name: Restore node runtime and packages
-          key: node18-packages-v10
+          key: node18-packages-v11
```
istankovic, line 29: @GabrielBuragev did you also update the other references in the same file?

.circleci/workflows.yml
istankovic, line 127: ... like this one? :-)

Issue #1229, opened 2022-11-28
Use brownie.chain.sleep instead of brownie.chain.mine


fredo:
brownie.chain.mine(timestamp=x) sets the timestamp of the next block to x+1. 
Typically we call this when we want to test a specific timestamp during execution of the next transaction.
brownie.chain.sleep sets the timestamp of ganache to the desired timestamp and then generates a new block with this timestamp. This is what we want.

Even brownie's documentation says that one should use sleep if a contract interaction should be tested on a specific timestamp.

fredo:
should wait for #1079

Issue #1079, opened 2022-09-27
Move from brownie to ape for our contract and agent tests


istankovic:
Brownie seems to be unmaintained and needs fixing which is frustrating. There is a fork if brownie called [ape](https://github.com/ApeWorX/ape) which seems to be more active. For the most part, it looks like a direct fork of brownie, with added features on top (e.g. support for multiple chains).

We should
- [x] convert contract tests to ape
- [x] convert agent tests to ape
- [x] update docs
- [x] remove `eth-brownie` from our dependencies (`pyproject.toml`)
- [x] update `pylintrc`

andfletcher:
possibly we need to shit this approach. for now, we will wait until ape team gets back to us and then decide on how to continue.

bilbeyt:
@andfletcher I have created 4 issues like we talked in the meeting, and mentioned this issue in their description.

istankovic:
After some more discussion with @bilbeyt, we determined #1286 is not really critical (though very much desired), so we're not marking it as being a blocker for this issue.

andfletcher:
@bilbeyt i understood that the 4 issues are created already within the ape organisation from them? Did i misunderstand it? How are we moving forward on checking progress on the issues, do we need to do them or are we just waiting for them to do it? I understood it in a way that there are issues and we just need to wait for them to finish them?

istankovic:
> How are we moving forward on checking progress on the issues, do we need to do them or are we just waiting for them to do it?

For the one regarding account creation (#1284), we asked them, but there's no guarantee when they're going to work on it, let alone when it will be available. They also told us we're welcome to work on it, which @bilbeyt already started.

For the other 3, we mentioned them, but unlike the 1st one, we have not yet reached an agreement on them. Once we reach an agreement, we will also probably write an issue and likely also contribute a patch.

These are the scenarios I can imagine:
1. we reach an agreement on all issues, write issues for ape and they implement all of them
2. we reach an agreement on all issues, write issues for ape and we implement all of them
3. we reach an agreement on all issues, write issues for ape and we implement some of them, the rest done by apeworx
4. we do not reach an agreement on all issues

I think scenarios 1 and 4 are unlikely, so that would leave 2 and 3, with 3 being less likely than 2.
@bilbeyt feel free to pitch in with your view.

bilbeyt:
They implemented something for #1286 already but when i checked it, it wasn't really working in my view. Tested with only one test but still.

[...]

PR #1253, opened 2022-12-06
Contracts: claim on behalf of others

### LINKED ISSUE #1201: Claiming on behalf of others
Currently an address only can claim for itself. This leads to a single point of failure risk.
We can open up this requirement to allow any address claim for another address. This would solve the fact that one address can actually be the filler, but another address can claim on behalf of the filler.

The interesting part is how the claim stake is handled and where it is coming from. As a first step, I would suggest to keep this question completely out of the protocol and let the involved parties decide this on their own.

### So what does it mean?

Anyone who is initiating the claim is responsible for paying the claim stake. This is necessary so that we protect the protocol from spams, or dragging any address into false claims.
The "claimer" in the claim struct remains to be the address which is submitted as a parameter to the claim function. The actual message sender is **NOT** the claimer but the (claim) initiator. However, the initiator must provide the initial claim stake. This means, it is the initiator who puts in stake and risks to lose it if it is a false claim.

### Claim stake distribution 
The easiest way would be, that the final claim stake distribution goes 100% back to the claimer. This means that the amount "claim stake" staked by the initiator is sent to the claimer. The claimer and the initiator have to have their own way to deal with this fact and this would be out of scope from a protocol point of view.
From my side this should be fine as a first implementation to this feature. The reason is that 1) it touches the protocol only slightly at first, 2) The parties could deal outside of this either in a trusted manner or in a form of a contract who deals with this exact relationship, 3) it opens up the possibility that the actual agent can run multiple processes to claim on behalf of the whitelisted address (initiator and claimer are different addresses but the same entity). 

If at one point there is a need for this feature to become a generalized service by a third party, we could reevaluate if further contract changes are needed.

### Claimer whitelist
In the latest contracts claimers are also whitelisted. This is because we wanted to reduce the pressure on the protocol due to false claims. I see three options how we can handle this. 
1) Agents who want to use this feature will have to ask us to whitelist additional addresses on the claiming side
2) we need to give up the whitelist and allow anybody to claim, opening up again the risk of system overload due to many claims
3) We check the claimer (not the initiator) to be a whitelisted address. In this case, a system overload can still occur but the only ones who can exploit funds are agents which we do have some sort of relationship to for now (also due to the whitelisting process)

#### Background: Why do we have a whitelist on the claiming side?
The reason for this was the fact, that our agents currently run with limited capacity in terms of funds. Additionally, the network of "watchers" is rather small in the bootstrapping phase (agents, voluntary watchers). In this phase it is easier for an attacker to lock up all eth, which are provided for challenges, in running challenges.
Once this happens, an attacker has 24h to claim a request and if nobody tops up the funds nobody will challenge it, and the attacker will be able to exploit the protocol.
The more ETH the agents will provide the less likely this is to happen. But with the current status of the network and with limited monitoring system in place, the decision was made to strengthen the security by introducing a more controlled structure. 

## Acceptance criteria
- [ ] @andfletcher checks with legal if we can do option 1
- [ ] We create a claim on behalf of others' whitelist
- [ ] Addresses on this list can claim on behave of others as long as they are on the whitelist
- [x] We can add and remove addresses from the whitelist

## Additional information
This only focuses on the contract changes that would allow a claim on behalf of someone. The software that will allow doing this is a separate issue that needs to be created @andfletcher

manuelwedler:
Closes #1201 

This overloads the `claimRequest` function with an additional parameter for setting the claimer. This means the claim can be initiated by another address than the claimer address. The initiator pays the stake. The claimer gets the stake distribution back. Adds a new modifier to the LpWhitelist in order to check if the given claimer is whitelisted.

### DIFF
--- a/beamer/tests/contracts/test_request_manager.py
+++ b/beamer/tests/contracts/test_request_manager.py
@@ -70,7 +70,7 @@ def test_claim(token, request_manager, claim_stake, deployer):
     blacklist_tx = request_manager.removeAllowedLp(claimer, {"from": deployer})
     assert "LpRemoved" in blacklist_tx.events
 
-    with brownie.reverts("Sender not whitelisted"):
+    with brownie.reverts("Not allowed"):
         request_manager.claimRequest(request_id, FILL_ID, {"from": claimer, "value": claim_stake})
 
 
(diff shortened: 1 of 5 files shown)


contracts/contracts/LpWhitelist.sol
```diff
@@ -29,6 +29,12 @@ contract LpWhitelist is Ownable {
         _;
     }
 
+    /// Modifier to check whether a given address is whitelisted
+    modifier whitelistedAddress(address addressToCheck) {
+        require(allowedLps[addressToCheck], "Given address not whitelisted");
+        _;
+    }
```
istankovic, line 36: I would propose to merge this with the `onlyWhitelist` modifier and so that we have just one modifier, e.g.
```
modifier onlyWhitelisted(address ...)
```
I would also shorten the error message to `"Not whitelisted"`.

Thinking further about the naming here, we have `allowedLps` and `whitelist` -- it would be nice to have some consistency there, e.g.

```
modifier onlyAllowed(address ...)
```
with the message then simply e.g. ` "Not allowed"`.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -99,6 +99,85 @@ def test_claim_with_different_stakes(token, request_manager, claim_stake):
...
+    assert claim_event["lastChallenger"] == brownie.ZERO_ADDRESS
+    assert claim_event["challengerStakeTotal"] == 0
+    assert claim_event["termination"] == expected_termination
+    assert claim_event["fillId"] == to_hex(FILL_ID)
+
+    assert web3.eth.get_balance(request_manager.address) == claim_stake
+    assert web3.eth.get_balance(initiator.address) == initiator_eth_balance - claim_stake
+    assert web3.eth.get_balance(claimer.address) == claimer_eth_balance
+
+    chain.mine(timedelta=claim_period)
```
fredo, line 153: it would be nice if chain.sleep is used to time travel

manuelwedler: We use chain.mine basically everywhere in the test to time travel. Should it be replaced everywhere?

fredo: ideally yes. But I think chain.sleep is limited to timedelta. So don't spend too much time on it

fredo: We have #1229, so from my side this will be tackled anyway. You don't need to change here anything, especially with #1079 in the making

contracts/contracts/RequestManager.sol
```diff
@@ -371,6 +371,31 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+    ///
+    /// @param claimer Address of the claimer.
+    /// @param requestId ID of the request.
+    /// @param fillId The fill ID.
+    /// @return The claim ID.
+    function claimRequest(
+        address claimer,
+        bytes32 requestId,
+        bytes32 fillId
+    )
```
istankovic, line 393: While this is totally fine, I think we should really reconsider the naming here.
For example,
```
    /// Claim that a request was filled by a claimer.
```
looks confusing and while I understand what is being meant here, we're not making it easy for other people.

I could not come up with an ideal naming, but my thinking goes like this:
- the word `claimer` should be used to denote whoever actually called `claimRequest`, regardless of the variant of that function
- the word `filler` should be used to denote whoever the `claimer` claims to have been the filler of the request

With this we would not need to distinguish between a "claimer" and an "initiator".
With such naming, we would also replace the `claimer` field in `Claim` struct with a `filler` field:
```
struct Claim {                                    
    bytes32 requestId;                            
-    address claimer;                              
+   address filler;                              
    uint256 claimerStake;                         
    ...
}                                                 
```

This may look a slightly confusing at first, but it does make sense: the `claimer` claims that `filler` filled the request. We may not know whether that is true or not yet, until the challenge game is done. Also note that `claimerStake` is unchanged -- it is the `claimer` that puts (and possibly loses) their stake, not the `filler`.

So I really think we should consider changing naming here to reflect the new possibilities, but at the same time, if you decide to do it, I would suggest to do it in another PR, to keep this one simple.

istankovic (approved):
Looking good!

I do think we should improve the naming, but that can be done in another PR.

contracts/contracts/RequestManager.sol
fredo: I get your point, although I'm not really happy to use the term filler as long as it is unclear whether it is really the filler. For example, to my understanding, a filler is the real filler of the request. In the Request Manager any address is accepted as the filler once a) a claim is finalized and valid or b) l1 resolved. In both cases we write the `filler` in the request struct.
It would be confusing to me to have a bunch of claims with lots of different fillers. So I would really consider keeping a unique name for a valid and accepted filler, and for an address who is claiming to be the filler.
Additionally, we also need a good naming for whoever called the function.

> Also note that claimerStake is unchanged -- it is the claimer that puts (and possibly loses) their stake, not the filler.

This is not quite true, or at least I would not argue in the way you just did. Technically, the msg.sender puts in the claim stake at first, but it is denominated to be the stake of the (in your case) filler. To be more precise, allthough the claimer puts in the first claim stake, any other counter challenge the actual filler has to increase its stake. At the end all of the stake goes back to the filler.
Note, that how the filler and the claimer deal with this "transfer" of claim stake from the claimer to the filler is out of the protocol and is not "our business". 

In any case, I would not name the stake after the caller of the transaction.

Issue #613, opened 2022-04-28
contracts: Allow withdraw on behalf of challengers


fredo:
It might be of interest to clean up the balance sheet of the contract to be able to withdraw on behalf of a challenger.

PR #1258, opened 2022-12-07
Withdraw on behalf of others

fredo:
implments: #613 

Edit: Note that in line 651 msg.Sender was changed to claimReceiver (not claimResolver)

### DIFF
--- a/beamer/tests/contracts/test_request_manager.py
+++ b/beamer/tests/contracts/test_request_manager.py
@@ -442,8 +442,7 @@ def test_withdraw_without_challenge(request_manager, token, claim_stake, claim_p
     # Timetravel after claim period
     chain.mine(timedelta=claim_period)
 
-    # Even if the requester calls withdraw, the deposit goes to the claimer
-    withdraw_tx = request_manager.withdraw(claim_id, {"from": requester})
+    withdraw_tx = request_manager.withdraw(claim_id, {"from": claimer})
     assert "DepositWithdrawn" in withdraw_tx.events
     assert "ClaimStakeWithdrawn" in withdraw_tx.events
     assert request_manager.isWithdrawn(request_id)
@@ -567,8 +566,7 @@ def test_withdraw_with_two_claims(deployer, request_manager, token, claim_stake,
     assert web3.eth.get_balance(request_manager.address) == 2 * claim_stake
 
     # The first claim gets withdrawn first
-    # Even if the requester calls withdraw, the deposit goes to the claimer1
-    withdraw1_tx = request_manager.withdraw(claim1_id, {"from": requester})
+    withdraw1_tx = request_manager.withdraw(claim1_id, {"from": claimer1})
     assert "DepositWithdrawn" in withdraw1_tx.events
     assert "ClaimStakeWithdrawn" in withdraw1_tx.events
 
@@ -587,7 +585,7 @@ def test_withdraw_with_two_claims(deployer, request_manager, token, claim_stake,
     # The other claim must be withdrawable, but the claim stakes go to the
     # contract owner as it is a false claim but no challenger exists.
     with earnings(web3, deployer) as owner_earnings:
-        withdraw2_tx = request_manager.withdraw(claim2_id, {"from": requester})
+        withdraw2_tx = request_manager.withdraw(claimer2, claim2_id, {"from": requester})
     assert "DepositWithdrawn" not in withdraw2_tx.events
     assert "ClaimStakeWithdrawn" in withdraw2_tx.events
 
@@ -725,8 +723,7 @@ def test_withdraw_with_two_claims_and_challenge(request_manager, token, claim_st
     assert web3.eth.get_balance(request_manager.address) == 3 * claim_stake + 1
 
     # The first claim gets withdrawn first
-    # Even if the requester calls withdraw, the deposit goes to the claimer1
-    withdraw1_tx = request_manager.withdraw(claim1_id, {"from": requester})
+    withdraw1_tx = request_manager.withdraw(claim1_id, {"from": claimer1})
     assert "DepositWithdrawn" in withdraw1_tx.events
     assert "ClaimStakeWithdrawn" in withdraw1_tx.events
     assert request_manager.isWithdrawn(request_id)
@@ -744,7 +741,7 @@ def test_withdraw_with_two_claims_and_challenge(request_manager, token, claim_st
     with brownie.reverts("Claim already withdrawn"):
         request_manager.withdraw(claim1_id, {"from": claimer1})
 
-    # The other claim must be withdrawable, but mustn't transfer tokens again
+    # The other claim must be withdrawable, but must not transfer tokens again
     request_manager.withdraw(claim2_id, {"from": challenger})
 
     assert token.balanceOf(requester) == 0
@@ -761,7 +758,7 @@ def test_withdraw_with_two_claims_first_unsuccessful_then_successful(
     request_manager, token, claim_stake, claim_period, finality_period
 ):
     """Test withdraw when a request was claimed twice. The first claim fails, while the second
-    is successful and should be paid out the request funds."""
+    is successful and should pay out the request funds."""
     requester, challenger = alloc_accounts(2)
     claimer1, claimer2 = alloc_whitelisted_accounts(2, {request_manager})
     transfer_amount = 23
@@ -829,9 +826,10 @@ def test_withdraw_with_two_claims_first_unsuccessful_then_successful(
     with brownie.reverts("Claim already withdrawn"):
         request_manager.withdraw(claim1_id, {"from": claimer1})
 
-    # The other claim must be withdrawable, but mustn't transfer the request tokens again
-    withdraw2_tx = request_manager.withdraw(claim2_id, {"from": requester})
+    # The other claim must be withdrawable and should pay out the funds
+    withdraw2_tx = request_manager.withdraw(claim2_id, {"from": claimer2})
     assert "ClaimStakeWithdrawn" in withdraw2_tx.events
+    assert "DepositWithdrawn" in withdraw2_tx.events
 
     assert token.balanceOf(requester) == 0
     assert token.balanceOf(claimer1) == 0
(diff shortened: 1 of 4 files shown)


contracts/contracts/RequestManager.sol
```diff
@@ -606,25 +630,25 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
-            ethToTransfer = claimerStake + challengerStakeTotal;
             claimReceiver = claimer;
+            ethToTransfer = claimerStake + challengerStakeTotal;
         } else if (challengerStakeTotal > 0) {
-            // If claim is invalid, partial withdrawal by the sender
-            ethToTransfer = 2 * claim.challengersStakes[msg.sender];
-            claimReceiver = msg.sender;
+            // If claim is invalid, partial withdrawal by the claimResolver
+            claimReceiver = claimResolver;
+            ethToTransfer = 2 * claim.challengersStakes[claimReceiver];
```
bilbeyt, line 638: Why are we multiplying with 2 here and saying it is partial withdraw?

beamer/tests/contracts/test_request_manager.py
```diff
@@ -1037,6 +1037,50 @@ def test_challenge_after_l1_resolution(request_manager, token, claim_stake, cont
...
+        web3.eth.get_balance(second_challenger.address)
+        == second_challenger_eth_balance + claim_stake
+    )
+
+    with brownie.reverts("Challenger has nothing to withdraw"):
+        request_manager.withdraw(claim_id, {"from": second_challenger})
+
+    request_manager.withdraw(first_challenger, claim_id, {"from": first_challenger})
+
+    assert (
```
bilbeyt, line 1078: If second challenger won the game, why first challenger has more assets in the end? This is now kind of saying first challenger won it as he has more assets.

contracts/contracts/RequestManager.sol
```diff
@@ -510,17 +510,42 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+    ///
+    /// In case the caller of this function is a challenger that won the game,
+    /// they will only get their staked native tokens plus the reward in the form
+    /// of full (sole challenger) or partial (multiple challengers) amount
+    /// of native tokens staked by the dishonest claimer.
+    ///
+    /// @param claimId The claim ID.
+    /// @param claimResolver The claim resolver.
+    /// @return The claim stakes receiver.
+    function withdraw(address claimResolver, uint96 claimId)
```
istankovic, line 536: I think the name `claimResolver` is quite confusing, especially when reading the code of `resolveClaim`. I would suggest using a different name that more directly indicates that this is the address that will get the stakes in case claim is invalid. E.g. `stakeRecipient` or something like that.

contracts/contracts/RequestManager.sol
istankovic: > Why are we multiplying with 2 here and saying it is partial withdraw?

Perhaps I can answer this. We support multiple challengers, i.e. anyone can join the challenge game:
```
(time flows to the right)
claimer -> ch1 -> claimer -> ch2 -> claimer -> ch1 -> claimer -> ch3 ...
```
Here each `chN` denotes a challenger. Claimer and challengers take turns in the game, but when the challengers get their turn, anyone can actually join in and make a challenge. In this case, we see `ch2` joining and challenging as a second challenger.

At the end of the game, each challenger has to call `withdraw` in order to get their stakes back, _plus_ the corresponding amount from the claimer's total stake. This is where both `partial` and `2 *` come from.
Partial, because each challenger only withdraws for themselves, i.e. they can only withdraw their part of the total stake. `2 *` because, in addition to their own stake that they put in during the game (`claim.challengersStakes[claimReceiver]`), they also get the same amount from the claimer's total stake. This is because, in order to win the game, the claimer has to match or exceed every challenger's stake.

Hope this helps.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -1037,6 +1037,50 @@ def test_challenge_after_l1_resolution(request_manager, token, claim_stake, cont
...
+    request_manager.challengeClaim(claim_id, {"from": first_challenger, "value": claim_stake + 1})
+    # Claimer outbids again
+    request_manager.challengeClaim(claim_id, {"from": claimer, "value": claim_stake + 1})
+    # Second challenger challenges
+    request_manager.challengeClaim(claim_id, {"from": second_challenger, "value": claim_stake + 1})
+
+    # Timetravel after claim period
+    chain.mine(timedelta=finality_period + challenge_period_extension)
+
+    request_manager.withdraw(second_challenger, claim_id, {"from": first_challenger})
```
istankovic, line 1064: This is fine, but I would (also) like to test the case where a non-participant to the game calls `withdraw`.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -1037,6 +1037,50 @@ def test_challenge_after_l1_resolution(request_manager, token, claim_stake, cont
...
+
+    # second challenger should have won claim stake which is the excess amount the
+    # claimer put in
+    assert (
+        web3.eth.get_balance(second_challenger.address)
+        == second_challenger_eth_balance + claim_stake
+    )
```
istankovic, line 1071: I would also assert the balance for the first challenger similarly.

contracts/contracts/RequestManager.sol
bilbeyt: thanks @istankovic

contracts/contracts/RequestManager.sol
fredo: I agree.

`stakeRecipient` is also not a good name, because resolveClaim will output the claimReceiver which will receive the stakes. In the case that the claimer actually wins, `stakeRecipient` can be set to any address but the claimer will receive the stakes. 
Do you have any other suggestion?

contracts/contracts/RequestManager.sol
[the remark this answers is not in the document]
fredo: > Should we maybe consider requiring that the claimReceiver is actually one of the challengers?
claimReceiver could also be the claimer. Or did you actually mean `claimResolver` here?

> claimResolver -> participant
good idea! 

> claimReceiver -> stakeReceiver/stakeRecipient
I like that!

beamer/tests/contracts/test_request_manager.py
fredo: good point

contracts/contracts/RequestManager.sol
istankovic: > `stakeRecipient` is also not a good name, because resolveClaim will output the claimReceiver which will receive the stakes. In the case that the claimer actually wins, `stakeRecipient` can be set to any address but the claimer will receive the stakes. Do you have any other suggestion?

Sadly, I can't think of anything better, at least for now.

Note that the confusion you describe arises only if you're acting dishonestly and calling the function.
In case you're honest, everything is still fine because
- if you as a claimer call this, you name yourself the state recipient because you're expecting to get your stake back
- if you as a challenger call this, similar to above, you name yourself the state recipient because you're expecting to get your stake back
- if you're anyone else (regardless of participation in the game), you name the stake recipient, and they will get their stake _if they are entitled to_

beamer/tests/contracts/test_request_manager.py
fredo: Actually both challengers won, because the challengers as a team won against the claimer. It is possible that multiple challengers join a challenge against a claimer. The distribution of the stakes will happen accordingly to the amount they put in. 
The claimer, in this case, won nothing.

[here](https://docs.beamerbridge.com/whitepaper.html#rightful-claims-resolutions) is a detailed description of how multiple  challengers join a challenge and how stakes are distributed. Please ask if oyu have any further questions :)

contracts/contracts/RequestManager.sol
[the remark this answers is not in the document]
istankovic: > claimReceiver could also be the claimer. Or did you actually mean `claimResolver` here?

I meant challenger in this case specifically because we're accessing `challengersStakes`, but the same question could be asked for `claimResolver` -- do we need to ensure that the... `participant` really is either the claimer or a challenger?

Personally, I would add that restriction.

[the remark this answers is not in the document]
fredo: > I meant challenger in this case specifically because we're accessing challengersStakes

we already do this exact check two lines below:

 `require(ethToTransfer > 0, "Challenger has nothing to withdraw");`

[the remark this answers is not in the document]
fredo: But I would go for an additional check on a higher level, namely require that participant is actually a participant of this game.

beamer/tests/contracts/test_request_manager.py
fredo: I don't know why this is relevant here, because I checked it a couple of lines below, but adding it here as well.

beamer/tests/contracts/test_request_manager.py
fredo: I tested this when withdrawing for the first challenger, so we test both cases.

contracts/contracts/RequestManager.sol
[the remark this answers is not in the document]
istankovic: > we already do this exact check two lines below:
> 
> `require(ethToTransfer > 0, "Challenger has nothing to withdraw");`

Ah, good point, somehow I missed that. Felt like we should check for that earlier.

fredo:
@istankovic I added this upper level check that participant should be either claimer or challenger which has not withdrawn its stakes yet. 

This required some changes in the tests. In a lot of tests we implicitly tested that withdraw for a claimer happens even if another address calls it. This doesnt work anymore with `withdraw(claimId)` as msg.sender is not an active participant. 
I could have either changed it in all places to `withdraw(claimer, claimId,{"from": other_address}`, but I decided to just test this in its own test.

contracts/contracts/RequestManager.sol
```diff
@@ -561,17 +561,47 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
         validClaimId(claimId)
         returns (address)
+    {
+        return withdraw(msg.sender, claimId);
+    }
+
+    /// Withdraw the deposit that the request submitter left with the contract,
+    /// as well as the staked native tokens associated with the claim.
+    ///
+    /// This function is called on behalf of participant. Only the participant
```
istankovic, line 576: Nitpick: should use indefinite article:
```
    /// This function is called on behalf of participant. Only the participant
```
```
    /// This function is called on behalf of a participant. Only a participant
```

istankovic:
[putting this here since github does not support comment on commit messages]

Commit  5823443d92a003cb5ee142244f9fff15ecc3fe76 has a typo in commit message:

`contracts: Add withdraw funciton on behalf of another address`

(`funciton` -> `function`)

istankovic (approved):
Looks good!

Issue #1254, opened 2022-12-06
Add sender to RequestCreated Event


fredo:
User facing services such as the graph node or the frontend could benefit from having the sender of the request in the Request Created Event. 
Especially if the address is indexed, it makes it easy to find all requests associated to a specific address. Currently you would have to query the whole history of events and iterate over them to find requests to a specific address. 

While trying to implement this feature, I ran into a `stack too deep error`, so we have to modify the variables in the function a bit.

PR #1261, opened 2022-12-12
Add sender to RequestCreated event

fredo:
implements: #1254

### DIFF
--- a/beamer/events.py
+++ b/beamer/events.py
@@ -72,6 +72,7 @@ class RequestCreated(RequestEvent):
     target_chain_id: ChainId
     source_token_address: ChecksumAddress
     target_token_address: ChecksumAddress
+    source_address: ChecksumAddress
     target_address: ChecksumAddress
     amount: TokenAmount
     nonce: Nonce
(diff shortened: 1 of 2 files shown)


contracts/contracts/RequestManager.sol
```diff
@@ -78,7 +78,8 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
         address targetAddress,
         uint256 amount,
         uint96 nonce,
-        uint32 validUntil
+        uint32 validUntil,
+        address indexed sender
```
istankovic, line 82: I'm thinking, since we already have `targetAddress`, shouldn't this be `sourceAddress`, to use terminology similar to the one already in the event?

fredo: perfect idea

beamer/events.py
```diff
@@ -72,6 +72,7 @@ class RequestCreated(RequestEvent):
     target_chain_id: ChainId
     source_token_address: ChecksumAddress
     target_token_address: ChecksumAddress
+    sender: ChecksumAddress
```
istankovic: This should be `source_address` after the latest change.

istankovic (approved):
One rename still needed, otherwise looks good!

PR #1265, opened 2022-12-13
relayer: refactor & integrate arbitrum

### LINKED ISSUE #687: Adapt message relayer for Arbitrum
We need to add support for Arbitrum to our relayer.
This means we need to:

- [x] refactor the existing relayer so that it hands over control to the chain-specific relayer code based on the L2 chain ID
- [x] implement a relayer using the Arbitrum SDK

The Arbitrum relayer needs to run roughly like this:
```
- check if the outbox entry exists; if it does not exist, exit
- while 1:
  - fee = arbitrum_l1_messenger.calculateRetryableSubmissionFee()
  - dfee = 1.1 * fee - arbitrum_l1_messenger.deposits[sender_address] 
  - call arbitrum_l1_messenger.deposit() with value dfee
  - l1_txhash = execute the outbox entry
  - if successful, break the loop

- wait for the ticket on L2
- redeem the ticket
```

### LINKED ISSUE #1277: Relayer: Implement Boba network relayer
# Rationale
While working on https://github.com/beamer-bridge/beamer/issues/687 i realized that we are lacking support for Boba network inside our relayer implementation.

Talked to the boba people and i got the following response: 
`We handle the tx from L1 to L2, but not from L2 to L1.`

This means that a relayer implementation in the direction `L1 -> L2 Boba`  is not needed but `L2 Boba -> L1` is necessary. 

### Why should this feature exist?
The relayer is one of the crucial components of our bridge. It is responsible for relaying the messages that are sent between the rollups. This happens when a challenge on one of the rollups needs to be relayed to L1/destination L2.
  
### What are the use-cases?
The relayer needs to support sending messages / challenge resolutions from L2 Boba via L1 to the destination L2.

# Acceptance criteria

- [x] Boba relayer service is implemented
- [ ] Relayer program is able to recognize boba as a supported network (mainnet & Goerli) & run the boba relayer service

GabrielBuragev:
closes #687 
closes #1277 

In order to support multi-rollup relayer support, the relayer logic had to be refactored.
With this code restructure we will be able to support relayer service implementations per rollup unlike the previous implementation, which was focusing on supporting Optimism only.

This PR contains the following work:
- Restructured relayer program to support multiple rollups as attached services
- Arbitrum relayer integrated
- Boba relayer integrated
- Optimism relayer improved by removing unnecessary dependencies & code
- Improved log messages
- Improved CLI documentation


GabrielBuragev:
So, i managed to fix the typescript issue. Now i just use `ethers: 5.7.2` instead of `ethers: ^5.7.2` inside package.json. Seems like that works.

Also rebased to latest main branch.

relayer/src/services/arbitrum.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  async relayTxToL1(l2TransactionHash: string): Promise<string | undefined> {
+    console.log("Arbitrum outbox execution");
+
+    /**
+   / * We start with a txn hash; we assume this is transaction that triggered an L2 to L1 Message on L2 (i.e., ArbSys.sendTxToL1)
+  */
+
+    if (!l2TransactionHash.startsWith("0x") || l2TransactionHash.trim().length != 66)
```
manuelwedler, line 78: This is only a small check but the same for all RelayerServices, so could probably be generalized in a way to not have code duplication.

relayer/src/types.ts
```diff
@@ -0,0 +1,48 @@
+import type { BaseRelayerService } from "./services/types";
```
manuelwedler, line 1: Hm the naming of this file sounds a bit odd to me. A lot of logic is happening here, so I would not say it is just declaring types. Can't we just call it `relayer-program.ts`?

manuelwedler (changes requested):
Looks mostly good to me. Nice splitting of the logic into the three steps with the abstract class. Also tried it out and works fine :+1: 
I have only some comments, mostly concerning code duplication, which would be nice to address

manuelwedler:
I am also considering if it makes more sense to use Vitest as a test runner. I think it makes sense to stick to the same tools inside one project. Vitest is also a pretty good tool in regards of performance and reliability from our experience so a far. Nothing against Jest, I think it would be just consistent to use Vitest. 
But this would also mean that we should use Vite for the build. Not sure if it adds more complexity for this simple relayer, could also give some benefits to have unified development tools. Overall, I would do it, as the configuration effort is minimal and it's a very performant tool.

relayer/package.json
```diff
@@ -18,17 +18,19 @@
...
+    "ethers": "^5.7.2",
     "pkg": "^5.8.0"
   },
   "devDependencies": {
+    "@typechain/ethers-v5": "9.0.0",
     "@types/node": "^16.11.39",
     "@typescript-eslint/eslint-plugin": "^5.30.4",
     "@typescript-eslint/parser": "^5.30.4",
     "eslint": "^8.19.0",
+    "prettier": "^2.8.1",
```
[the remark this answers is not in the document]
GabrielBuragev: Done ✅

relayer/src/service.ts
```diff
@@ -1,137 +1,47 @@
...
-      this.options.l1RpcProvider
-    );
+  const relayProgram = await RelayerProgram.createFromArgs(args);
 
-    this.state.messenger = new CrossChainMessenger({
-      l1SignerOrProvider: this.state.wallet,
-      l2SignerOrProvider: this.options.l2RpcProvider,
-      l1ChainId: await getChainId(this.state.wallet.provider),
-      l2ChainId: await getChainId(this.options.l2RpcProvider),
+  Promise.race([relayProgram.run(), killOnParentProcessChange(startPpid)])
```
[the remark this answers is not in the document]
GabrielBuragev: Good point ✅

manuelwedler:
Note: since Vite is mainly meant to be used by frontend projects and the backend plugins are badly maintained, I decided to not go with Vitest. So the addition of jest can stay in this PR.

manuelwedler:
This does not include a configuration for the CI yet. I created an issue for that: https://github.com/beamer-bridge/beamer/issues/1291

relayer/src/services/boba.ts
```diff
@@ -0,0 +1,95 @@
+import { CrossChainMessenger, MessageStatus, MessageReceiptStatus } from "@eth-optimism/sdk-1.0.2";
```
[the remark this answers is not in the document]
GabrielBuragev: Yes, but we have a problem due to the different versions of optimism sdk being used in these 2 files. 

In the `BobaRelayerService` we have to use `optimism/sdk@1.0.2` since this SDK version provides us with the correct API/types for communication with the boba network contracts. Boba haven't upgraded their network to use/fork the latest features implemented on Optimism therefore we cannot use the latest version of optimism SDK for their network.)

Unlike the boba relayer the `OptimismRelayerService` is using `optimism/sdk@latest` (obviously) which has the up-to-date API we need for communicating with the Optimism specific contracts responsible for the message bridging.

relayer/src/types.ts
GabrielBuragev: ✅

relayer/src/services/arbitrum.ts
GabrielBuragev: I moved the argument validations to its own function located in the `relayer-program.ts`. Thanks for catching this ✅

GabrielBuragev:
@istankovic your concern has been addressed! Thanks for reporting it.

manuelwedler (approved):
Everything got addressed. So we are good to go. Thank you :)

GabrielBuragev:
Thank you for the review Manu!

PR #1274, opened 2022-12-15
Add an `invalidate-fill` command to call_contracts

istankovic:


scripts/call_contracts.py
```diff
@@ -346,13 +346,11 @@ def whitelist(
...
-)
-@click.option("--transfer-limit", type=int, help="New transfer limit of token")
-@click.option("--min-lp-fee", type=int, help="New minimum lp fee")
-@click.option("lp-fee-ppm", type=int, help="New lp fee in ppm")
-@click.argument("protocol-fee-ppm", type=int, help="New protocol fee in ppm")
+@click.argument("token-address", type=str, metavar="ADDRESS", callback=validate_address)
+@click.option("--transfer-limit", type=int)
+@click.option("--min-lp-fee", type=int)
+@click.option("lp-fee-ppm", type=int)
+@click.argument("protocol-fee-ppm", type=int)
```
GabrielBuragev, line 353: What is the reason for removing the `help` arguments from the argument definitions?

scripts/call_contracts.py
```diff
@@ -371,7 +369,28 @@ def update_token(
     new_token_data = tuple(value or token_data[i] for i, value in enumerate(params))
     tx_receipt = transact(request_manager.functions.updateToken(token_address, new_token_data))
 
-    print(f"Transaction sent, tx_hash: {tx_receipt['hash']}")
+    print(f"Transaction sent, tx_hash: {tx_receipt['transactionHash'].hex()}")
+
+
+@cli.command("invalidate-fill")
+@click.argument("request-id", type=str, callback=validate_bytes)
```
GabrielBuragev, line 376: I just noticed that we are using type `str` here but everywhere else type `int` for `request-id`. 
I believe this is due to the change of the `request-id` type in the contract.
Looking at the current contract version the type should be `str` everywhere else. 

This change can be done either in this PR or another one since i don't know the amount of effort required for this change. I leave it to you to decide if you want to include them here or in another PR. 
I am fine with both approaches.

GabrielBuragev (approved):
One question and an update suggestion. Doesn't need to be a part of this PR.
Would be happy if they are addressed / an issue is created for the necessary type update.
But also fine if you include the update here.

For the python syntax, i would appreciate if @tolga can take a quick look as well since i am not a python expert.

scripts/call_contracts.py
istankovic: > What is the reason for removing the `help` arguments from the argument definitions?

Because `click.argument` does not have that argument (heh) and the script fails without even getting to run any of the commands. It simply was never tested with those changes. I did, however, remove the help for `click.option`, which does support `help`, which was wrong. I fixed this now so we only remove it for `click.argument`. Thanks for asking!

scripts/call_contracts.py
istankovic: > I just noticed that we are using type `str` here but everywhere else type `int` for `request-id`. I believe this is due to the change of the `request-id` type in the contract. Looking at the current contract version the type should be `str` everywhere else.

Exactly. This script has quite some technical debt and it shows. I noticed the same and decided to just do the small change here as was needed for my v1 checks. The other things can be fixed later.

istankovic:
>  Would be happy if they are addressed / an issue is created for the necessary type update.

Instead of just fixing the types, I would suggest we rethink the whole script approach, because the way it is implemented now is not really nice. For instance, it would be nice if the script would actually make use of ABI files to generate proper commands automatically. We could even use the Solidity doc comments for help text etc. The script could be made much more powrful, yet simpler, than it is now.

PR #1280, opened 2022-12-16
beamer: check tx receipt for status

bilbeyt:


scripts/call_contracts.py
```diff
@@ -101,7 +101,7 @@ def submit_request(
 
     request_manager = contracts["RequestManager"]
     token = contracts["MintableToken"]
-    total = amount + request_manager.functions.totalFee(amount).call()
+    total = amount + request_manager.functions.totalFee(token.address, amount).call()
```
istankovic, line 104: Yeah, this script really accumulated some technical debt, didn't it. Thanks for catching and fixing this!

beamer/util.py
```diff
@@ -67,7 +67,7 @@ def transact(
     receipt = func.web3.eth.wait_for_transaction_receipt(
         txn_hash, timeout=timeout, poll_latency=poll_latency
     )
-    if receipt.status == 0:
+    if receipt["status"] == 0:
```
istankovic, line 70: This change should be squashed into the first commit of the PR, it does not belong in this commit (which is about the test).

beamer/chain.py
```diff
@@ -318,7 +318,7 @@ def fill_request(request: Request, context: Context) -> None:
     log.info(
         "Filled request",
         request=request,
-        txn_hash=receipt.transactionHash.hex(),
+        txn_hash=receipt["transactionHash"].hex(),
```
istankovic, line 321: Why do we need this?

beamer/util.py
```diff
@@ -64,9 +64,12 @@ def transact(
     except (ContractLogicError, requests.exceptions.RequestException) as exc:
         raise TransactionFailed() from exc
 
-    return func.web3.eth.wait_for_transaction_receipt(
+    receipt = func.web3.eth.wait_for_transaction_receipt(
         txn_hash, timeout=timeout, poll_latency=poll_latency
     )
+    if receipt.status == 0:
+        raise TransactionFailed("Receipt status failed")
```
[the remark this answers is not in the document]
bilbeyt: TxReceipt has no such thing like error_message you can see example one that i created for test

beamer/chain.py
bilbeyt: TxReceipt is TypedDict and basically there is no attribute.

istankovic: > TxReceipt is TypedDict and basically there is no attribute.

So how did our code work until now?
For sure, the agent would crash if it was not there.

beamer/util.py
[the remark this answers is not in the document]
istankovic: > TxReceipt has no such thing like error_message you can see example one that i created for test

Ah, that's unfortunate. Then I would just use something like `"unknown error"` instead of the existing message,
because the existing message may give an impression that getting the status or receipt failed.

beamer/tests/agent/unit/utils.py
```diff
@@ -158,7 +158,13 @@ def make_context() -> Tuple[Context, Config]:
...
                     "timestamp": TIMESTAMP,
                 }
-            )
+            ),
+            TARGET_CHAIN_ID: BlockData(
+                {
+                    "number": BlockNumber(43),
+                    "timestamp": Timestamp(458),
+                }
+            ),
```
istankovic, line 167: Is `TARGET_CHAIN_ID` used anywhere?

beamer/tests/agent/unit/test_handlers.py
```diff
@@ -27,6 +28,27 @@
 from beamer.typing import FillId, Termination
 
 
+def get_tx_receipt(status, hashStr) -> TxReceipt:
```
istankovic, line 31: Python uses snake_case so `hashStr` should be renamed. Also, we tend to avoid putting types in names.
I think we should just name this variable `tx_hash`, like we have in dozens of other places.

beamer/tests/agent/unit/utils.py
bilbeyt: fill_request is using it that is why i added another key for TARGET_CHAIN_ID

beamer/chain.py
bilbeyt: i mean i just checked the wait_for_transaction_receipt also with mypy and the return type was TxReceipt which is typed dict. then maybe it wasnt actually returning TxReceipt ?

istankovic:
Commit `47fdbedb07af9bb79ae5af55b3b35c94e531a453` has unclear message:
```
 fix: add total for submit_request script 
```

I'm not sure what "add total" means here, and the script is called `call_contracts`, not `submit_request`.
(`submit_request` is the name of the command)

beamer/chain.py
istankovic: > i mean i just checked the wait_for_transaction_receipt also with mypy and the return type was TxReceipt which is typed dict. then maybe it wasnt actually returning TxReceipt ?

I don't know, but we should no be changing that for no reason. Could be that your mock tx receipt is not really a typed dict?

bilbeyt: Before this i just had a dataclass called Receipt and status and transactionHash was its fields and that way code was working but linter said no then i created this change

bilbeyt: anyhow, this is the running version of type of result web3.datastructures.AttributeDict

bilbeyt: will use that should work

istankovic: > web3.datastructures.AttributeDict

Yeah, I think that's the missing part in your mock impl of tx receipt.
Once you fix that, none of the changes in `chain.py` are needed and you should also move the `test_handlers.py` changes from this commit to the earlier one. Which means this commit can be dropped completely.

beamer/tests/agent/unit/test_handlers.py
```diff
@@ -2,10 +2,11 @@
 from unittest.mock import patch
 
 import pytest
+from eth_typing import BlockNumber, HexStr
 from hexbytes import HexBytes
-from web3.types import ChecksumAddress, Wei
+from web3.types import ChecksumAddress, TxReceipt, Wei
 
-from beamer.chain import claim_request, process_claims, process_requests
+from beamer.chain import _ERC20_ABI, claim_request, fill_request, process_claims, process_requests
```
istankovic, line 9: Please don't import private things: `_ERC20_ABI`

beamer/tests/agent/unit/test_handlers.py
```diff
@@ -25,6 +26,28 @@
...
+            "effectiveGasPrice": 1,
+            "gasUsed": Wei(1),
+            "from": ADDRESS1,
+            "logs": [],
+            "logsBloom": HexBytes(0),
+            "root": HexStr("1"),
+            "to": ADDRESS1,
+            "transactionIndex": 1,
+        }
+    return cast(TxReceipt, AttributeDict(data))
```
istankovic, line 50: I assume `AttributeDict` supports kwargs so instead of first creating a dict and then copying it to make another one, why not just create `AttributeDict` directly, with kwargs?

bilbeyt: from is making the problem :(

bilbeyt: and AttributeDict needs dict so kwargs are not working.

istankovic: :woozy_face:

beamer/tests/agent/unit/test_handlers.py
```diff
@@ -25,6 +28,30 @@
...
+def get_tx_receipt(status, tx_hash) -> TxReceipt:
+    data = {
+        "status": status,
+        "transactionHash": tx_hash,
+        "blockHash": HexBytes(0),
+        "blockNumber": BlockNumber(1),
+        "contractAddress": None,
+        "cumulativeGasUsed": 0,
+        "effectiveGasPrice": 1,
+        "gasUsed": Wei(1),
```
istankovic, line 46: Why is `gasUsed` in Wei?

istankovic (approved):

PR #1287, opened 2022-12-19
Fix recent CI breakage

istankovic:


.circleci/workflows.yml
```diff
@@ -83,28 +90,29 @@ jobs:
...
 
       - restore_cache:
           keys:
-            - beamer-dependencies-v4-{{ checksum "poetry.lock" }}
-            - beamer-dependencies-v4
-      - run: poetry install
-      - run: echo "source $(poetry env info -p)/bin/activate" >> ~/.bashrc
+            - beamer-dependencies-v5-{{ checksum "poetry.lock" }}
+            - beamer-dependencies-v5
+      - run: poetry install -vvv
```
GabrielBuragev, line 104: Do we still need the verbose logging (-vvv) here?

.circleci/workflows.yml
```diff
@@ -83,28 +90,29 @@ jobs:
...
+      - run: grep -q activate ~/.bashrc || echo "source $(poetry env info -p)/bin/activate" >> ~/.bashrc
```
GabrielBuragev, line 105: Could we be more explicit here and check against `/bin/activate` for example?

.circleci/workflows.yml
```diff
@@ -18,16 +18,21 @@ commands:
...
       # Restore node runtime and packages at /opt/circleci/.nvm
       - restore_cache:
           name: Restore node runtime and packages
-          key: node18-packages-v17
-      - run: echo "nvm use default" >> "${BASH_ENV}"
+          key: node18-packages-v21
+
+  ensure-node-environment:
+    steps:
+      - run: if grep -q "nvm.sh" "${BASH_ENV}"; then circleci-agent step halt; fi
```
GabrielBuragev, line 33: `circleci-agent step halt` nice solution!

GabrielBuragev (changes requested):
Well done. I like the tricks you used there to solve the issues!

One question and one minor change request.

.circleci/workflows.yml
istankovic: We don't strictly need it now, but I left it in case something goes wrong in the future with poetry, we have the whole log right away, without having to rerun the workflow again.

But if you think we should just remove it, I will remove it.

.circleci/workflows.yml
istankovic: Done.

.circleci/workflows.yml
GabrielBuragev: Nah, just leave it. We can remove it in the futrure.

GabrielBuragev (approved):
Awesome ✅

Issue #1226, opened 2022-11-25
Implement Matomo individually for dapp


andfletcher:
# Description
The current tracking setup doesn't allow accurate and clean tracking because they're aggregated and the app is merely a subdomain. This does not allow to separate people that convert from website to app from people that directly land on the app, for example, making it harder to understand the user journey.

To Do
- [ ] implement code on the beamer app website so that we can track it individually from the rest of our pages
- [ ] Ping Dina that is has been done

```<!-- Matomo -->
<script>
  var _paq = window._paq = window._paq || [];
  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
  _paq.push(['trackPageView']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="https://beamerbridge.matomo.cloud/";
    _paq.push(['setTrackerUrl', u+'matomo.php']);
    _paq.push(['setSiteId', '3']);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.async=true; g.src='//cdn.matomo.cloud/beamerbridge.matomo.cloud/matomo.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
<!-- End Matomo Code -->´´´

PR #1293, opened 2022-12-21
frontend: modify matomo code

compojoom:
fixes: https://github.com/beamer-bridge/beamer/issues/1226


GabrielBuragev (approved):
Looks good. Thanks for closing this! 

Do we need this on the current mainnet deployment (have it live asap)? 
That would mean we need to cherry-pick this commit to the 0.x branch & redeploy the frontend.
If there is no urge for this, we can wait until our next deployment (v1).

Issue #1307, opened 2022-12-23
Agent needs to filter events for source and target chains


bilbeyt:
Currently with v1 contracts deployment, we found in tests, agent is trying to fill request on chains that is not configured for. 

We had the following example:
Agent is configured to run for `boba <-> arbitrum` but it tries to fill request for optimism which is raising an error and makes the container stop & restart.

Just before this error raise, there is a log like this:

```
2022-12-23 14:10:28.729806 [debug    ] New events                     [EventProcessor] events=[FinalityPeriodUpdated(chain_id=2888, block_number=955, tx_hash=HexBytes('0x563b3ae40d0419745a02ce330334af7c259901ea837a4b6b58432e9e696e38c0'), target_chain_id=421613, finality_period=18000), FinalityPeriodUpdated(chain_id=2888, block_number=956, tx_hash=HexBytes('0xcb518aab4f5c9454786cfd10aaf8d1e248b2b72dc3226df84f779b1106eb3918'), target_chain_id=420, finality_period=1800), RequestCreated(chain_id=2888, block_number=1062, tx_hash=HexBytes('0x2aa00fa706d98e7a78f7d5f2586799447ab5adfa59f18a9b74e79a844e0105d3'), request_id=RequestId('0x95e58bf24fea77e30dba5d0b1e397ad33c6d6e3c535abb391f0b5c341a12c231'), target_chain_id=420, source_token_address='0x8f4aA3BD1eA3De8BeAAe075Bd48026cf38B33b1a', target_token_address='0x66c538Ba7402C37eEaFd1Df5026FF23431B9d653', source_address='0xf761D67597Bcc6021BE660414a001ee88e54511a', target_address='0xf761D67597Bcc6021BE660414a001ee88e54511a', amount=1, nonce=1, valid_until=1671728319), <LatestBlockUpdatedEvent chain_id=2888 block_number=1085 hash=0x2e3772b898e54462bce4963a853ca9f4bbbb7b0abbf5e50a771ce78011caef03>]
```

and this is the error: 

```
Traceback (most recent call last):
  File "/src/beamer/chain.py", line 36, in wrapper
    return func(*args, **kwargs)
  File "/src/beamer/chain.py", line 162, in _thread_func
    process_requests(self._context)
  File "/src/beamer/chain.py", line 220, in process_requests
    fill_request(request, context)
  File "/src/beamer/chain.py", line 283, in fill_request
    block = context.latest_blocks[request.target_chain_id]
[...]

PR #1308, opened 2022-12-23
agent: fix listening wrong target chain request

bilbeyt:


fredo (dismissed):
I would prefer, if the requestCreated events would already be discarded instead of removing it during process request.

events are handled by the event handlers for each event.
After all events are processed, process_request and process_claims are called to generate an output, namely the actions to take on the current state.
A request which is not of interest is already known when the event gets consumed and processed and can actually be discarded.

btw we already have a module for ,this, the TokenMatchChecker. Maybe this would be a good place to include this check.

bilbeyt:
> I would prefer, if the requestCreated events would already be discarded instead of removing it during process request.
> 
> events are handled by the event handlers for each event. After all events are processed, process_request and process_claims are called to generate an output, namely the actions to take on the current state. A request which is not of interest is already known when the event gets consumed and processed and can actually be discarded.
> 
> btw we already have a module for ,this, the TokenMatchChecker. Maybe this would be a good place to include this check.

TokenMatchChecker -> We are adding all supported tokens in all rollups for agent, but an agent is actually working on only source and target chain. That was the cause of this problem. An agent working for boba -> op was also getting logs related arbitrum lets say and then 

```
context.latest_blocks[request.target_chain_id]
```

was raising. Hence, i dont think TokenMatchChecker is a good place as it doesnt know which chains an agent is running.

bilbeyt:
I will move this to event handler of RequestCreated.

bilbeyt:
never mind, i used is_valid_pair now :)

beamer/util.py
```diff
@@ -176,10 +176,14 @@ def is_valid_pair(
         source_token_address: ChecksumAddress,
         target_chain_id: ChainId,
         target_token_address: ChecksumAddress,
+        current_target_chain_id: ChainId,
```
fredo, line 179: During the whole runtime, `current_target_chain_id` will be a constant value. I'd rather add this to the TokenMatchChecker once (during init) instead of passing to the function everytime we call it. Seems a bit inefficient.

beamer/util.py
```diff
@@ -176,10 +176,14 @@ def is_valid_pair(
...
     ) -> bool:
         source_token = source_chain_id, source_token_address
         target_token = target_chain_id, target_token_address
-        return target_token in self._tokens.get(source_token, frozenset())
+        return (
+            target_token in self._tokens.get(source_token, frozenset())
+            and current_target_chain_id == target_chain_id
```
fredo, line 185: Shouldn't the test also include a check against the source chain ID?
What happens if the agent is configured for (Op -> Boba) but the source chain is Arbitrum? 
In this case it would return true wouldnt it?

fredo:
Thinking about this problem, we could probably get rid of all of this by filtering the TokenMatchChecker during init for the configured chain ids right? 
I think the agent should only be concerned about the chain ids which it is running on, so it could discard all other chain ids, and this should already resolve the problem, shouldnt it? 
I'd point you preciesly to where the `equiv_class` is created in `TokenMatchChecker.init()`

beamer/util.py
```diff
@@ -147,16 +147,22 @@ def make_web3(
...
         self._tokens: dict[_Token, frozenset[_Token]] = {}
 
+        running_chain_ids = [source_chain_id, target_chain_id]
+
         for token_mapping in tokens:
             equiv_class = frozenset(
                 (ChainId(int(token[0])), to_checksum_address(token[1])) for token in token_mapping
             )
-            l2_chain_ids = frozenset(chain_id for chain_id, _ in equiv_class)
+            l2_chain_ids = frozenset(
```
fredo, line 163: I think the check for l2_chain_ids is not enough, but if we pull the filtering out as suggested above the TokenMatchChecker should just work out of the box and wouldnt require any changes

istankovic:
@bilbeyt is this related to #1307?

beamer/tests/agent/unit/utils.py
```diff
@@ -148,6 +148,8 @@ def make_context() -> Tuple[Context, Config]:
     context = Context(
         requests=Tracker(),
         claims=Tracker(),
+        source_chain_id=SOURCE_CHAIN_ID,
+        target_chain_id=TARGET_CHAIN_ID,
```
istankovic, line 152: Why do we need this in `Context`?

beamer/state_machine.py
```diff
@@ -44,6 +44,8 @@
 class Context:
     requests: Tracker[RequestId, Request]
     claims: Tracker[ClaimId, Claim]
+    source_chain_id: ChainId
+    target_chain_id: ChainId
```
istankovic, line 48: Why do we need this in `Context`?

bilbeyt:
> @bilbeyt is this related to #1307?

yes

beamer/tests/agent/unit/utils.py
fredo: We don't need it in the current state. But we should also consider other events which are irrelevant for the agent.
For example, `RequestFilled` event is irrelevant for the agent if the source chain id doesnt match the agent's source chain id.
I haven't thought through all events for now.

fredo: By comparing the source chain id of the RequestFilled event with the agent's source chain id we know that the agent is not interested in this event

beamer/state_machine.py
```diff
@@ -155,6 +157,7 @@ def _handle_latest_block_updated(
 def _handle_request_created(event: RequestCreated, context: Context) -> HandlerResult:
     # If `BEAMER_ALLOW_UNLISTED_PAIRS` is set, ignore the token configuration
     if os.environ.get("BEAMER_ALLOW_UNLISTED_PAIRS") is not None:
+        print(os.environ.get("BEAMER_ALLOW_UNLISTED_PAIRS"))
```
istankovic, line 160: Stray `print`.

beamer/config.py
```diff
@@ -124,8 +125,16 @@ def load(config_path: Path, options: dict[str, Any]) -> Config:
...
+    target_chain_id = ChainId(target_w3.eth.chain_id)
+
     deployment_info = load_deployment_info(Path(config["deployment-dir"]))
-    token_match_checker = TokenMatchChecker(list(config["tokens"].values()))
+    token_match_checker = TokenMatchChecker(
+        list(config["tokens"].values()), source_chain_id, target_chain_id
+    )
     return Config(
         account=account,
         deployment_info=deployment_info,
```
istankovic, line 140: I am going to comment here, but it goes for the whole commit/PR: I don't think it is a good idea to do this check via `TokenMatchChecker`.

Firstly, as can be seen above, we have network traffic during configuration load. It also mixes mechanism (loading of configuration) with policy (filtering). We should not do that.

Secondly, I feel the current token match checker is simply not a good fit for the new filter. Maybe we should consider renaming it? The issues mentions "listening wrong target chain request", which leads me to think we should rather consider a different way to reject requests that we are not interesting in. Personally, I would rather see an `if` check somewhere than push this check into token match checker as is. (This only under the assumption that there are not many places for such a check, naturally.)

fredo: Thanks for the comment.
In my view, we should revise or refactor respectively, the TokenMatchChecker into a module which filters for a certain set of conditions.
We have the issue that with adding a third rollup, we increased one dimension of potential filters. Currently we only filtered if events are incorrect semantically (wrong addresses/tokenpairs, etc). 
Now we have the fact that we created a new dimension of "valid" events but not valid for this particular agent instance.
This creates the need of filtering out events which are not of interest for the agent. For each event and therefore each event handler, we need to check whether we should filter this event because it is irrelevant for the agent.

If we dont do this then two things can happen:
a) the agent eventually crashes (this is what we saw and why this issue is here)
b) the event starves in the event loop.

We are lucky that most Claim related events will be filtered out naturally by the fact that we dont create a request object. But for the future, we need to have a more systematic way of filtering unwanted events as we may introduce new events by the contracts.

IMO the TokenMatchChecker is doomed to be replaced by something much more powerful, a generalized module which checks if this event is ok to be processed.

The real question you are raising here (besides the technical opinions like having network traffic during configuration load) is if we should assume that a token pair is valid if it belongs two a chain which the agent doesn't know or should it raise a invalid token pair error. In the past, a chain id other than the two the agent is running on would probably have failed because the agent doesn't know about this other chain.

Since we are now using one agent config for all agent instances we have more than 2 chains per token in the config which are considered valid. The TokenMatchChecker does not catch this issue. IMO it is the purpose of the TokenMatchChecker of each agent instance to check the `(chain_id, token_address)` tuple if this is a valid token pair. Now, we should ask ourselves if it is a valid token pair even though the agent is not running and therefore doesnt know anything about this chain. In the past the TokenMatchChecker would have noticed this and filtered out the event. 

In general, I think it is totally fine and it's the purpose of the TokenMatchChecker to detect a request which is not a valid pair for this particular agent instance. 

Nonetheless, for a long time I want to rework the way we validate events if they are of interest for the agent instance, because I also think having tons of if statements here and there in the event handlers is also not a good design for the future. 
As a result I don't see this as a change for long as I think, as mentioned, that we need to replace the TokenMatchChecker with a more generalized way of analyzing events.

fredo: hmm let's look from a different angle on this issue, maybe this is a better approach.

Currently, an agent runs on two roll ups, A and B. So the agent is only concerned about information of A and B. And I strongly think that the agent should only consume information about these roll ups. 
Now, we started to create a config file which also contains information of roll up C. We created - let's call it - a multi instance config.
We now have the conflict that each agent instance will consume information of rollups, it doesnt know. The reason why I think we should avoid storing information of roll up C in the A->B agent instance is that the agent cannot possibly know even the existence of this roll up.
And as we have seen, things break if we do this. So by creating a multi instance agent config, I see the need of filtering necessary information for each of the instances. IOW, Agent A->B should only care and consume/store the information of A and B and not of C.

This is basically what is done in the TokenMatchChecker by dropping the token information of roll up C. This could also be done on a higher level but I would definetly not want to have information of roll up C inside of A->B.
By filtering out the token information of roll up C, it will naturally drop any request which is not A->B. Since I would recommend to do this anyway, there is no need of a specific if statement inside of the RequestCreated handler. The filtering comes for free.

I'm open for suggestions of a different way how to filter the config for A->B information, if you think this isn't the right place.

All this being said, it shows that this change is not here for long as we will collapse all agent instances into one agent process. With the planned changes of our agent software, the filtering process will need to be implemented properly in order to build the foundation for a scalable agent software.

istankovic: Thank you for elaborating on the background for these changes, I really appreciate it. And I do see where you are coming from and agree for the most part. And yes, you are fully on-point regarding the multi-chain instances/config.

My concern here is rather technical and quite specific. After having some more look into the code, I think
removing the unsupported tokens (pairs of `chain_id, token_address`) from the loaded config just after
```
config = beamer.config.load(config_path, options)
```
in `beamer.cli.main` would be the best way to go, given circumstances. The token match checker would then stay as-is, and only the `cli` module which is tasked with setting up the concrete agent instance can do this fiddling, query chain ID and remove those tokens we do not need. I think that'd be a much cleaner solution for now. (And of course, fully agreed that this will change in the future.)

What do you think?

fredo: Sounds like a good idea. Let me change it.

istankovic (approved):
Looks good!

PR #1309, opened 2022-12-27
agent check time request expiry

bilbeyt:


beamer/chain.py
```diff
@@ -281,6 +281,11 @@ def process_claims(context: Context) -> None:
 
 def fill_request(request: Request, context: Context) -> None:
     block = context.latest_blocks[request.target_chain_id]
+    time_until_expiry = request.valid_until - block["timestamp"]
```
fredo, line 284: I think `time.time()` instead of `block["timestamp"] would be more appropriate and consistent here. 
the timestamp of the latest block is likely to be in the past. This means that the actual expiry time is less than the calculated one.

beamer/chain.py
```diff
@@ -281,6 +281,11 @@ def process_claims(context: Context) -> None:
...
+    if time_until_expiry > context.config.expiry_wait_time:
+        log.info("Request is not ready for process", request=request)
```
istankovic, line 286: Could we have a more informative message here?

beamer/cli.py
```diff
@@ -51,6 +51,11 @@ def _sigint_handler(agent: Agent) -> None:
     type=int,
     help="Time in seconds to wait for a fill event before challenging a false claim.",
 )
+@click.option(
+    "--expiry-wait-time",
+    type=int,
+    help="Time left until expiry in seconds to fill a request.",
```
istankovic, line 57: I find the `expiry-wait-time` naming confusing and would prefer if we came up with a clearer name.
The way I was thinking about this:
```
|---------|--------------------------------|
A    T1   B                 T2             C
```
where 
```
A = request creation
B = earliest moment an agent can fill the request
C = request expiration
```

If we were interested in period `T1`, then it would be easier because that's easier to express. E.g. `fill-age` i.e. the age at which the request can be filled by the agent. However, I don't have a nice naming yet for `T2`.

istankovic (changes requested):
I think we need to come up with a better name for the option.
Also, we lack documentation for the new option.

beamer/cli.py
fredo: Here was some missunderstanding going on. The value describes the time left, a request must be still valid. If the a request es valid for less then the added value then the agent will ignore the request. So according to your drawing: 



```
|---------|--------------------------------|
A    T1   B                 T2             C
```
where 
```
A = request creation
B = last moment an agent is willing to fill the request
C = request expiration
```

bilbeyt:
About the option name, we are considering unsafe-fill-time. Is it ok ?

beamer/cli.py
bilbeyt: unsafe-fill-time @fredo @istankovic ?

istankovic: I think `unsafe-fill-time` is definitely an improvement over the current name.

beamer/agent.py
```diff
@@ -51,6 +51,10 @@ def __init__(self, config: Config):
         request_manager = l2a_contracts["RequestManager"]
         fill_manager = l2b_contracts["FillManager"]
 
+        max_validity_period = request_manager.functions.MAX_VALIDITY_PERIOD().call()
+
+        if config.unsafe_fill_time > max_validity_period:
+            raise RuntimeError(f"Unsafe fill time should not be bigger than {max_validity_period}")
```
istankovic, line 57: We should probably use `>=` instead of `>` because if these two times are equal, the agent will not fill any requests.

I would also suggest replacing `should not` with `must not` in the exception message.

beamer/tests/agent/test_request.py
```diff
@@ -302,3 +302,23 @@ def test_agent_ignores_invalid_fill(_, request_manager, token, agent: Agent):
...
             sleeper.sleep(0.1)
+
+
+def test_unsafe_fill_time(
+    request_manager, config, token, set_allow_unlisted_pairs
+):  # pylint:disable=unused-argument
+    requester, target = alloc_accounts(2)
+    request_id = make_request(request_manager, token, requester, target, 1)
+    # 1 second to fill the request
+    config.unsafe_fill_time = 1799
```
istankovic, line 313: Nitpick: might be better to compute the value based on the contract's max validity time.

docs/source/agent.rst
```diff
@@ -127,3 +127,23 @@ game is finished and one participant withdraws the stake, the claim will move in
...
 If an agent is not participating in a claim, that claim will transition into the ``ignored`` state.
+
+.. _Unsafe Fill Time:
+
+Unsafe Fill Time
+----------------
+
+Agent cli command has an option ``--unsafe-fill-time`` to configure. With this option, agent can 
+decide which requests to ignore looking at request validity period. This option added to lower the risk 
+of filling a request which is too close to expire.
```
istankovic, line 138: I would reformulate this to read a bit better, e.g.

```
In order to lower the risk of filling a request that is too close to expiration, an agent has a notion of 
_unsafe fill time_. This is the time window, expressed in seconds, before request expiration that the agent
considers unsafe for doing a fill. Once an unfilled request enters the unsafe fill time, the agent will
simply ignore the request.

This time can be configured via the ``--unsafe-fill-time`` command-line option and the configuration file
option ``unsafe-fill-time``.
```

docs/source/unsafe_fill_time.dot
```diff
@@ -0,0 +1,13 @@
...
+    "start" [xlabel=T1];
+    "unsafe_start" [xlabel=T2]
+    "end" [xlabel=T3]
+    edge[arrowhead="none"];
+    {rank = same; start; unsafe_start}
+    {rank = same; unsafe_start; end}
+
+    start -> unsafe_start [label = "safe", minlen=6]
+    unsafe_start -> end [label = "unsafe", minlen=4]
+}
```
istankovic, line 13: Nice graph, thanks for doing this! :heart:

docs/source/configuration.rst
```diff
@@ -147,6 +148,14 @@ Reference
      - Time in seconds to wait for a fill event before challenging a false claim.
        Default: ``120``.
 
+   * - ``--unsafe-fill-time TIME``
+     - ::
+
+        unsafe-fill-time = TIME
+
+     - Time in seconds to ignore requests according to closeness of request expiry.
```
istankovic, line 156: I would reformulate this a bit, e.g.
```
     - Time in seconds before request expiry, during which the agent will consider it
       unsafe to fill and ignore the request.
```

docs/source/agent.rst
```diff
@@ -127,3 +127,23 @@ game is finished and one participant withdraws the stake, the claim will move in
...
+
+.. graphviz:: unsafe_fill_time.dot
+   :align: center
+   :caption: Unsafe fill time
+
+In the graph, T1 is representing the request creation time, T2 is calculated using ``unsafe-fill-time`` option, and T3 is the request expiration.
+
+If unsafe time increases, T2 moves to left, so agent can process more requests.
```
istankovic, line 146: I think this should be
```
If unsafe time increases, T2 moves to left, so the agent will fill fewer requests.
```

(The safe window decreases so we can fill fewer requests.)

docs/source/agent.rst
```diff
@@ -127,3 +127,23 @@ game is finished and one participant withdraws the stake, the claim will move in
...
+
+If unsafe time decreases, T2 moves to right, so agent can prevent filling requests which are nearly expired.
```
istankovic, line 148: Similarly to above, I think this should be

```
If unsafe time decreases, T2 moves to right, so the agent will get an opportunity to fill more requests.
```

istankovic (changes requested):
Looks good! I left a few comments regarding the docs etc. It's especially useful to have the graph!

beamer/tests/agent/test_request.py
bilbeyt: then agent just raise runtime error ?

istankovic: > then agent just raise runtime error ?

I'm not sure what you mean, but I meant something like
```
config.unsafe_fill_time = RequestManager.MAX_VALIDITY_PERIOD - 1
```

bilbeyt: ok now got it :)

beamer/cli.py
```diff
@@ -51,6 +51,11 @@ def _sigint_handler(agent: Agent) -> None:
     type=int,
     help="Time in seconds to wait for a fill event before challenging a false claim.",
 )
+@click.option(
+    "--unsafe-fill-time",
+    type=int,
+    help="Time in seconds to ignore requests according to closeness of request expiry",
```
istankovic, line 57: Maybe we should reuse the line from docs here for the help string, it's not much longer:
```Time in seconds before request expiry, during which the agent will consider it unsafe to fill and ignore the request.
```

beamer/agent.py
```diff
@@ -51,6 +51,10 @@ def __init__(self, config: Config):
         request_manager = l2a_contracts["RequestManager"]
         fill_manager = l2b_contracts["FillManager"]
 
+        max_validity_period = request_manager.functions.MAX_VALIDITY_PERIOD().call()
+
+        if config.unsafe_fill_time >= max_validity_period:
+            raise RuntimeError(f"Unsafe fill time must not be bigger than {max_validity_period}")
```
istankovic, line 57: Nitpick: slight better would be "must be less than" instead of "must not be bigger than" because we're also disallowing them being equal.

istankovic (approved):
Looks good, thanks!

Issue #1319, opened 2023-01-06
Agent: Requests should be claimed only once


GabrielBuragev:
#### Expected behavior
Agents should claim (call `claimRequest` method) the requests only once.

#### Actual behavior
Agents are claiming (calling `claimRequest` method) the same request multiple times.

#### Steps to reproduce the behavior
- Create a request on any rollup
- Agent claims the request once after filling it on destination rollup
- Restart the agent
- Now, the agent claims the same request again

fredo:
can you reproduce this behavior on Goerli with the latest Agent version (the one we run from source currently) between Optimism and Arbitrum (excluding Boba)?

GabrielBuragev:
Yeah, you can see the last `Claim Request` transactions here: https://goerli-optimism.etherscan.io/address/0x9d3F5EE65BF588C89d690272D0984774Fb77590A

istankovic:
We should have a test for this in our test suite.

PR #1321, opened 2023-01-06
Request state change upon own ClaimMade event

fredo:
The request object needs to be transitioned to state claimed once a ClaimMade event emitted by the agent happens. 
Note that the state claimed is only relevant for request objects which are filled by the agent.

If a request is filled by somebody else then we may not proceed to the claimed state. 

fixes: #1319

### DIFF
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -31,19 +31,22 @@ class Agent:
         self._config = config
         self._stopped = threading.Event()
         self._stopped.set()
+        self._init()
+
+    def _init(self) -> None:
         # Just add one worker, as this effectively serializes the work
         # This is necessary as we use the account for all resolutions and
         # would otherwise run into nonce problems
         self._task_pool = ThreadPoolExecutor(max_workers=1)
 
-        w3_l1 = make_web3(config.l1_rpc_url, config.account)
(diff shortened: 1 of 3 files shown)


GabrielBuragev:
I guess this fixes https://github.com/beamer-bridge/beamer/issues/1319 ?

fredo:
yes, but it's not ready to review. We should fix circle ci before piling up the stack more and more

bilbeyt (approved):
Seems good!

beamer/agent.py
```diff
@@ -115,6 +118,7 @@ def stop(self) -> None:
         self._event_monitor_l2b.stop()
         self._task_pool.shutdown(wait=True, cancel_futures=False)
         self._stopped.set()
+        self._init()
```
istankovic, line 121: I would put the call to `_init` one line above, before `self._stopped.set()` so that whoever calls `wait()` from a different thread does not get to run before we reinitialize the agent instance.

istankovic (approved):
It is incredible that we haven't been updating the request state upon seeing claims.
Amazing that things worked this far. Good find!

Also, that's a very nice test. :heart:

PR #1322, opened 2023-01-06
Downgrade poetry to 1.2.2 on circle ci

fredo:
Poetry 1.3.3 does not seem to work on CircleCI. It fails upon poetry install.

Until we have found a solution ´, we have to downgrade poetry to the last known working version


fredo:
closed in favor of #1294

PR #1327, opened 2023-01-09
deployments: add new mainnet deployment

GabrielBuragev:


fredo:
looks good

Note, that we now move to a new contracts version on mainnet

fredo (approved):

PR #1328, opened 2023-01-09
Frontend: add multi token support

### LINKED ISSUE #1222: Frontend: Adapt fee usage per new contract design
# Rationale
Based on the changes introduced (agreed upon) in https://github.com/beamer-bridge/beamer/issues/941 we need to adapt the front-end code respectively.

From the issue:
> We are gonna have a mapping on a per token basis:
	    values: protocolFee, lpFee, minLpFee, transferLimit

At this moment, the front-end consumes these properties from the contract by fetching them from the primitive properties of the contract. This is because contract v0.x is designed to support only 1 token (USDC ATM).
With the changes introduced in #941 we are supporting multiple tokens defined as structs which are stored in a mapping. 
We need to adapt the frontend codebase so that it fetches the Token definition object from the contract and uses that internally for representing/calculating the fees / transfer limits etc.

# Acceptance Criteria
- [ ] Front-end is fetching the Token definition object from the contract
- [ ] Property types are adjusted (for those that have types different then `uint256`) - check if we need to use numbers or uint256
- [ ] Fees and other derived values from this Token definition are properly reflected in the views

GabrielBuragev:
Fixes #1222 

This PR adds support for the new multi-token design of our contracts.
It touches all the functions that were previously responsible for fetching the token attribute values from the contract. 

Newest ABI files had to be pulled in for proper typing support which led to easier refactoring.

All the corresponding tests were adjusted as well.


compojoom (approved):
Looks fine to me.

Issue #1183, opened 2022-11-09
frontend: MetaMask shows errors for provided chain data


manuelwedler:
When a user does not have a network and the dapp triggers to add the network, MetaMask shows an error and a warning that the chain data might not be correct. The issue is that MetaMask checks against this list: https://github.com/ethereum-lists/chains

We should make sure to use the same network name, currency symbol, etc. as in the repo. It might make sense to just fetch it from the list when we generate the configuration files.

#### Expected behavior
No errors / warnings in MetaMask.

#### Actual behavior
![image](https://user-images.githubusercontent.com/34456797/200811139-9f02599a-4ca7-4bf8-a545-5ac955b4dc18.png)

#### Steps to reproduce the behavior
(Remove network in case you already have it)
Select a source chain and let the dapp trigger the chain switch.

GabrielBuragev:
Good idea, lets do this!

andfletcher:
@GabrielBuragev to create followup issue to address the actual problem for future rollups-

PR #1330, opened 2023-01-10
Frontend/arbitrum

### LINKED ISSUE #1278: Frontend: Add arbitrum mainnet support
# Rationale
Once we deploy our contracts on Arbitrum we need to include the metadata files in the frontend. These metadata files a being picked up by the pre-install config script and are being merged with the contract addresses found inside the `deployments` folder.

# Acceptance criteria
- [x] Chain metadata file is created
- [x] Contract ABI's are up to date (copied inside `src/assets` if needed)
- [x] On `yarn install` the `config.production.json` file includes the arbitrum configuration object
- [x] Arbitrum is visible inside the chain selectors on the app
- [x] By clicking on Arbitrum selector option, wallet provider tries to connect to Arbitrum network (if not found tries to import it)

compojoom:
fixes #1278 and #1183


GabrielBuragev:
I guess this fixes the bug reported in https://github.com/beamer-bridge/beamer/pull/1330 but it doesn't implement it by following the suggested solution there:
> The issue is that MetaMask checks against this list: https://github.com/ethereum-lists/chains
> It might make sense to just fetch it from the list when we generate the configuration files.

@manuelwedler Is it okay if we close the issue with this PR (since the bug will no longer be there) and create another issue to optimize this approach by moving the fetching part to the config script (automize it).

manuelwedler:
Yeah, think it could make sense to fetch the list during build to eliminate human error. Otherwise, we would have to check against it every time we add a network manually, and this process is not documented. If you prefer to create a new issue, do so, we could also just leave the current open.

GabrielBuragev:
Since the original issue is documenting a bug i would rather have that issue closed so it doesn't mess with our planning ahead. 
The new issue will be documenting an optimization instead.

frontend/src/services/web3-provider/ethereum-provider.ts
```diff
@@ -66,13 +67,17 @@ export abstract class EthereumProvider extends EventEmitter implements IEthereum
         chainName: name,
         rpcUrls: [rpcUrl],
         blockExplorerUrls: [explorerUrl],
-        nativeCurrency: {
-          name: 'Ethereum',
-          symbol: 'ETH',
-          decimals: 18,
-        },
+        nativeCurrency: nativeCurrency
+          ? toRaw(nativeCurrency)
```
GabrielBuragev, line 71: Lest move this `toRaw` function call [here](https://github.com/beamer-bridge/beamer/pull/1330/files#diff-8aa979d6df2fe2634cc8483af13978d4bb68643da290dd7c781a791d247290fbR34) instead.
I guess it would be nice to keep the `ethereum-provider.ts` module less dependent on vue's internals.
Composables anyways are supposed to be strongly dependent on the vue's specifics.

GabrielBuragev (changes requested):
Lets have consistent chain config definitions and include the `nativeCurrency` in the other config files.
Also, a few improvement suggestions. 

Besides that everything looks fine.

Also, big thanks for going beyond what the initial issues are about and fixing things on the way!

frontend/src/services/web3-provider/ethereum-provider.ts
GabrielBuragev: Actually, i am just trying this out now, seems like it will not work if we move the `toRaw`  to the composable. So ignore this for now.

frontend/src/services/web3-provider/ethereum-provider.ts
```diff
@@ -66,13 +67,17 @@ export abstract class EthereumProvider extends EventEmitter implements IEthereum
...
+          : {
+              name: 'Ethereum',
```
compojoom, line 73: The name of the nativeCurrency is actually Ether and not Ethereum.

GabrielBuragev: True!

GabrielBuragev (approved):
Awesome, thanks!

PR #1342, opened 2023-01-11
beamer: util: remove Rinkeby and Kovan since we don't support them

istankovic:


fredo (approved):
looks good. We should move this check out of the code base into a static config analyzer

PR #1344, opened 2023-01-11
frontend: Hotfix for MetaMask throwing disconnect events when switching chain

### LINKED ISSUE #1118: Frontend: Switching between networks triggers a wallet disconnect
The implementation of the internal wallet disconnect detection somehow breaks the network switching flow. 
When a network switch is being innitiated the `Eip1193Provider` emits a `disconnect` event for some reason.
We need to rethink the implementation here and come up with another solution.

wallet-disconnect

## Accaptence criteria
- [ ] Make sure problem does not happen for all wallets that we support

Related issue: #692 
Blame PR: https://github.com/beamer-bridge/beamer/pull/1087

manuelwedler:
Fixes #1118 

MetaMask sometimes throws a disconnect event when switching the chain. In this case we need to ignore the event, because the account is actually still connected. This adds a flag that is true for 1 second after a chain switch happened. A short time period is necessary because the event is not directly thrown when the `wallet_switchEthereumChain` rpc method returns. If this is fixed on the side of MetaMask, we can remove the flag again.

### DIFF
--- a/frontend/src/services/web3-provider/ethereum-provider.ts
+++ b/frontend/src/services/web3-provider/ethereum-provider.ts
@@ -18,6 +18,8 @@ export abstract class EthereumProvider extends EventEmitter implements IEthereum
   protected web3Provider: Web3Provider;
   protected externalProvider: Eip1193Provider;
 
+  private isSwitchingNetwork = false;
+
   constructor(_provider: Eip1193Provider) {
     super();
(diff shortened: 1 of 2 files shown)


GabrielBuragev (approved):
Really sad that we need to implement this but it is what is.
Well done!

PR #1347, opened 2023-01-11
frontend: Fix fee query on transfer creation

### LINKED ISSUE #1340: Incorrect token allowance
Use current master and launch app with --mode staging. 
Try to transfer from optimism to arbitrum (maybe from where to where doesn't matter, but haven't tested).

Set an amount that you want to transfer that is higher than anything you've allowed. Submit. Metamask asks you to confirm the allowance and once we do the transfer fails with  insufficient balance.

I think that the issue is that we don't account for the transfer token fees. 
So, if I transfer 50, we actually need 51, but we only set the allowance to 50.

manuelwedler:
closes #1340 

When creating a transfer, the request manager was queried for the fees of the target token address (which only exists on the target chain). This obviously returns an erroneous value as the request manager only knows about the source token address. This led to the bug that the fees were not taken into account for the getting the token allowance. This simply changes the `getRequestFee` call to use the correct variable.

### DIFF
--- a/frontend/src/composables/useTransferRequest.ts
+++ b/frontend/src/composables/useTransferRequest.ts
@@ -26,7 +26,7 @@ export function useTransferRequest() {
 
     const validityPeriod = new UInt256(VALIDITY_PERIOD_SECONDS);
     const { rpcUrl, requestManagerAddress } = options.sourceChain;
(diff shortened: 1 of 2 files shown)


GabrielBuragev (approved):
Looks good! Thanks for extending the test!

Issue #1195, opened 2022-11-21
Frontend: Increase font-sizes for small text components


GabrielBuragev:
## Rationale

Many community members, team members, as well as one [UX auditor](https://github.com/beamer-bridge/beamer/issues/1132) reported small font-sizes for some of the app's components:
- short sentences (text)
- button texts
- error (validation) messages

### UX Auditor's recommendation
> Some text like the "Add to wallet" button and connection status is too small. No text should be below 14px font size. Only long contigous text like the Terms & Conditions page could have 12px.

## Implementation

Use 14px as a minimal font-size for all the components or the page that are not long contigous texts.

PR #1351, opened 2023-01-12
frontend: improve ui

### LINKED ISSUE #1219: Frontend: Too many decimals error breaks layout
Small layout issue with the input validation error messages: When using more decimals than allowed by the selected token, the error message breaks the layout by letting the upper box grow. So the box moves in an ugly way while the user types.

Looks like this only happens on Firefox.

#### Expected behavior
The text fits the line below the input.

#### Actual behavior
image

#### System
Firefox 107.0

compojoom:
The font-size on the page was set to 50% which was causing some buttons to be barelly readable on screens with high resolution. In adition to that one had to zoom the page to get it to a comfortable size.

I’ve moved the whole container lower on the page -> it is now bellow the "logo area". The logo container is now also restricted in width so that the logo won’t appear on the far left of the page and the content of the page would be in the middle.

The page is now also more responsive - on a bigger browser the actual content is wider.

The commit also fixes #1219 . I've set the height of the whole block to be bigger than the content. This way if we end up with content on 2 lines the form won't jump.

Here are some screenshot. First the currently deployed website:
image

image

Here is what the the changes look like:

image

image

image

image


istankovic:
I like the changes in here. Is it possible to also have the "terms & conditions" line centered or at least aligned somewhere so it does not move (left-right)?

compojoom:
Hey Ivan,
Yes, I actually made a second commit after posting the changes exactly for this.

image

olympe-silve:
That's awesome Daniel, thank you so much

GabrielBuragev (changes requested):
Great work Daniel! Fixing the design was a very tedious but yet necessary task, so thank you!
I noticed a few issues with how the rest of the components scaled due to these changes.

GabrielBuragev:
The wallet selection menu grew in size. Can we scale the width/height of the wallet option cards down so it comes slightly closer to the sizes on the current mainnet frontend?

Screenshot 2023-01-13 at 10 39 08

GabrielBuragev:
Can we introduce a max-width to the tooltip card so it doesn't grow unnecesarily?
The preferred sizing can be found on the current mainnet frontend.

Screenshot 2023-01-13 at 10 39 58

GabrielBuragev:
The loading indicators are broken somehow:


https://user-images.githubusercontent.com/18466333/212296026-e88ce502-58d5-4584-9a43-9d102a38e601.mov

GabrielBuragev:
I think we are also fixing issue https://github.com/beamer-bridge/beamer/issues/1195 with this PR right?

compojoom:
> The wallet selection menu grew in size. Can we scale the width/height of the wallet option cards down so it comes slightly closer to the sizes on the current mainnet frontend?
> 
> Screenshot 2023-01-13 at 10 39 08

I've synced with @GabrielBuragev @andfletcher and we've decided to modify the "connect wallet" screen as follow:
image

Notice the disclaimer: "By connecting a wallet, you agree to the terms of service" there. 

I've removed the checkbox from the main screen:
image

compojoom:
> Can we introduce a max-width to the tooltip card so it doesn't grow unnecesarily? The preferred sizing can be found on the current mainnet frontend.
> 
> Screenshot 2023-01-13 at 10 39 58

There was a max width, but it was set to 40rem... I've changed it 20. In addition to that I improved the tooltip, by adding a border to it and fixing the "arrow". The arrow was not an arrow, but a rotated square. Since the square had the same color as the background it was looking like a triangle. The problem is the tooltip background is the same background as other elements on the page it was very hard to read... Now they look like this:

image

image

compojoom:
image

image

When the user hovers over the tooltip:
image

image


image

GabrielBuragev (approved):
Well, this has been a loong process. We fixed and improved many things. Thanks for your patience and hard work here @compojoom!

manuelwedler (approved):
Just checked the activity tab. Looks much nicer with the enforced height. Thanks!

PR #1356, opened 2023-01-13
docker: Update dockerfile to use shiv

bilbeyt:


docker/Dockerfile.agent
```diff
@@ -1,17 +1,15 @@
...
-
-COPY README.md pyproject.toml poetry.lock /src/
-COPY beamer /src/beamer
+FROM python:3.10-alpine
 
+COPY . /src
 RUN mkdir -p /src/beamer/data/relayers
-COPY ./relayer/relayer-node16-*-x64 /src/beamer/data/relayers/
-
+COPY ./relayer/relayer-node18-linux-x64 /src/beamer/data/relayers/
```
istankovic, line 5: We should copy both relayers since we want to use the same image on both platforms.

istankovic: Nevermind the above comment. People on mac can use the same alpine-based image and it will work.

istankovic (approved):
Awesome! :purple_heart:

PR #1358, opened 2023-01-13
docker: Dockerfile.agent: use python:3.10-slim instead of -alpine

istankovic:
The -alpine image does not have the C++ runtime lib, which is required by our relayer binaries.


istankovic:
After merging this we need to
- release v1.0.1
- update beamer-bridge/run-your-own-agent to use v1.0.1
- update testnet agents to use v1.0.1

bilbeyt (approved):

PR #1367, opened 2023-01-16
chore: remove mentionings of rinkeby

fredo:
I looked for the remaining mentionings of rinkeby and either removed or rpleaced them with goerli

I also removed URLs in example configs because we had multiple confusions by contributors that they can use these URLs


GabrielBuragev (approved):
✅

PR #1384, opened 2023-01-18
frontend: deal better with reverted txs

compojoom:
We were fetching the transaction receipt and were always assuming that if we have a tx receipt we have a succesfull tx. In some situations though, we had reverted transaction. We were trying to parse the logs on those, which triggered a js error: `Cannot read properties of undefined (reading 'topics')`

We now check the receipt status and try to parse the logs only if it is succesful.

### DIFF
--- a/frontend/src/services/transactions/request-manager.ts
+++ b/frontend/src/services/transactions/request-manager.ts
@@ -139,14 +139,24 @@ export async function getRequestIdentifier(
     provider,
   );
   const receipt = await provider.waitForTransaction(transactionHash, 1);
+
   if (receipt) {
-    const event = contract.interface.parseLog(receipt.logs[0]);
-    if (event) {
-      return event.args.requestId;
+    if (receipt.status === 1) {
+      try {
+        const event = contract.interface.parseLog(receipt.logs[0]);
+
+        if (event) {
+          return event.args.requestId;
+        }
+      } catch (e) {
+        throw new Error("Request Failed. Couldn't retrieve Request ID.");
+      }
+    } else {
+      throw new Error('Transaction reverted on chain.');
     }
   }
 
-  throw new Error("Request Failed. Couldn't retrieve Request ID");
+  throw new Error('Transaction not found.');
 }
 
 export type RequestData = {
--- a/frontend/tests/unit/services/transactions/request-manager.spec.ts
+++ b/frontend/tests/unit/services/transactions/request-manager.spec.ts
@@ -313,6 +313,52 @@ describe('request-manager', () => {
       expect(identifier).toBe('1');
     });
 
+    it('throws an error if the transaction has reverted', async () => {
+      const transactionHash = getRandomString();
+
+      mockGetContract();
+
+      const provider = mockGetProvider();
+      provider.waitForTransaction = vi
+        .fn()
+        .mockReturnValue(new MockedTransactionReceipt({ status: 0 }));
+
+      expect(
+        getRequestIdentifier(RPC_URL, REQUEST_MANAGER_ADDRESS, transactionHash),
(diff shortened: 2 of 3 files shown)


frontend/tests/utils/mocks/ethers.ts
```diff
@@ -12,7 +12,13 @@ export class MockedTransaction {
 }
 export class MockedTransactionReceipt {
   logs = [];
+  status = 1;
```
GabrielBuragev, line 15: We can create a constructor here and set the value of the status based on the value of the passed argument to the constructor. 
Default will be 1.

Then we can use this mock class for both cases.

frontend/src/services/transactions/request-manager.ts
```diff
@@ -139,14 +139,15 @@ export async function getRequestIdentifier(
...
+
+  if (receipt && receipt.status === 1) {
     const event = contract.interface.parseLog(receipt.logs[0]);
     if (event) {
       return event.args.requestId;
     }
   }
 
-  throw new Error("Request Failed. Couldn't retrieve Request ID");
+  throw new Error('Transaction reverted on chain.');
```
GabrielBuragev, line 150: i think the case where a receipt is undefined and a receipt status is 0 are two different cases right?

1. When its undefined it means that the request id was not found for this transaction.
2. When `receipt.status === 0` it means that the transaction was reverted.

GabrielBuragev (changes requested):

frontend/src/services/transactions/request-manager.ts
compojoom: How do you end up with an undefined for a transactionReceipt?

GabrielBuragev: Check this out: https://docs.ethers.org/v5/api/providers/provider/#Provider-waitForTransaction

I don't know how to interpret this part: `If confirms is 0, this method is non-blocking and if the transaction has not been mined returns null.`

Does it return null when transaction has not been mined or does it only return null when confirms is 0 AND transaction has not been mined?

frontend/tests/unit/services/transactions/request-manager.spec.ts
```diff
@@ -320,7 +340,7 @@ describe('request-manager', () => {
 
       expect(
         getRequestIdentifier(RPC_URL, REQUEST_MANAGER_ADDRESS, transactionHash),
-      ).rejects.toThrow("Request Failed. Couldn't retrieve Request ID");
+      ).rejects.toThrow('Transaction reverted on chain.');
```
GabrielBuragev, line 343: This message is a bit missleading in this case then. It should throw "Request Failed. Couldn't retrieve Request ID" when there were no logs found in the receipt.logs right?

GabrielBuragev (approved):
Looks perfect now! Splitting the different error cases inside the `getRequestIdentifier` will definitely help by providing more valuable feedback to the user.

I am squashing the commits & applying a few fixes regarding typo & redundant code.

frontend/tests/unit/services/transactions/request-manager.spec.ts
```diff
@@ -331,8 +330,7 @@ describe('request-manager', () => {
     it('throws an error if the transaction receipt is undefined', async () => {
       const transactionHash = getRandomString();
 
-      const contract = mockGetContract();
-      contract.interface.parseLog = vi.fn().mockReturnValue({ args: { requestId: '1' } });
```
GabrielBuragev: There was no need of mocking the `parseLog` here as this test case was not reaching that scope due to the condition checks.

compojoom:
awesome!

PR #1387, opened 2023-01-18
frontend: Add Arbitrum notification banner

manuelwedler:
See https://github.com/beamer-bridge/marketing/issues/234

Just adds a banner to notify users that Beamer is live on Arbitrum.


GabrielBuragev (approved):
Looks good!

PR #1406, opened 2023-01-19
frontend: prevent user from submitting a second tx

compojoom:
If the user has another transfer pending, we prevent the user from submitting a second transfer.

We don’t treat a transfer to be pending if the user has signed all transactions and they’ve been mined. If that is the case and an agent still hasn’t fullfiled the request - this is not a pending transaction.

The reason we do that is that before submiting a transaction we check the allowance and create an allowance tx. But if the actual transaction that submit the money to the contract is not mined yet, then the second tx that the user starts has an incorrect allowance and the tx that sends the money to the contract fails with "incorrect allowance".

We are currently outputting this message:
image

I've moved the footer to an absolute position and suck it on the bottom of the page. I applied bottom padding to the main container so that on smaller screens the footer is always some pixels below the content & they don't overlap
image

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="app-content">
+  <div class="app-content relative">
     <div
       class="flex h-16 w-full flex-col justify-center bg-fire text-center text-xl font-semibold text-rosa lg:text-2xl"
     >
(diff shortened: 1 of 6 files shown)


GabrielBuragev (changes requested):
Looks good overall, thanks! However, i would like to try moving the logic to the store. 
It would simplify things by allowing access to the `pendingTxs` reactive object across the whole app. Also, since this logic solely depends on 2 arguments (transfers & chain id) it can easily be integrated into the `transfer-history` store. 

Taking over this PR as discussed with @compojoom to implement these changes.

compojoom (commented):
perfect!

PR #1423, opened 2023-01-25
Fix UI

### LINKED ISSUE #1383: frontend: layout issues
I experienced two layout issues during testing:

(1) When the token balance is 0, the balance text is shown in white text and is much bigger than usual. Besides, the balance text is not on the same line as the `Add to Wallet` button, also for balances bigger than 0.
Screenshot from 2023-01-18 12-29-03
Screenshot from 2023-01-18 12-29-38

(2) In the activity tab, there might be a gap between the line and the next circle of the steps. This only happens if the text does not fit in one line.
Screenshot from 2023-01-18 12-33-24

### LINKED ISSUE #1390: Flickering when transfer in progress
If you have a previous transfer in the queue the previous transfer flickers in the UI. For some reason the min-height property is always changing between 20px and 470px

GabrielBuragev:
Fixes #1383 
Fixes #1390 

A couple of UI layout fixes & a few other minor improvements

### DIFF
--- a/frontend/src/assets/images/copy.svg
+++ b/frontend/src/assets/images/copy.svg
@@ -1,0 +1,15 @@
+<svg
+    class="inline"
+    xmlns="http://www.w3.org/2000/svg"
+    width="16"
+    height="16"
+    viewBox="0 0 24 24"
+    fill="none"
+    stroke="currentColor"
+    stroke-width="1.5"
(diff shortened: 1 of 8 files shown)


compojoom:
Ok, with the latest change it looks perfect! Thank you!

compojoom (approved):

Issue #1242, opened 2022-12-02
Test OP relayer in e2e script


fredo:
We want to be able to test the relayers locally. We can integrate them into the e2e test script. 

- [x] turn off relayer instances provided by optimism
- [x] fetch optimism addresses from local L1
- [x] provide them to the relayer
- [x] adapt the relayer that it can consume a custom network for optimism
- [x] adapt the scripts to run the optimism relayer

andfletcher:
@andfletcher create a followup issue for arbitrum

GabrielBuragev:
@andfletcher have you already created a followup issue for arbitrum?

PR #1433, opened 2023-01-27
Arbitrum e2e

istankovic:
TODO:
- [x] switch Optimism E2E test to use new helper scripts
- [x] remove the old `e2e-test.py` script
- [x] investigate why the outbox entry sometimes takes too long to appear after a fill


fredo:
looks very good! 
I have a couple of questions, which you can probably answer easier. 

1) It looks like that the relayer is not running for the optimism test. Is Optimism still running it's own internal relayer to relay the message on L1? 
2) It looks that there is quite some room left for code reduction. A bunch of the code for the Arbitrum and Optimism e2e scripts and CI calls looks quite similiar. Is there more room for optimization? And if you don't want to do it now (which is totally fine), could you create a follow up issue so that it does not get lost?

Otherwise, amazing job!

fredo (approved):

istankovic:
>     1. It looks like that the relayer is not running for the optimism test. Is Optimism still running its own internal relayer to relay the message on L1?

Yes, that part was out of scope for this issue and it's covered by #1242.
This PR just makes the Optimism script use the same e2e test scripts as the Arbitrum one, so that we
can remove the old e2e test script.

>     2. It looks that there is quite some room left for code reduction. A bunch of the code for the Arbitrum and Optimism e2e scripts and CI calls looks quite similar/redundant. Is there more room for optimization? And if you don't want to do it now (which is totally fine), could you create a follow up issue so that it does not get lost?

It's true the two scripts look quite similar, but I'm not sure how much we could optimise things further. I guess what could be done is to try to move the command-dispatch code to a common area, but other than that, I think it would be difficult to do much more.

Thanks for the review!

PR #1445, opened 2023-02-01
frontend: fix misconfigured code coverage

### LINKED ISSUE #1408: Code coverage doesn't report anything
Code coverage is reporting 0% for the current frontend code. The report seems to be generated, but the tools don't display it correctly. We have to investigate why.

GabrielBuragev:
fixes #1408 

After a recent update of the vitest library from 0.9.3 to 0.12.10 the code coverage was broken. The `yarn test:coverage` was not outputting anything.
The reason for this was a change in the config interface that is used inside vite.
Also, the yarn.lock file had 2 different versions of vite installations so the interfaces of both version were conflicting.


manuelwedler (approved):
Looks good :+1:

PR #1446, opened 2023-02-01
frontend: add GPL license

GabrielBuragev:
The frontend had no license specified in its `package.json`.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -2,6 +2,7 @@
   "name": "beamer-request-interface",
   "version": "2.0.0",
   "private": true,
+  "license": "GPL-3.0-or-later",
   "scripts": {
     "dev": "vite",
     "build": "yarn configure && vite build",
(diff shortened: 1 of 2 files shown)


istankovic:
Are we ok with GPL?
What licenses are used by our dependencies?
Did you check `NODE_ENV=production yarn licenses list`?

andfletcher:
MIT and GPL are fine for me.

andfletcher:
Also everything we do should have a licens

istankovic (approved):

istankovic:
@GabrielBuragev could you please check whether we need to do the same for relayer?

GabrielBuragev:
I was thinking if we use the GPL for the relayer as well. What do you guys think?

istankovic:
> I was thinking if we use the GPL for the relayer as well. What do you guys think?

Yes, makes sense to use the same for both.

PR #1447, opened 2023-02-01
frontend: add placeholder image inside selector

### LINKED ISSUE #933: Frontend: Add placeholder for SelectorOptions without imageUrl
At this moment, inside the token selection we don't show the token icons.
It would be way more convenient for users to see their token icon, if available.
If no icon is available for their token, we would like to show a placeholder icon with a question mark for example.

image

GabrielBuragev:
fixes #933


manuelwedler (approved):
Looks good!

Issue #1435, opened 2023-01-30
agent: ConnectionError upon sync breaks agent's invariants


fredo:
There is a bug when a `ConnectionError` happens during sync.
A connection error will cause the `fetch()` function returning `[]` as a result not noting the EventMonitor that something went wrong.
The EventMonitor will then stay unnoticed and call `on_sync_done()`, noticing the EventProcessor that it can start processing. This is a very critical failure as the agent will then start processing events with missing events on one (or both) of the chains. This can result in misbehavior.

https://github.com/beamer-bridge/beamer/blob/f8b68975be6f98db087f616087458e33b0380851/beamer/events.py#L295-L298 

During runtime this may be fine, but especially during sync, this is a critical bug.

PR #1458, opened 2023-02-04
Handle ConnectionError on sync

fredo:
I saw two solutions: 
Either throwing the error up until the point where `on_sync_done()`, namely the EventMonitor. I didn't like that option because I think handling request exceptions inside of the EventFetcher is the right place.

The other option, was to check whether the fetch call synced up until a certain point. I went with that option. Fetching the BlockNumber before calling fetch() for the sync. Exposing the latest synced block by a property in the EventFetcher. And checking whether the fetcher synced at least until this block.

I realized that not only the ConnectionError described in #1435  is a problem, also a RequestException during fetching the blockNumber in https://github.com/beamer-bridge/beamer/blob/9935c4ab77044d17645329e68c0e259639cebab0/beamer/events.py#L288 results in the same problem.

### DIFF
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -84,7 +84,10 @@ class EventMonitor:
         )
 
         fetcher = EventFetcher(self._web3, self._contracts, self._deployment_block)
-        events = fetcher.fetch()
+        current_block = self._web3.eth.block_number
+        events = []
+        while fetcher.synced_block < current_block:
+            events = fetcher.fetch()
         if events:
             self._on_new_events(events)
         self._on_sync_done()
--- a/beamer/chain.py
+++ b/beamer/chain.py
@@ -87,7 +87,7 @@ class EventMonitor:
         current_block = self._web3.eth.block_number
         events = []
         while fetcher.synced_block < current_block:
-            events = fetcher.fetch()
+            events.extend(fetcher.fetch())
         if events:
             self._on_new_events(events)
         self._on_sync_done()
--- a/beamer/events.py
+++ b/beamer/events.py
@@ -285,7 +285,8 @@ class EventFetcher:
 
     def fetch(self) -> list[Event]:
         try:
-            block_number = BlockNumber(self._web3.eth.block_number)
+            block_data = self._web3.eth.get_block("latest")
+            block_number = BlockNumber(block_data["number"])
         except requests.exceptions.RequestException:
             return []
 
--- a/beamer/events.py
+++ b/beamer/events.py
@@ -225,6 +225,10 @@ class EventFetcher:
                 self._chain_id == contract.web3.eth.chain_id
             ), f"Chain id mismatch for {contract}"
 
+    @property
+    def synced_block(self) -> BlockNumber:
+        return BlockNumber(self._next_block_number - 1)
+
     def _fetch_range(
         self, from_block: BlockNumber, to_block: BlockNumber
     ) -> Optional[list[Event]]:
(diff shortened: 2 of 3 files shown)


beamer/chain.py
```diff
@@ -84,7 +84,10 @@ def _thread_func(self) -> None:
         )
 
         fetcher = EventFetcher(self._web3, self._contracts, self._deployment_block)
-        events = fetcher.fetch()
+        current_block = self._web3.eth.block_number
+        events = []
+        while fetcher.synced_block < current_block:
+            events = fetcher.fetch()
```
istankovic, line 90: I think this could lose events in case the first `fetch` call returns a nonempty list, but still did not manage to fetch until `current_block`, due to a `ConnectionError`. Since we overwrite `events` on each iteration, we would lose all but the last batch of events.

However, I'm thinking moving this 3-liner to `EventFetcher` would be more appropriate since
`EventFetcher.fetch` semantics has been to fetch until the current block. In that case you would not even
need the `synced` property and the current code in monitor could stay the same.

Something like:
```python
    def fetch(self) -> list[Event]:
           events = []
           while fetcher.synced_block < current_block:
               events.extend(self._fetch())
```
(here `self._fetch` would be the current `EventFetcher.fetch` -- yeah, naming is not perfect...)

What do you think?

Additional concern here is that there's a possibility that we end up with multiple `LatestBlockUpdatedEvent` , instead of just one, ideally. But that would be trivial to filter out in even fetcher and we could just leave the last 
(most recent one) `LatestBlockUpdatedEvent` in the resulting list.

beamer/events.py
```diff
@@ -285,7 +285,8 @@ def _fetch_range(
 
     def fetch(self) -> list[Event]:
         try:
-            block_number = BlockNumber(self._web3.eth.block_number)
+            block_data = self._web3.eth.get_block("latest")
+            block_number = BlockNumber(block_data["number"])
```
istankovic, line 289: This is nice. If we take care to just return one `LatestBlockUpdatedEvent` from `fetch`, we should be able to reuse the data we got here during construction of the `LatestBlockUpdatedEvent` at the end of this function. We can thereby save another `get_block` call and unnecessary work.

beamer/chain.py
fredo: Moving this into the EventFetcher would result in a change in the logic that a connection error does not break and return an empty array. It would get stuck in the loop until it was fetched. I actually liked the approach that during normal runtime we just return on a ConnectionError and wait for the next cycle. 
The different desired behavior is only on sync

fredo: > I think this could lose events in case the first fetch call returns a nonempty list, but still did not manage to fetch until current_block, due to a ConnectionError. Since we overwrite events on each iteration, we would lose all but the last batch of events.

🤦

fredo: > Additional concern here is that there's a possibility that we end up with multiple LatestBlockUpdatedEvent , instead of just one, ideally. But that would be trivial to filter out in even fetcher and we could just leave the last
(most recent one) LatestBlockUpdatedEvent in the resulting list.

I don't see a problem with that. The behavior on processing LatestBlockUpdatedEvent is exactly equal to if it came in different cycles. A more recent event overwrites the older data.
Since it comes in the correct order so there should be no problem.

beamer/events.py
fredo: see my comment above

beamer/chain.py
istankovic: > I don't see a problem with that. The behavior on processing LatestBlockUpdatedEvent is exactly equal to if it came in different cycles. A more recent event overwrites the older data.
> Since it comes in the correct order so there should be no problem.

Right, I was mostly worried about being possibly less efficient in case of multiple events, but I guess as things stand, it's not a problem at all.

istankovic (approved):
Looks good! Assuming the "lost events" issue is fixed. ;-)

PR #1462, opened 2023-02-07
frontend: Fix wallet disconnection after chain switch

### LINKED ISSUE #1118: Frontend: Switching between networks triggers a wallet disconnect

manuelwedler:
Closes #1118 

Slow RPCs can still lead to the problem that after a chain switch a wallet disconnection occurs. Looks like the timeout of 1 second is not enough. To fix it deterministically, I removed the listener on the disconnect event completely. The listener was used to catch the case when a user clicks disconnect inside MetaMask. In this case also an accountsChanged event is fired. We can simply rely on this event to call our disconnect function. I also removed the no disconnection after chain switch test as it is not applicable anymore. There is no test, because I think the new behavior is not coverable in a unit test.

### DIFF
--- a/frontend/src/services/web3-provider/ethereum-provider.ts
+++ b/frontend/src/services/web3-provider/ethereum-provider.ts
@@ -18,8 +18,6 @@ export abstract class EthereumProvider extends EventEmitter implements IEthereum
   protected web3Provider: Web3Provider;
   protected externalProvider: Eip1193Provider;
 
-  private isSwitchingNetwork = false;
-
   constructor(_provider: Eip1193Provider) {
     super();
     this.web3Provider = new Web3Provider(_provider);
@@ -35,7 +33,6 @@ export abstract class EthereumProvider extends EventEmitter implements IEthereum
   async switchChainSafely(newChain: Chain): Promise<boolean> {
     let successful = true;
     if (newChain.identifier !== this.chainId.value) {
-      this.isSwitchingNetwork = true;
       try {
         const isSuccessfulSwitch = await this.switchChain(newChain.identifier);
         if (!isSuccessfulSwitch) {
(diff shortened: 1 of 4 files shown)


GabrielBuragev (changes requested):
Nice! I like this solution. Works on my browsers as well. I have one suggestion though.

I tested WalletConnect on mobile with BitKeep, and seems like the dapp is not disconnecting when i force a disconnect from BitKeep which is to be expected since we removed the `disconnect` listener.

I think the right approach here would be to overwrite the `listenToEvents` inside `metamask-provider.ts` instead of removing the `disconnect` listener for all providers that extend from the base provider inside `ethereum-provider.ts`.

GabrielBuragev:
Well, here is my assumption:

It mostly depends on how the underlying software behaves (BitKeep in my example).
WalletConnect just creates a communication channel between the dapp and the underlying wallet app
The underlying wallet app then decides if, and when, they are emitting the events and such. 
This conclusion is based on my experience testing the different wallet apps. Haven't looked into how it is really integrated internally though.

Slightly off the record: I saw pretty inconsistent behavior on different wallet apps when testing wallet connect v2. I hope this will change as wallet connect v2 migration is expected to be completed beginning of March.

GabrielBuragev:
Yeah, i am thinking of showing both wallet connect v1 and v2 in the menu until we see some progress there. 
But lets discuss this in the respective PR that will come for the v2 integration.

GabrielBuragev (approved):
Perfect! Thanks for addressing my suggestion.

PR #1468, opened 2023-02-08
frontend: Add social media links to footer

### LINKED ISSUE #1467: Frontend: Add social media buttons in app footer
## Description of Deliverables(s)

We are looking to grow our community and increase traffic to medium. Add our social media buttons in the  footer. 

image

What we expect: more traffic to our community + medium channels

Reference: https://www.optimism.io/

## Acceptance criteria
- [ ] social media buttons are also located in the footer (there where we have the imprint and the terms)

manuelwedler:
closes #1467


manuelwedler:
Actually, I don't think they add much value here. If we would have to change a link, we would copy it to two places. When you have to copy-paste a value from the template to the tests, I don't think the tests ensure that the value is correct. So, I usually don't write tests for templates that do not involve any actual code. It's better to have a closer look when reviewing for such things imo. But since you already wrote them, I will leave them in.

PR #1469, opened 2023-02-09
EventMonitor fetch all events per chain

bilbeyt:
EventMonitor is adapted to use related chain_ids and using both contracts to fetch events.

### DIFF
--- a/beamer/agent.py
+++ b/beamer/agent.py
@@ -1,8 +1,7 @@
 import threading
 from concurrent.futures import ThreadPoolExecutor
-
 import structlog
-from eth_typing import Address
+from eth_typing import Address, BlockNumber
 from web3 import Web3
 from web3.middleware import latest_block_based_cache_middleware
 
@@ -26,6 +25,12 @@ def _get_contracts_info(config: Config, web3: Web3) -> dict[str, ContractInfo]:
     return info
 
 
+def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber:
+    request_manager_deployment_block = contract_info["RequestManager"].deployment_block
+    fill_manager_deployment_block = contract_info["FillManager"].deployment_block
+    return min(request_manager_deployment_block, fill_manager_deployment_block)
+
+
 class Agent:
     def __init__(self, config: Config):
         self._config = config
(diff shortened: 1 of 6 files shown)


beamer/agent.py
```diff
@@ -1,8 +1,8 @@
 import threading
 from concurrent.futures import ThreadPoolExecutor
-
+from typing import Dict
```
istankovic, line 3: Nitpick: instead of `Dict` you can just use regular type `dict`, it will work.

beamer/chain.py
```diff
@@ -117,15 +116,16 @@ def __init__(self, context: Context):
         # 2 = both chains synced
```
fredo, line 116: can you update the comments here?

beamer/agent.py
```diff
@@ -84,37 +97,41 @@ def _init(self) -> None:
...
+                    poll_period=POLL_PERIOD,
+                )
 
     def start(self) -> None:
         assert self._stopped.is_set()
         beamer.metrics.init(self._config)
         self._event_processor.start()
-        self._event_monitor_l2a.start()
-        self._event_monitor_l2b.start()
+        for em in self._event_monitors.values():
```
fredo, line 126: nitpick: I would appreciate a clearer naming like `event_monitor` in general.

beamer/agent.py
```diff
@@ -84,37 +97,41 @@ def _init(self) -> None:
...
+                )
+            else:
+                self._event_monitors[chain_id] = EventMonitor(
+                    web3=w3_target,
+                    contracts=(target_request_manager, target_fill_manager),
+                    deployment_block=_get_deployment_block(target_contracts_info),
+                    on_new_events=[self._event_processor.add_events],
+                    on_sync_done=self._event_processor.mark_sync_done,
+                    poll_period=POLL_PERIOD,
+                )
```
fredo, line 120: minor nitpick: you could pull out the creation of the EventMonitor and assignment in the dict to after the if clause, and assign the source and target rollup dependent variables inside of the if/else

beamer/state_machine.py
```diff
@@ -66,9 +66,9 @@ class Context:
 
 def process_event(event: Event, context: Context) -> HandlerResult:
     if isinstance(event, SourceChainEvent) and event.chain_id != context.source_chain_id:
-        return True, []
+        return True, None
     if isinstance(event, TargetChainEvent) and event.chain_id != context.target_chain_id:
-        return True, []
+        return True, None
```
istankovic, line 71: Nitpick: could you please move this change into the previous commit, where the list was introduced? Seems like that commit should have this fix which is anyway not all that related to other changes in this commit.

beamer/state_machine.py
```diff
@@ -63,6 +65,11 @@ class Context:
 
 
 def process_event(event: Event, context: Context) -> HandlerResult:
+    if isinstance(event, SourceChainEvent) and event.chain_id != context.source_chain_id:
+        return True, None
+    if isinstance(event, TargetChainEvent) and event.chain_id != context.target_chain_id:
+        return True, None
```
istankovic, line 71: The idea to mark the events via a base type is quite nice. This is really elegant. :+1:

fredo (approved):
looks good to me! 

In my opinion you wouldn't have needed to pack all the EventMonitor parameters into a dict. It would have been sufficient to just set the three source/target dependent parameters in the if/else clause and then initialize the EventMonitor below directly. But I'll leave that up to you, if you want to change it.

beamer/agent.py
```diff
@@ -84,37 +96,43 @@ def _init(self) -> None:
...
+                    "deployment_block": _get_deployment_block(source_contracts_info),
+                }
+            else:
+                chain_related_data = {
+                    "web3": w3_target,
+                    "contracts": (target_request_manager, target_fill_manager),
+                    "deployment_block": _get_deployment_block(target_contracts_info),
+                }
+            event_monitor_data.update(chain_related_data)
+            self._event_monitors[chain_id] = EventMonitor(**event_monitor_data)  # type: ignore
```
istankovic, line 121: I would avoid updating `event_monitor` on every loop iteration:
```python
            event_monitor_data.update(chain_related_data)
```
because even though it currently works, it's relatively fragile to reuse the same `event_monitor_data` dict instance.

Instead, I would suggest something like
```python
            kwargs = dict(event_monitor_data, **chain_related_data)
            self._event_monitors[chain_id] = EventMonitor(**kwargs)  # type: ignore
```

istankovic: Having thought some more about this, I now think we should rewrite the whole function so that there is almost no distinction between source and target chain IDs (the exceptions would be e.g. checvks like the one for max validity period above).

So, instead of having `source_contracts` and `target_contracts` and similarly for other variables, we should intend to treat all chains the same way, to the fullest possible extent.

To that end, I think a dict such as the following could help:

```python
   chains: dict[ChainID, dict[str, Any]] = {}
```

(not sure about the value type here yet, but something along those lines).
This would bring us closer to what we want for the multiple-chain-pairs support.
Of course, for now we would still honor the `source-chain` and `target-chain` config values, but later in the epic we will simply deprecate those and look at everything.

So what I propose to do here would be:

- for each non-L1 chain from the config:
- - if the chain is not one of (source-chain, target-chain), continue [note: this will be dropped later]
- - build the contract info, web3 and all other needed objects and put them into the `chains` dict under the chain_id key
- perform the sanity checks we do now [this is one place where we need to distinguish between source and target chains]
-  create the context for the source and target chain [this is another place where we need to distinguish between them]
- for each chain in `chains`:
- - create an appropriate event monitor instance [note: there should be no need for an `if ` here]

istankovic (changes requested):
Overall looks pretty good, but I think we need to some more work in `Agent.__init__`.

beamer/agent.py
fredo: can we continue this discussion somewhere? I think this is important. 
We do have and will continue to have the concept of source and target chain in the EventProcessor. Will will just not have it anymore on the agent's level.

Having read the above, I think there are some details missing in the Epic to make the new architecture work with our current setup. The Event processors will be created generically by the config. For each creation there exists a source and a target chain. 
I just realized that we need to have the option to create a A -> A EventProcessor out of the config. So let's move the above to a new issue or to the respective epic, please.

istankovic: > can we continue this discussion somewhere? I think this is important.
> ...

Reading the above, I think we are pretty much on the same page. I would have preferred to have made a step in this direction as part of this PR, but it does not really matter so much, as long as we do it at some point. Sooner or later we will run into the need for this so it is fine.

PR #1470, opened 2023-02-09
Create only one process created per resolution and invalidation

fredo:
Although we stored the future of the relayer process in a dictionary, we did not  check whether there was a process running on a new InitiateL1Resolution event.

Additionally, we also stored invalidations in `context.l1_resolutions` which would lead to invalidation relayer processes blocking l1 resolutions. we need to separate that into their own dicts.

### DIFF
--- a/beamer/state_machine.py
+++ b/beamer/state_machine.py
@@ -55,7 +55,8 @@ class Context:
     web3_l1: Web3
     task_pool: Executor
     claim_request_extension: int
-    l1_resolutions: dict[tuple[RequestId, ClaimId], Future]
+    l1_resolutions: dict[RequestId, Future]
+    l1_invalidations: dict[ClaimId, Future]
     finality_periods: dict[ChainId, int] = field(default_factory=dict)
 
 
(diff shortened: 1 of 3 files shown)


beamer/state_machine.py
```diff
@@ -438,6 +439,10 @@ def _handle_initiate_l1_resolution(
     if not _timestamp_is_l1_finalized(request.fill_timestamp, context, request.target_chain_id):
         return False, None
 
+    # There is already a process running
+    if (request.id, claim.id) in context.l1_resolutions:
+        return True, None
```
istankovic, line 444: I am wondering: why do we have an L1 resolution per `(request, claim)`?
Isn't L1 resolution, in the sense of proving who the filler was, actually per `request`?
The way I see it, once a request is L1 resolved, the result of all existing challenge games becomes known, so there is no need to do multiple L1 resolutions.

fredo: true.. I don't know why it was like it. Let me think of any edge case, if I can't think of anything, I'll change it.

fredo: I think it even removes the possibility, that two different claims for the same request would both create a relayer process

istankovic: > I think it even removes the possibility, that two different claims for the same request would both create a relayer process

Exactly.

istankovic (approved):

PR #1475, opened 2023-02-13
frontend: Fix app release version repository URL

### LINKED ISSUE #1474: Frontend: Wrong repository commit hash url when deployed on digital ocean
# Rationale
Since [we included the commit hash](https://github.com/beamer-bridge/beamer/pull/1454) in the frontend footer i figured out that we MIGHT have introduced security vulnerability when deploying the app as a DO App. 
The security vulnerability? is introduced by exposing the github_access_token that the DO App uses internally.  

The commit hash link is redirecting to this URL:
`https://x-access-token:<github_access_token>@github.com/beamer-bridge/beamer.git/tree/bc188d1`
instead of:
`https://github.com/beamer-bridge/beamer/tree/bc188d1`

2 options here:
1. We need to rethink the way we are fetching the repo URL [here](https://github.com/beamer-bridge/beamer/blob/main/frontend/vite.config.ts#L15)
2. Include the URL in the frontend code as a static string (https://github.com/beamer-bridge/beamer/tree). 

Option 1 would be the optimal approach as it is fully automated and probably wont ever require maintenance.
Option 2 requires maintenance (adjusting the static string) if we ever move the frontend to a separate repository or if the app is being ran from a forked repository.

# Acceptance Criteria
- [ ] When app is deployed as DO App it links to the HTTPS URL of our repository without exposing any sensitive information

GabrielBuragev:
closes #1474 

Please note, this solution won't work with forked repositories so it is tailored only for use with our official repo (repository URL value is a static `https://github.com/beamer-bridge/beamer` instead of a dynamic value based on `git ls-remote --get-url`).

When using `git ls-remote --get-url` for determining the repository URL we had several problems for which i couldn't find a simple yet inclusive solution:
- determining the HTTPS URLs when the local repo config is using SSH URLs
- not exposing the sensitive information stored inside `git config` (like github access tokens etc.). Check #1474 for more info.

If someone has any suggestion on how to tackle all of this at once with a simple command please let me know.

### DIFF
--- a/frontend/tests/unit/components/Version.spec.ts
+++ b/frontend/tests/unit/components/Version.spec.ts
@@ -2,29 +2,57 @@ import { mount } from '@vue/test-utils';
 
 import Version from '@/components/Version.vue';
 
-const envCopy = Object.assign({}, APP_RELEASE);
+const globalCopy = Object.assign({}, APP_RELEASE);
+const importEnvCopy = Object.assign({}, import.meta.env);
 
-const resetEnvMock = () => {
-  Object.assign(APP_RELEASE, envCopy);
+const resetImportEnv = (): void => {
+  Object.assign(import.meta.env, importEnvCopy);
+};
+
+const resetGlobalVariables = (): void => {
+  Object.assign(APP_RELEASE, globalCopy);
 };
 
 describe('Version.vue', () => {
   afterEach(() => {
-    resetEnvMock();
+    resetImportEnv();
+    resetGlobalVariables();
   });
 
-  it('is hidden when no version is defined', () => {
+  it('is hidden when in development mode', () => {
+    Object.assign(import.meta.env, {
+      MODE: 'development',
+    });
+    const wrapper = mount(Version);
+
+    const commitHashEl = wrapper.find('[data-test="version"]');
+    expect(commitHashEl.exists()).toBe(false);
+  });
+
+  it('is visible when in production mode', () => {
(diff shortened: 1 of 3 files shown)


GabrielBuragev:
@istankovic your suggested approach is used here.

frontend/tests/unit/components/Version.spec.ts
```diff
@@ -2,29 +2,57 @@ import { mount } from '@vue/test-utils';
...
+
+    const commitHashEl = wrapper.find('[data-test="version"]');
+    expect(commitHashEl.exists()).toBe(false);
   });
 
-  it('is hidden when no version is defined', () => {
+  it('is visible when in production mode', () => {
+    Object.assign(import.meta.env, {
+      MODE: 'production',
+    });
```
istankovic, line 35: Does this mean the commit hash/version won't be visible in development mode? If so, why would we hide it?

GabrielBuragev: Because it introduces a few problems and doesn't provide any value for the developers.
1. When working locally on the frontend and you are several commits ahead of upstream branch, the link will not work as that commit hash cannot be found upstream. This will not happen on staging/production deployments as we are usually deploying from our github repo.
2. I don't see any benefit of having it in dev mode. What this version string refers to is the `Build version` of the deployed app.

manuelwedler (approved):
Looks good!

PR #1479, opened 2023-02-14
Monitoring script

compojoom:
Still work in progress

- [x] current functionality is tested
- [x] agent balance through rpc
- [x] can send to telegram


scripts/shtf.py
```diff
@@ -0,0 +1,536 @@
...
+
+
+def make_web3(rpc: str) -> Web3:
+    web3 = Web3(Web3.HTTPProvider(rpc, request_kwargs=dict(timeout=5)))
+    web3.middleware_onion.inject(geth_poa_middleware, layer=0)
+    return web3
+
+
+def create_transfers_object(events_grouped_by_chain_id: ChainToEvent) -> transfersType:
+    transfers = defaultdict(dict)  # type: ignore[var-annotated]
```
compojoom, line 245: I would be happy if someone can let me know how to type transfers here as transfersType?

bilbeyt: `
from typing import cast

transfers = cast(transfersType, transfers)
`

bilbeyt: and normally new type should be also `TransfersType` instead of `transfersType`

scripts/monitoring-mainnet.conf
```diff
@@ -0,0 +1,24 @@
...
+
+[notification.telegram]
+enabled=true
+url=
+
+[chains.arbitrum]
+rpc-url=
+explorer=https://arbiscan.io/tx/
+chain-id=42161
+start-block=46158611
```
bilbeyt, line 18: maybe better to get this info in script using deployment.json ? so we dont need to change conf with every possible release

bilbeyt: same true for chain_id as you can get it with w3 object like `w3.eth.chain_id` but as it is not changing not so important

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+    REQUEST_EXPIRED = "RequestExpired"
+    DEPOSIT_WITHDRAWN = "DepositWithdrawn"
+    UNCLAIMED_FILL = "UnclaimedFill"
+    CHALLENGE_GAME = "ChallengeGame"
+    CHALLENGE_GAME_CLAIM_BY_SOMEONE_ELSE = "ChallengeGameClaimedBySomeoneElse"
+
+
+class MessageType(TypedDict):
+    text: str
+    message_link: NotRequired[str]
```
bilbeyt, line 41: maybe just Optional[str]

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+import beamer.events
+from beamer.config import _merge_dicts
+from beamer.events import ClaimMade, DepositWithdrawn, RequestCreated, RequestFilled
+from beamer.typing import ChainId
+from beamer.util import load_ERC20_abi, setup_logging
+
+logging.disable()
+
+
+class MessageTypes:
```
bilbeyt, line 31: This seems to me an Enum

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+    agent_address: str
+    deployment_dir: Path
+    notification_system: str
+    rpcs: dict[int, tuple[str, BlockNumber]]
+    explorers: dict[int, str]
+    notification: dict[str, dict[str, str]]
+    tokens: TokensConfig
+
+
+class TransferType(TypedDict):
```
bilbeyt, line 57: I didnt really get this actually. Name implies that this is just types like an enum, if not maybe we can just rename it to TransferInfo or something

scripts/shtf.py
compojoom: I first had optional, but mypy was complaining that I was missing a message_link on some places and that is why I changed it to NotRequired. 

I think optional is - it is allowed to not have a value, but the key must be present in the dict.

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+    expired_requests: int = 0
+    expired_requests_volume: int = 0
+
+
+class NotificationState:
+    def __init__(self) -> None:
+        with open(os.path.join(os.path.dirname(__file__), "breadcrumbs.json"), "r") as f:
+            self.notified_state = json.load(f)
+
+    def is_set(self, request_id: str, notified_on: str) -> bool:
```
bilbeyt, line 104: we can rename to get_breadcrumb

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+        breadcrumb = self.notified_state.get(str(request_id))
+        if breadcrumb is None:
+            return False
+        else:
+            return breadcrumb.get(notified_on, False)
```
bilbeyt, line 109: just to correct the return type you can use bool(breadcrumb.get..) as if this returns normally then it is probably returning datetime or str?

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+    agent_address: str = ""
+
+    notification_state: NotificationState = field(default_factory=NotificationState)
+
+    def add_message(self, message: NotificationType) -> None:
+        self.messages.append(message)
+
+
+class Notify:
+    def __init__(self, system: str, config: dict[str, dict[str, str]]) -> None:
```
bilbeyt, line 137: config can be a class i guess as this is hard to read

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+        self.system = system
+        self.config = config
+
+    def send(self, message: MessageType, callback: Optional[Callable] = None) -> None:
+        global notifications_sent
+        if notifications_sent > 0 and notifications_sent % 5 == 0:
```
bilbeyt, line 143: just looked at notifications_sent and it is only used in Notify class why we are using global here ?

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+        if status:
+            notifications_sent += 1
+            if callback is not None:
+                callback()
+
+    def send_to_rocketchat(self, text: str) -> bool:
+        headers = {"Content-Type": "application/json"}
+
+        body = {
+            "username": "beamer-monitoring-bot",
```
bilbeyt, line 167: is this just sending it? i cant see any authentication

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+
+        if r["ok"]:
+            return True
+
+        return False
+
+
+notifications_sent = 0
+timeout_between_sends = 61
+claimRequestExtension = 86400
```
bilbeyt, line 197: better to read this from contract actually even though this is true

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+global_config: Optional[ConfigType] = None
+
+
+def _default_config() -> dict:
+    return {"chains": {}, "tokens": {}}
+
+
+def set_config(path: Path) -> None:
+    global global_config
```
bilbeyt, line 206: maybe instead of global config you can pass this to other functions/classes in main ?

scripts/shtf.py
```diff
@@ -0,0 +1,629 @@
...
+        if values.get("created") is None:
+            del transfers[requestId]
+
+        claims = values.get("claimed")
+        if claims is not None:
+            for events in claims.values():
+                events.sort(key=lambda a: -1 if a.last_challenger == ADDRESS_ZERO else 1)
+
+
+def fetch_events() -> ChainToEvent:
```
bilbeyt, line 277: you can give the config to this as argument so we dont execute get_config that much

scripts/shtf.py
compojoom: No! 
I need the config deep in some functions and the parent functions don't need the config, so I don't want to pass around config as a parameter.

scripts/shtf.py
compojoom: You don't need authentication. This is a webhook url - anyone could send to it, if you knew the url.

bilbeyt (commented):
Partially looked to the script, there might be some other things.

scripts/shtf.py
compojoom: Fetch events is executed just once, so there is no real difference if we call get_config in it or outside of it. I could do
`config = get_config()` at the beginning of the function and it will have the same effect.

Passing the config as an argument to the function only makes it easier to test, as you could pass a different test config. (and not have to  monkeypatch the get_config)

scripts/monitoring-mainnet.conf
compojoom: It would be kind of hard to get the start_block from the deployments.json. 

First I'll have to get the chain_id for the given rpc, then look inside of the deployments for that chainid, then for the RequestManager and then block_number. 

Of course it is doable, but sounds like useless code if those values are hardly ever going to change.

scripts/shtf.py
compojoom: I would actually prefer to have call it just Transfer, but python doesn't seem to have a way to indicate that this is a type - other than adding this to the name?

In typescript I could do
type Transfer = ...
or 
interface Transfer

A beamer Transfer is composed of a created, filled, claimed and withdrawn events. If you want we can call this TransferInfo, but I don't see it as info, but as an actual Transfer :D

bilbeyt:
General thing: Circleci is only testing contracts and agents folders in tests so test you provided is not included to the pipeline. We should add it.

istankovic:
Could you please squash all the changes so that we have two commits: one for the script impl and one for the tests?

scripts/shtf.py
bilbeyt: Like i wrote to you in rocketchat, we may have RequestEventHistory as dataclass for this purpose.

scripts/monitoring-mainnet.conf
bilbeyt: Well we are doing it already in agent.py but if others say ok then it is ok.

scripts/shtf.py
bilbeyt: you mean deep copy in a function ? are you changing config in some of them ?

scripts/shtf.py
bilbeyt: I believe you can define config at start at least. We dont want to read from file everytime. Maybe a small performance increase but still.

scripts/shtf.py
compojoom: No, i don't mean deep copy. 

that analyze_transfers call anyalze_transfer on every transfer and analyze_transfer then calls
```
    check_if_request_created(transfer, ctx)
    check_if_request_has_fill(transfer, ctx)
    check_if_claim_is_made_for_fill(transfer, ctx)
    check_if_challenge_game(transfer, ctx)
    check_if_request_has_been_withdrawn(transfer, ctx)
```

and eventually on eof this will call create_expired_request_message which will call link_to_explorer which will then need the config. So I find it useless to pass the config between 5 functions only to be used by the one deepest in the stack.

scripts/shtf.py
compojoom: It's not read from file. set_config reads the file and sets it in the global_conf. From there on it's read from memory.

scripts/shtf.py
compojoom: I've moved it to the notify class now.

compojoom:
squashed.

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+class Transfer(TypedDict):
+    created: NotRequired[RequestCreated]
+    filled: NotRequired[RequestFilled]
+    claimed: NotRequired[dict[int, list[ClaimMade]]]
+    withdrawn: NotRequired[DepositWithdrawn]
+
+
+TransfersMap = DefaultDict[str, Transfer]
+
+ChainToEvent = dict[ChainId, list[beamer.events.Event]]
```
bilbeyt, line 59: Lets rename to ChainEventMap

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+
+
+class Transfer(TypedDict):
+    created: NotRequired[RequestCreated]
+    filled: NotRequired[RequestFilled]
+    claimed: NotRequired[dict[int, list[ClaimMade]]]
+    withdrawn: NotRequired[DepositWithdrawn]
+
+
+TransfersMap = DefaultDict[str, Transfer]
```
bilbeyt, line 57: Imo, this shouldnt be plural so TransferMap

istankovic:
> squashed.

Thanks!

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+
+class MessageTypes:
+    REQUEST_EXPIRED = "RequestExpired"
+    DEPOSIT_WITHDRAWN = "DepositWithdrawn"
+    UNCLAIMED_FILL = "UnclaimedFill"
+    CHALLENGE_GAME = "ChallengeGame"
+    CHALLENGE_GAME_CLAIM_BY_SOMEONE_ELSE = "ChallengeGameClaimedBySomeoneElse"
+
+
+TokensConfig = dict[str, list[tuple[str, str]]]
```
bilbeyt, line 37: TokenMap would be better naming

bilbeyt: or maybe just TokenConfig

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+    claimed: NotRequired[dict[int, list[ClaimMade]]]
+    withdrawn: NotRequired[DepositWithdrawn]
+
+
+TransfersMap = DefaultDict[str, Transfer]
+
+ChainToEvent = dict[ChainId, list[beamer.events.Event]]
+
+
+class MetaType(TypedDict):
```
bilbeyt, line 62: NotificationMeta

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+    request_id: str
+    message_type: str
+    message_link: str
+
+
+class NotificationType(TypedDict):
```
bilbeyt, line 68: Just Notification

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+    messages: list = field(default_factory=list)
+    agent_address: str = ""
+
+    notification_state: NotificationState = field(default_factory=NotificationState)
+
+    def add_message(self, message: NotificationType) -> None:
+        self.messages.append(message)
+
+
+claimRequestExtension = 86400
```
bilbeyt, line 104: snake_case usage: claim_request_extension

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+    DEPOSIT_WITHDRAWN = "DepositWithdrawn"
+    UNCLAIMED_FILL = "UnclaimedFill"
+    CHALLENGE_GAME = "ChallengeGame"
+    CHALLENGE_GAME_CLAIM_BY_SOMEONE_ELSE = "ChallengeGameClaimedBySomeoneElse"
+
+
+TokensConfig = dict[str, list[tuple[str, str]]]
+
+
+class ConfigType(TypedDict):
```
bilbeyt, line 40: just Config or MonitorConfig

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+    cleanup_transfers(transfers)
+
+    ctx = Context()
+    ctx.agent_address = get_config()["agent_address"]
+
+    analyze_transfers(transfers, ctx)
+
+
+def cleanup_transfers(transfers: TransfersMap) -> None:
+    for requestId, values in transfers.copy().items():
```
bilbeyt, line 174: request_id, values

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+            token_decimals = contract.functions.decimals().call()
+            liquidity[name][chain_id] = (
+                contract.functions.balanceOf(agent_address).call() * 1 * 10**-token_decimals
+            )
+
+    for chain_id, (rpc, _start_block) in rpcs.items():
+        web3 = make_web3(rpc)
+        agent_address = Web3.toChecksumAddress(agent_address)
+        assert ChainId(int(chain_id)) == ChainId(web3.eth.chain_id)
+        liquidity["eth"][chain_id] = web3.eth.get_balance(agent_address) * 1 * 10**-18
```
bilbeyt, line 231: are we repeating ourselves here ? the previous loops is kind of setting funds imo

scripts/monitoring/monitoring.py
```diff
@@ -0,0 +1,536 @@
...
+
+    return liquidity
+
+
+def create_transfers_object(events_grouped_by_chain_id: ChainToEvent) -> TransfersMap:
+    transfers = defaultdict(dict)  # type: ignore[var-annotated]
+
+    for _chain_id, events in events_grouped_by_chain_id.items():
```
bilbeyt, line 239: as you dont use _chain_id you can instead .values() in this way you dont need _chain_id part and you will only have events

istankovic:
@compojoom I think we should also make this script runnable the same way agent is.
We have this in our `pyproject.toml`:
```
[tool.poetry.scripts]
beamer-agent = 'beamer.cli:main'
```
I think we could extend this:
```
[tool.poetry.scripts]
beamer-agent = 'beamer.cli:main'
beamer-monitor = 'monitoring:main'
```
(Note: you may need to look into https://python-poetry.org/docs/pyproject/#packages to make this work)

By doing this, we could run the monitor script the same way we run the agent, which makes it simpler for the users.

scripts/monitoring/notify.py
```diff
@@ -0,0 +1,76 @@
...
+from typing_extensions import NotRequired
+
+
+timeout_between_sends = 61
+
+
+NotificationConfig = dict[str, dict[str, str]]
+
+
+class MessageType(TypedDict):
```
bilbeyt, line 15: can be just Message

scripts/monitoring/notification_state.py
```diff
@@ -0,0 +1,27 @@
...
+        if breadcrumb is None:
+            return False
+        else:
+            return breadcrumb.get(notified_on, False)
+
+    def update(self, request_id: str, notified_on: str) -> None:
+        breadcrumb = self.notified_state.get(str(request_id))
+        if breadcrumb is None:
+            self.notified_state[str(request_id)] = dict()
+            self.notified_state[str(request_id)][notified_on] = True
```
istankovic, line 21: Why are we calling `str(request_id)` if `request_id` is already of type `str`, per function signature?

Also, we should use dict literal `{}` instead of `dict()` when creating empty dictionaries.

bilbeyt:
We should use make lint command to check as this is now failing.

scripts/monitoring/notify.py
```diff
@@ -0,0 +1,76 @@
+import json
+import time
+from typing import Callable, Optional, TypedDict
+
+import requests
+from typing_extensions import NotRequired
+
+
+timeout_between_sends = 61
```
istankovic, line 9: Per PEP8, constants should be all uppercase. Please also consider making this private if this is not being used outside of this module.

bilbeyt:
And can you change the PR title also ?

compojoom:
@bilbeyt since when did we become that corporate 😄

beamer/cli.py
```diff
@@ -1,125 +1,13 @@
...
-        "target-chain": target_chain,
-        "account.path": account_path,
-        "account.password": account_password,
-        "unsafe-fill-time": unsafe_fill_time,
-    }
 
-    for chainspec in chain:
-        name, rpc_url = chainspec.split("=", 1)
-        options[f"chains.{name}.rpc-url"] = rpc_url
+from beamer.agent import commands as beamer_cli_commands
```
istankovic, line 3: To be more consistent, I would use
```python
from beamer.agent import commands as agent_commands
```

beamer/agent/commands.py
```diff
@@ -0,0 +1,127 @@
...
+log = structlog.get_logger(__name__)
+
+
+def _sigint_handler(agent: Agent) -> None:
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
+    log.info("Received SIGINT, shutting down")
+    agent.stop()
+
+
+@click.command("agent")
```
istankovic, line 25: I would name this file just `command.py` instead of `commands.py` because there is only going to be one "beamer agent" command, the other things in here are options. Similarly for `monitoring`.

istankovic:
Putting the comment here because... github: I would move everything under `beamer/monitoring/chain` to just `beamer/monitoring` and get rid of the `chain` subdir since it does not seem necessary anymore. It would also make it more in line with how `beamer/agent` is organized.

Additionally, how about renaming `beamer/monitoring` to `beamer/healthcheck` (in light of the recent naming discussions)?

beamer/monitoring/chain/commands.py
```diff
@@ -0,0 +1,25 @@
...
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, path_type=Path),
+    metavar="PATH",
+    help="Config file with chains configuration",
+)
+def monitor(
+    config_path: Path,
+) -> None:
+    """Beamer Chain Monitoring
+
+    Beamer Chain Monitoring is a tool for monitoring the beamer protocol and agent activity.
```
istankovic, line 22: We should probably not mention monitoring anymore, due to new naming.

breadcrumbs.json
```diff
@@ -0,0 +1 @@
...823f6128548c872137d74b7093": {"RequestExpired": true}, "0xc5e82a76293a681125c59ad069e5c872f8a09008856d73b0db7fc4bf34ce0428": {"RequestExpired": true}, "0x2085d985eac7794954d2e1f33a97b42ab86e47f5870dd42042c334dbedf85274": {"RequestExpired": true}, "0x4e2c4addd91e76e463ae64b3a66d3618b9517ce3f7f554d220734ccdfbaee86a": {"RequestExpired": true}, "0x445367f3cccd795e9b9187643c50d99f079124a7e7ddf8263fe48b0566c1592f": {"RequestExpired": true}, "0x7426863c72072c4bb2c22d63567cd029e61ee5349396772dd411db302c24bf18": {"RequestExpired": true}, "0xa250ff414ae4ef58fed118e9cff0e3f6089d28fddc457d9d0e74a6471a82d772": {"ChallengeGame": true}, "0xb4834679b759033c0ce32371c36c610d1bef1b3b661ec6df86872288cf456311": {"RequestExpired": true}, "0xc34968cc63c3a0492aa120f5226923c92b087e4261cf0930645e46a434417943": {"ChallengeGame": true}, "0x53b4b0340d143c991d25e043df4a38112316dc74593c328e71f03bd6fbe1e076": {"ChallengeGame": true}, "0xc05f08ee7edcf81d413363b2ae2235273b70cfa5699fd973f050e0455b605e7f": {"ChallengeGame": true}, "0xa943c22e6f9ecc179bbacadffc18464e4173fbdf9260c14b45fecd34cc4b7f6f": {"ChallengeGame": true}, "0xb0eebc2b2e93b8817c69232f429e48e444d9c9f9e8fa3eae635f9a448e2bf5fe": {"ChallengeGame": true}}
```
istankovic: This file should not be in the repo.

pyproject.toml
```diff
@@ -71,6 +72,10 @@ ignore_whitespace = true
 module = ["brownie.*", "twisted.*"]
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = ["beamer.tests.*"]
+disallow_untyped_defs = false
+
```
istankovic, line 78: Why did you add this? Perhaps a `rebase` artifact?

beamer/agent/commands.py
compojoom: It's one right now. But what if we add:
`beamer health-check agent-balance AGENT_ADDRESS`

tomorrow?

breadcrumbs.json
compojoom: already deleted, sorry.

pyproject.toml
compojoom: yes, somehow it ended up there.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+import beamer.agent.contracts
+import beamer.agent.events
+from beamer.agent.config import _merge_dicts
+from beamer.agent.events import ClaimMade, DepositWithdrawn, RequestCreated, RequestFilled
+from beamer.agent.typing import ChainId
+from beamer.agent.util import load_ERC20_abi, setup_logging
+from beamer.health.notification_state import NotificationState
+from beamer.health.notify import Message, NotificationConfig, Notify
+
+logging.disable()
```
istankovic, line 25: Why do you need this?

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+    messages: list = field(default_factory=list)
+    agent_address: str = ""
+
+    notification_state: NotificationState = NotificationState(Path("chain/"))
+
+    def add_message(self, message: Notification) -> None:
+        self.messages.append(message)
+
+
+claim_request_extension = 86400
```
istankovic, line 104: Why do you not query this value from the contract?

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+    DEPOSIT_WITHDRAWN = "DepositWithdrawn"
+    UNCLAIMED_FILL = "UnclaimedFill"
+    CHALLENGE_GAME = "ChallengeGame"
+    CHALLENGE_GAME_CLAIM_BY_SOMEONE_ELSE = "ChallengeGameClaimedBySomeoneElse"
+
+
+TokenMap = dict[str, list[tuple[str, str]]]
+
+
+class MonitorConfig(TypedDict):
```
istankovic, line 39: Should probably be renamed to `HealthConfig` or something like that now that we changed naming.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+    agent_address: str = ""
+
+    notification_state: NotificationState = NotificationState(Path("chain/"))
+
+    def add_message(self, message: Notification) -> None:
+        self.messages.append(message)
+
+
+claim_request_extension = 86400
+global_config: Optional[MonitorConfig] = None
```
istankovic, line 105: Please use all-uppercase for such vars:
```python
GLOBAL_CONFIG: None | MonitorConfig = None
```

Note that I also used the new, preferred union typing syntax (`A | B`).

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+    web3.middleware_onion.inject(geth_poa_middleware, layer=0)
+    return web3
+
+
+def get_agent_liquidity(
+    agent_address: str, rpcs: dict, tokens: TokenMap
+) -> dict[str, dict[ChainId, int]]:
+    contract_abi = load_ERC20_abi()
+    liquidity: dict = defaultdict(dict)
+    agent_address = Web3.toChecksumAddress(agent_address)
```
istankovic, line 211: This is weird. Please use `eth_utils.to_checksum_address`, which is commonly used elsewhere in the codebase.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+
+    for name, chain_to_token_mapping in tokens.items():
+        for chain_id, token_address in chain_to_token_mapping:
+            contract_address = Web3.toChecksumAddress(token_address)
```
istankovic, line 215: Similarly here.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+            web3 = make_web3(rpcs[int(chain_id)][0])
+            contract = web3.eth.contract(address=contract_address, abi=contract_abi)
+            token_decimals = contract.functions.decimals().call()
+            liquidity[name][chain_id] = (
+                contract.functions.balanceOf(agent_address).call() * 1 * 10**-token_decimals
```
istankovic, line 220: Nitpick: you don't need the multiplication by `1`.
I would also split the expression for readability e.g.
```python
            balance = contract.functions.balanceOf(agent_address).call()
            liquidity[name][chain_id] = balance * 10**-token_decimals
```

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+            )
+
+    for chain_id, (rpc, _start_block) in rpcs.items():
+        web3 = make_web3(rpc)
+        agent_address = Web3.toChecksumAddress(agent_address)
```
istankovic, line 225: Please use the function from `eth_utils`.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+        assert ChainId(int(chain_id)) == ChainId(web3.eth.chain_id)
+        liquidity["eth"][chain_id] = web3.eth.get_balance(agent_address) * 1 * 10**-18
```
istankovic, line 227: Nitpick: unneeded multiplication by `1`.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+
+                transfers[x.request_id.hex()]["claimed"] = claims
+
+    return cast(TransferMap, transfers)
+
+
+def get_block(chain_id: int, block_number: int) -> BlockData:
+    rpc = get_config()["rpcs"][chain_id][0]
+
+    web3 = make_web3(rpc)
```
istankovic, line 259: Please consider reusing `Web3` instances as this is making an entirely new connection to the RPC each time you request a block and may be inefficient. I am fine with it if it would be a bigger change, so it isn't critical for this PR, but at least we should have it in mind.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+def link_to_explorer(chain_id: int, tx_hash: str) -> str:
+    return f"{get_config()['explorers'][chain_id]}{tx_hash}"
+
+
+def analyze_transfer(transfer: Transfer, ctx: Context) -> None:
+    check_if_request_created(transfer, ctx)
+    check_if_request_has_fill(transfer, ctx)
+    check_if_claim_is_made_for_fill(transfer, ctx)
+    check_if_challenge_game(transfer, ctx)
+    check_if_request_has_been_withdrawn(transfer, ctx)
```
istankovic, line 273: This is nice!

beamer/health/check.py
compojoom: Actually, I don't know why I have get_block anymore. It seems that I don't use it anywhere.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+
+def check_if_request_has_fill(transfer: Transfer, ctx: Context) -> None:
+    if "created" not in transfer:
+        return
+
+    if "filled" in transfer:
+        ctx.stats.fills += 1
+        ctx.stats.volume += transfer["created"].amount
+    else:
+        if datetime.fromtimestamp(transfer["created"].valid_until) < datetime.now():
```
istankovic, line 289: Why not use just

```python
        if transfer["created"].valid_until < time.time():
```
?

Also in other places where you use `datetime`.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+
+    request = transfer["created"]
+    fill = transfer["filled"]
+    claim_events = transfer["claimed"]
+
+    agent_participating = False
+    for claims in claim_events.values():
+        if any(
+            ctx.agent_address in [obj.claimer.lower(), obj.last_challenger.lower()]
+            for obj in claims
```
istankovic, line 341: `ctx.agent_address` is a checksummed address, right?
If so, will the expression `ctx.agent_address in [obj.claimer.lower(), obj.last_challenger.lower()]` ever be true?

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+        ):
+            agent_participating = True
+
+    agent_filled = fill.filler == ctx.agent_address
+
+    if ctx.agent_address and not (agent_filled or agent_participating):
+        return
```
istankovic, line 348: Why do you check the boolean value of `ctx.agent_address` here?

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+Total claims in network: {ctx.stats.claims}
+Total withdrawals in network: {ctx.stats.withdrawals}
+Total total volume in network: {ctx.stats.volume * 1e-6} USDC
+Total expired requests volume in network: {ctx.stats.expired_requests_volume * 1e-6} USDC
+    """,  # noqa: E501
+    }
+
+    notify = Notify(config["notification_system"], config["notification"])
+
+    notify.send(message)
```
istankovic, line 423: Nitpick: extraneous empty line above.

beamer/health/check.py
```diff
@@ -0,0 +1,526 @@
...
+TX_Hash `{last_claim.tx_hash.hex()}`
+       """,  # noqa: E501
+    }
+
+
+def process_messages(ctx: Context) -> None:
+    notify = Notify(get_config()["notification_system"], get_config()["notification"])
+    for message in ctx.messages:
+
+        def callback(mess: Notification = message) -> None:
```
istankovic, line 519: This looks strange. `mess` is supposed to be a message, yet the type is `Notification`?

beamer/health/health-check-mainnet.conf
```diff
@@ -0,0 +1,34 @@
...
+explorer=https://arbiscan.io/tx/
+chain-id=42161
+start-block=46158611
+
+[chains.optimism]
+rpc-url=
+explorer=https://optimistic.etherscan.io/tx/
+chain-id=10
+start-block=49645351
+
```
istankovic, line 26: Why do we need `start-block` in the config file?

beamer/health/notification_state.py
```diff
@@ -0,0 +1,34 @@
+import json
+import os
+from pathlib import Path
+
+
+class NotificationState:
```
istankovic, line 6: I think we can just move everything from this file to `beamer.health.notify` and remove this module. It's related to notifications anyway.

beamer/health/notification_state.py
```diff
@@ -0,0 +1,34 @@
...
+                self.notified_state = json.load(f)
+        except FileNotFoundError:
+            self.notified_state = {}
+
+    def is_set(self, request_id: str, notified_on: str) -> bool:
+        breadcrumb = self.notified_state.get(str(request_id))
+        if breadcrumb is None:
+            return False
+        else:
+            return breadcrumb.get(notified_on, False)
```
istankovic, line 23: Nitpick: could be more concisely written:
```python
return breadcrumb is not None and breadcrumb.get(notified_on, False)
```

beamer/health/notification_state.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def update(self, request_id: str, notified_on: str) -> None:
+        breadcrumb = self.notified_state.get(request_id)
+        if breadcrumb is None:
+            self.notified_state[request_id] = {notified_on: True}
+        else:
+            breadcrumb[notified_on] = True
```
istankovic, line 30: If `breadcrumb` is just a `dict` that maps to a `bool`, consider instead using a `set` -- from this I imagine you only need membership testing, and that's what's `set` is made for.

beamer/health/notification_state.py
```diff
@@ -0,0 +1,34 @@
...
+    cache_path: Path
+
+    def __init__(self, cache_path: Path) -> None:
+        self.cache_path = cache_path
+
+        try:
+            with open(os.path.join(self.cache_path, "breadcrumbs.json"), "r") as f:
+                self.notified_state = json.load(f)
+        except FileNotFoundError:
+            self.notified_state = {}
```
istankovic, line 16: Please consider making `notified_state` private if it is not used outside of this class.

beamer/health/notify.py
```diff
@@ -0,0 +1,76 @@
...
+        self.config = config
+
+    def send(self, message: Message, callback: Optional[Callable] = None) -> None:
+        if self.notifications_sent > 0 and self.notifications_sent % 5 == 0:
+            time.sleep(_TIMEOUT_BETWEEN_SENDS)
+
+        text = message["text"]
+
+        if message.get("message_link"):
+            text += "\n" + f"View on [explorer]({message['message_link']})"
```
istankovic, line 34: You can save one string concatenation operation by including the `\n` in the string:
```python
            text += f"\nView on [explorer]({message['message_link']})"
```

beamer/health/notify.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if self.system == "rocketchat":
+            status = self.send_to_rocketchat(text)
+        elif self.system == "telegram":
+            status = self.send_to_telegram(text)
+        else:
+            raise ValueError("Unknown notification system")
```
istankovic, line 41: Nitpick: this is totally fine, but if you haven't know about the new `match` feature:

```python
match self.system:
        case "rocketchat":
            status = self.send_to_rocketchat(text)
        case "telegram":
            status = self.send_to_telegram(text)
        case _:
            raise ValueError("Unknown notification system")
```

beamer/health/notify.py
```diff
@@ -0,0 +1,76 @@
...
+        r = requests.post(self.config["rocketchat"]["url"], data=json.dumps(body), headers=headers)
+
+        if r.status_code == 200:
+            return True
+
+        return False
+
+    def send_to_telegram(self, text: str) -> bool:
+
+        token = self.config["telegram"]["token"]
```
istankovic, line 67: Nitpick: extraneous empty line

beamer/health/notify.py
```diff
@@ -0,0 +1,76 @@
...
+            "icon": ":heart",
+            "text": text,
+        }
+
+        r = requests.post(self.config["rocketchat"]["url"], data=json.dumps(body), headers=headers)
+
+        if r.status_code == 200:
+            return True
+
+        return False
```
istankovic, line 63: This can be more concisely written as

```python
return r.status_code == 200
```

beamer/health/notify.py
```diff
@@ -0,0 +1,76 @@
...
+        token = self.config["telegram"]["token"]
+        chat_id = self.config["telegram"]["chat_id"]
+
+        url = f"https://api.telegram.org/bot{token}/sendMessage?chat_id={chat_id}&text={text}"
+        r = requests.get(url).json()
+
+        if r["ok"]:
+            return True
+
+        return False
```
istankovic, line 76: This can be more concisely written as
```python
return bool(r["ok"])
```

beamer/tests/health/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+@pytest.fixture(autouse=True)
+def get_config(monkeypatch):
+    """Set default config for script."""
+    monkeypatch.setattr("beamer.health.check.get_config", lambda: config)
+
+
+@pytest.fixture
+def ctx():
+    ctx = Context(agent_address=AGENT_ADDRESS)
+    ctx.notification_state = NotificationState(Path("./"))
```
istankovic, line 43: This will leave `breadcrumbs.json` wherever the current working directory happens to be, right?
Wouldn't it be better if we used a temporary directory that gets cleaned up after the tests are done?
Please take a look at https://docs.pytest.org/en/6.2.x/tmpdir.html#temporary-directories-and-files.

beamer/tests/health/test_check.py
```diff
@@ -0,0 +1,313 @@
...
+
+def test_check_if_request_has_fill_with_no_fill_and_request_created_not_expired(ctx):
+    transfer: Transfer = {
+        "created": dataclasses.replace(request, valid_until=Termination(int(time.time() + 1000))),
+    }
+
+    check_if_request_has_fill(transfer, ctx)
+
+    assert ctx.stats.fills == 0
+    assert ctx.stats.expired_requests == 0
```
istankovic, line 133: I'm going to make a general comment regarding testing here. I find this kind of granular testing a bad approach in terms of maintainability and developer productivity facilitation. Such "unit testing" does not make us more confident in our code than alternative approaches (outlined below). It just forces exposing all the internal check methods and, instead of helping future changes, makes it hard to modify anything because of inevitable test breakages. For an excellent overview of what is wrong with this, and general with modern-day testing, I can recommend the following talk by Ian Cooper: https://www.youtube.com/watch?v=EZ05e7EMOLM.

An alternative approach:
- have a well-defined, public entry point e.g. `analyze_transfers` -- this would be your "unit" of testing
- feed the entry point with various input -- these would be different test cases/functions
- check the resulting output to make sure it is correct

In this way, you do not need to care about the particular implementation of the checks, such as function names, arguments etc. What you care about is _functional correctness_ i.e. that your checks perform the job they were designed to do. This achieves the same effect, however it almost completely decouples the implementation from functionality and allows for far easier changes and refactorings in the future.

istankovic (changes requested):
Epic work. There are still things to be worked and improved upon, but it is getting there.

beamer/health/check.py
compojoom: when I create the global config I set it to .lower(). So it should be triggered.

beamer/health/check.py
compojoom: agent_address is provided through the config. If you don't provide an agent_address - which we will do in our own cron - then we will get notification no matter who the involved party is. 
But if the agent has entered an agent_address -> then they only want to see events in which they are participating.

beamer/health/health-check-mainnet.conf
compojoom: because I'm too lazy to fetch the start block from the chain. :(

beamer/health/notification_state.py
compojoom: yes, but what is concisely for you is basically unreadable for me, because I don't live in python :)

beamer/health/notification_state.py
compojoom: the problem is that for the same requestId -> I could have several notifications. I wouldn't want to notify the user for missed claim twice, but if the claim was missed, then made and then someone challenged - I sure want to have the challenge notification sent.

beamer/tests/health/conftest.py
compojoom: Well, no. Because I have to explicitly call notification_state.persist for the breadcrumbs file to be written do the filesystem and currently the tests in the PR don't test this part of the code.

beamer/health/notification_state.py
istankovic: > the problem is that for the same requestId -> I could have several notifications. I wouldn't want to notify the user for missed claim twice, but if the claim was missed, then made and then someone challenged - I sure want to have the 
challenge notification sent.

I do not see how using a `set` here prevents that. Could you please explain this in a bit more detail?

The way I understand, `update` just sets the value for the passed-in `notified_on` key to `True`:
```python
 def update(self, request_id: str, notified_on: str) -> None:
     breadcrumb = self.notified_state.get(request_id)
     if breadcrumb is None:
         self.notified_state[request_id] = {notified_on: True}
     else:
         breadcrumb[notified_on] = True
```

Functionally, this is completely equivalent to this set-based impl:
```python
 def update(self, request_id: str, notified_on: str) -> None:
     breadcrumb = self.notified_state.get(request_id)
     if breadcrumb is None:
         self.notified_state[request_id] = {notified_on}
     else:
         breadcrumb.add(notified_on)
```

I see no benefit of using a `dict` because you are not using the value part of the `dict` in any way: if an entry exists, the value is always `True`. And actually, that was precisely the pattern used in Python before sets were introduced; people just used a `dict` with the value typically always being `True` or `None`, i.e. a value that is simply ignored.

beamer/health/health-check-mainnet.conf
compojoom: yes, yesterday while playing spike-ball I also was struck by enlightenment and realized that I could take all this form the deployment info and don't need an RPC call. Thanks. Will look into this.

beamer/tests/health/conftest.py
istankovic: > Well, no. Because I have to explicitly call notification_state.persist for the breadcrumbs file to be written do the filesystem and currently the tests in the PR don't test this part of the code.

Ah, I see. I'm ok with not having tests for that part just yet,
but I would still prefer to use a temporary path here, like
```python
@pytest.fixture
def ctx(tmp_path):
    ctx = Context(agent_address=AGENT_ADDRESS)
    ctx.notification_state = NotificationState(tmp_path)
```

beamer/health/notification_state.py
compojoom: How about:
```
        if breadcrumb is None:
            return False
        
        return breadcrumb.get(notified_on, False)
```
?

istankovic: Perfect. :+1:

beamer/health/commands.py
```diff
@@ -0,0 +1,37 @@
...
+    "config_path",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, path_type=Path),
+    metavar="PATH",
+    help="Config file with chains configuration",
+)
+@click.option(
+    "--log-level",
+    type=click.Choice(("debug", "info", "warning", "error", "critical")),
+    help="The log level. Default: info",
+    default="error",
```
istankovic, line 23: `help` and `default` are in disagreement, it seems.

beamer/health/check.py
compojoom: Hm, maybe I have some kind of mix up in the whole class. 

In my head it all comes from Notification message / Push Notification. 

On iOS/Android a push notification is a data object sent to a device. This object contains params such as:

title
text
but also might contain meta information such as -> badge_icon, sound, content_id etc. 

So originally I started up with a message - which in the context of rocketchat is just a text, but then we needed some more meta information and I went with Notification. So now the callback expects a notification, because writing what has been sent needs the request_id,, but the actual sending needs the text :D

Another funny thing that I realize right now is that on my context I do
`ctx.add_message` and then call one of `create...message` functions, but they all actually return a `Notification`, so a more correct naming would be: `add_notification`, `create...notification`

scripts/shtf.py
compojoom: We discussed this, but want to also post here - just for the protocol. I think that I don't want to use Enum here as it might be the pythonic way, but I find it the moronic way :D :D :D 

If I make this class a Enum - then as far as I understand I can't directly access the value under `MessageTypes.REQUEST_EXPIRED`, but now have to do `MessageTypes.REQUEST_EXPIRED.value` and I don't think I earn anything by doing this?

Or the other possibility `def __repr__(self): return self.value` on the enum class in order for it to behave like a class... 

I think I can live with this less than ideal Class over here :)

beamer/health/check.py
```diff
@@ -0,0 +1,514 @@
...
+from beamer.health.notify import Message, NotificationConfig, Notify, NotificationState
+from eth_utils import to_checksum_address
+
+
+class NotificationTypes:
+    REQUEST_EXPIRED = "RequestExpired"
+    DEPOSIT_WITHDRAWN = "DepositWithdrawn"
+    UNCLAIMED_FILL = "UnclaimedFill"
+    CHALLENGE_GAME = "ChallengeGame"
+    CHALLENGE_GAME_CLAIM_BY_SOMEONE_ELSE = "ChallengeGameClaimedBySomeoneElse"
```
istankovic, line 30: Not sure whether this is intentional, but I notice `CLAIM_BY` on the left and `ClaimedBy` on the right. Other types have the same expression on left and right.

pyproject.toml
```diff
@@ -81,10 +86,6 @@ strict_equality = true
 disallow_untyped_defs = true
 explicit_package_bases = true
 
-[[tool.mypy.overrides]]
-module = ["beamer.tests.*"]
-disallow_untyped_defs = false
-
```
istankovic, line 87: It seems we have this rebase artifact again. Could you please remove this and the related `tool.mypy.overrides` hunk above so your changes don't affect those?

beamer/health/notify.py
```diff
@@ -0,0 +1,105 @@
...
+        if status:
+            self.notifications_sent += 1
+            if callback is not None:
+                callback()
+
+    def send_to_rocketchat(self, text: str) -> bool:
+        headers = {"Content-Type": "application/json"}
+
+        body = {
+            "username": "beamer-monitoring-bot",
```
istankovic, line 88: A really small nitpick: maybe update the `monitoring` part since we moved away from that wording...

beamer/tests/agent/test_rate_limiter.py
```diff
@@ -6,7 +6,7 @@
 import structlog.testing
 
 import beamer.agent
-from beamer.agent import Agent
+from beamer.agent.agent import Agent
```
istankovic, line 9: This looks funny/silly as hell, but oh well. :sweat_smile:

beamer/tests/health/conftest.py
```diff
@@ -0,0 +1,43 @@
+import pytest
+
+from beamer.health.check import Context
+from beamer.health.notify import NotificationState
+from beamer.tests.health.test_check import AGENT_ADDRESS
```
istankovic, line 5: `conftest.py` is a configuration file for `pytest` and as such runs fairly early in the testing session and other tests actually depend on `conftest.py` things, so we should not introduce a reverse dependency here.

My suggestion is to move `AGENT_ADDRESS` here and make it either a package- or a session-scoped fixture,
e.g.

```python
@pytest.fixture(scope="session")
def agent_address():
    return make_address()
```

beamer/tests/health/test_check.py
```diff
@@ -0,0 +1,307 @@
...
+    fill_id=FillId(b"1"),
+    source_chain_id=SOURCE_CHAIN_ID,
+    target_token_address=TARGET_TOKEN_ADDRESS,
+    filler=AGENT_ADDRESS,
+    amount=TokenAmount(100),
+    block_number=BLOCK_NUMBER,
+    tx_hash=HexBytes(b"1"),
+)
+
+claim = ClaimMade(
```
istankovic, line 62: `request`, `fill` and `claim` should be all-uppercase to be consistent with the rest of module-level vars.

pyproject.toml
compojoom: I would really like to understand how this ends up there.

beamer/tests/health/test_notify.py
```diff
@@ -0,0 +1,32 @@
+from beamer.health.notify import NotificationState
+
+
+def test_value_is_set(tmpdir):
+    state = NotificationState(tmpdir)
+
+    state.update("a", "RequestExpired")
+
+    assert state.is_set("a", "RequestExpired") is True
+    assert state.is_set("b", "RequestExpired") is False
```
istankovic, line 10: Please do not compare directly to literal booleans (PEP8):

```suggestion
    assert state.is_set("a", "RequestExpired")
    assert not state.is_set("b", "RequestExpired")
```

beamer/tests/health/test_notify.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def test_value_is_set_on_already_existing_breadcrumb(tmpdir):
+    state = NotificationState(tmpdir)
+
+    state.update("a", "RequestExpired")
+    state.update("a", "UnclaimedFill")
+
+    assert state.is_set("a", "RequestExpired") is True
+    assert state.is_set("a", "UnclaimedFill") is True
```
istankovic, line 20: Same here, as above. Strange that none of our linting tools are complaining about it.

beamer/health/notification_state.py
istankovic: > I decided that I won't got for a set here. I wrote the tests for the class and I was able to exchange dict for set, but it turns out that since I use json.dumps I can't use set as set is not JSON.serializable and I'll have to resort to pickle.

Yeah, that's unfortunate, but what is wrong with converting the set to a list just before serialization, and converting the list to a set after you deserialize JSON?

I imagine something like
```python
json.dump(list(self._notified_state), f)
```
and 
```python
self._notified_state = set(json.load(f))
```

beamer/tests/agent/test_rate_limiter.py
compojoom: Yes, though the same. Is there a better way?

istankovic: Nah, with the current setup, don't think so. But it's fine, just a bit funny to look at.

istankovic (approved):
Looks great!
Thanks a lot for the effort! :purple_heart:

Issue #1439, opened 2023-01-31
Max Amount Error - Report from Community


DocSmoove:
#### System information

Below is the video that the community member noticed. When a user clicks "max amount" and then deletes the amount, an error appears. 
Uploading 010101011 (1).mov…

#### Expected behavior
The error should not appear
@GabrielBuragev

andfletcher:
@DocSmoove can you reupload the video and wait a bit longer. You created the issue without the video being uploaded.

DocSmoove:
https://user-images.githubusercontent.com/107990484/216257746-3b2c1e75-33d8-4354-8262-09108a5eef7a.mov

Ah, my bad. Here is the uploaded video. @andfletcher @GabrielBuragev

PR #1482, opened 2023-02-14
Frontend: improve use max button

### LINKED ISSUE #1206: Frontend: Rephrase "Use Max" button & related tooltip content
# Rationale
We need to provide a more intuitive label for the "Use Max" button & it's related tooltip text.
Currently it is not so clear that the button is clickable.

image

---

TLDR; of the discussion in this issue and what needs to be done:
- [ ] Follow sushiswap's approach - change the text inside the button to "Balance: xxx USDC" and leave the `click` functionality.
image

- [ ] Change button tooltip text to `You have {{tokenAmount}}  {{ tokenSymbol }} in your wallet`

GabrielBuragev:
closes #1206 #1439 

New behavior when users wallet balance is higher than the transfer limit:

https://user-images.githubusercontent.com/18466333/218754464-5d88d999-f67e-4f88-bbbb-c4c1a7135f13.mov

---
~~@olympe-silve i rephrased the button & tooltip and added some color adjustments to the tooltip text so things are more visible.~~

### DIFF
--- a/frontend/src/components/RequestSourceInputs.vue
+++ b/frontend/src/components/RequestSourceInputs.vue
@@ -273,6 +273,8 @@ const totalRequestTokenAmount = computed(() => {
   return totalAmount;
 });
 
+watch(totalRequestTokenAmount, () => v$.value.selectedTokenAmount?.$touch());
+
 const inputValues: WritableComputedRef<RequestSource> = computed({
   get: () => ({
     amount: selectedAmount.value,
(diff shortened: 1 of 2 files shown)


olympe-silve:
@GabrielBuragev Thank you ! 
I think we should avoid using red text against teal like that. Could you keep it all in light teal for the moment ? I have to review the colors and maybe add one or two to fix this chronic readability issue.

GabrielBuragev:
@olympe-silve Thank you for your input. Yeah i tried to find a color that would indicate a warning or an info but couldn't find any suitable. Will revert back to the other one.

Revisiting the coloring scheme and possibly extending it would be awesome. I would happily help you if needed.

compojoom (approved):
Haven't tested the frontend, but code-wise LGTM.

PR #1483, opened 2023-02-14
agent: add multiple event processors

bilbeyt:


beamer/state_machine.py
```diff
@@ -523,13 +560,24 @@ def on_future_done(f: Future) -> None:
...
         claim.l1_invalidate()
 
-        log.info("Initiated L1 invalidation", request=request, claim=claim)
+        log.info(
+            "Initiated L1 invalidation",
+            request=request,
+            claim=claim,
+            source_chain_id=context.source_chain_id,
+            target_chain_id=context.target_chain_id,
+        )
```
istankovic, line 580: Please use the binding feature of `structlog`, which we already use elsewhere:
```python
self._log = structlog.get_logger(type(self).__name__).bind(chain_id=self._chain_id)
```
Bound values will be emitted with all the `log` calls, i.e. they are implicitly output. This would save us some lines and improve readability and maintenance.

istankovic (changes requested):
Just did a partial, quick pass and spotted some things.

beamer/agent.py
```diff
@@ -36,103 +39,166 @@ def __init__(self, config: Config):
...
+            w3 = make_web3(rpc_url, self._config.account)
+            chain_id = ChainId(w3.eth.chain_id)
+            if chain == "l1":
+                w3.middleware_onion.add(latest_block_based_cache_middleware)
+            contracts_info = _get_contracts_info(self._config, w3)
+            contracts = make_contracts(w3, contracts_info)
+            request_manager = contracts.get("RequestManager")
+            fill_manager = contracts.get("FillManager")
+            event_monitor = None
+            if request_manager and fill_manager:
```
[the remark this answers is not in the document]
bilbeyt: with l1 we dont have RequestManager or FillManager

beamer/agent.py
```diff
@@ -36,103 +39,166 @@ def __init__(self, config: Config):
         self._config = config
         self._stopped = threading.Event()
         self._stopped.set()
+        self._cache: dict[str, Any] = {}
```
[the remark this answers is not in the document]
bilbeyt: it is just simple like this as i dont need to pass parameters to other things

beamer/agent.py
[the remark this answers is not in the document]
istankovic: > with l1 we dont have RequestManager or FillManager

Ah, so once we move l1 out of the way, those would also disappear. :+1:

beamer/agent.py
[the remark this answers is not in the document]
istankovic: > it is just simple like this as i dont need to pass parameters to other things

It may be simpler in terms of fewer arguments being passed around, but it makes it harder to reason about and understand the code and makes it harder to reason about the lifetime of those objects. It also prevents the garbage collector from collecting those objects for the duration of the agent's life. Since, as far as I understand, we only need those objects during initialization, we should not retain them for the entire life of the agent.

beamer/agent.py
```diff
@@ -31,107 +37,152 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
+    def _init(self) -> None:
+        # Just add one worker, as this effectively serializes the work
+        # This is necessary as we use the account for all resolutions and
+        # would otherwise run into nonce problems
+        self._task_pool = ThreadPoolExecutor(max_workers=1)
+        self._event_processors: dict[str, EventProcessor] = {}
         self._event_monitors: dict[ChainId, EventMonitor] = {}
+        l1 = self._init_l1_chain()
+        rollups = self._init_rollups()
+        rollup_chain_ids = [chain_id for chain_id in rollups]
```
fredo, line 157: couldnt you just do 

`rollups.keys()` (if you want to have it in a list `list(rollups.keys())`

beamer/chain.py
```diff
@@ -112,14 +116,22 @@ def __init__(self, context: Context):
...
         # 1 = one of the chains was synced, if there is only one chain, sync done.
         # 2 = If there are 2 different chains, the sync is done
         self._num_syncs_done = 0
         self._context = context
         self._chain_ids = {self._context.source_chain_id, self._context.target_chain_id}
+        global log
+        log = log.bind(
+            source_chain_id=self._context.source_chain_id,
+            target_chain_id=self._context.target_chain_id,
+        )
```
istankovic, line 134: This won't work -- you're changing the global `log` object, which will not have the intended effect: all the event processors will use the same log object and therefore the same pair of chain IDs. This needs to be per-instance,
e.g.
```python
self._log = log.bind(...)
```

Additionally, could you please move these logging-related changes into their own, separate commit, to make the whole PR easier to review?

bilbeyt: This was aiming to affect all parts like proceess_request and process_claims that they are not part of event processor and if i dont have logger in context itself it wont work with bind

beamer/config.py
```diff
@@ -20,9 +20,7 @@ class ConfigError(Exception):
 class Config:
     account: LocalAccount
     deployment_info: DeploymentInfo
-    l1_rpc_url: URL
-    l2a_rpc_url: URL
-    l2b_rpc_url: URL
+    chains: dict[str, URL]
```
istankovic, line 23: I would rename this attribute to something like `rpc_urls` -- `chains` is misleading.

beamer/agent.py
```diff
@@ -31,107 +37,152 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
+        rollups: dict[ChainId, Rollup] = {}
+        for chain_name, rpc_url in self._config.chains.items():
+            if chain_name == "l1":
+                continue
+            w3 = make_web3(rpc_url, self._config.account)
+            chain_id = ChainId(w3.eth.chain_id)
+            contracts_info = _get_contracts_info(self._config, w3)
+            contracts = make_contracts(w3, contracts_info)
+            request_manager = contracts.get("RequestManager")
+            fill_manager = contracts.get("FillManager")
```
istankovic, line 78: Why use `get` here instead of `[]`? We should always have these contracts and, if not, we should crash with an exception.

beamer/agent.py
```diff
@@ -31,107 +37,152 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
+        chain_id = ChainId(l1_w3.eth.chain_id)
+        return Chain(w3=l1_w3, id=chain_id, rpc_url=self._config.chains["l1"])
+
+    def _init_rollups(self) -> dict[ChainId, Rollup]:
+        rollups: dict[ChainId, Rollup] = {}
+        for chain_name, rpc_url in self._config.chains.items():
+            if chain_name == "l1":
+                continue
+            w3 = make_web3(rpc_url, self._config.account)
+            chain_id = ChainId(w3.eth.chain_id)
```
istankovic, line 74: What if there's already an entry in `rollups` with this `chain_id`?

beamer/agent.py
```diff
@@ -31,107 +37,154 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
             raise RuntimeError("Agent address is not whitelisted on RequestManager")
 
-        claim_request_extension = source_request_manager.functions.claimRequestExtension().call()
+    def _setup_direction(
+        self, direction: tuple[ChainId, ChainId], rollups: dict[ChainId, Rollup], l1: Chain
+    ) -> None:
+        source_chain = rollups[direction[0]]
+        target_chain = rollups[direction[1]]
+
+        self._check_rollup(source_chain, target_chain)
```
istankovic, line 118: I think we should split `_check_rollup` into two: `_check_source_rollup` and `_check_target_rollup`.
(Here we are providing both, which reads very weirdly.)
Then we could have
```python
        self._check_source_rollup(source_chain)
        self._check_target_rollup(source_chain)
```
Actually, even better would be 
```python
        self._check_source_chain(source_chain)
        self._check_target_chain(target_chain)
```
(this would require renaming `Rollup` to `Chain` and the base type `Chain` to something else (`ChainBase`?).

I feel the chain would also be more general since we're going to support Ethereum L1 in the short term and that is not really a rollup, so the "chain" naming would be more appropriate.

beamer/agent.py
```diff
@@ -31,107 +37,154 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
+        # This is necessary as we use the account for all resolutions and
+        # would otherwise run into nonce problems
+        self._task_pool = ThreadPoolExecutor(max_workers=1)
+        self._event_processors: dict[str, EventProcessor] = {}
         self._event_monitors: dict[ChainId, EventMonitor] = {}
+        l1 = self._init_l1_chain()
+        rollups = self._init_rollups()
+        rollup_chain_ids = list(rollups.keys())
+        if len(rollup_chain_ids) == 1:
+            rollup_chain_ids.append(rollup_chain_ids[0])
```
istankovic, line 164: Why do we need this?

bilbeyt: we dont block people who has only one chain so if that is the case there should be one event processor in the direction of A->A which is happening in our tests actually with ganache being the only chain.

beamer/agent.py
```diff
@@ -31,107 +37,154 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
+            direction = cast(tuple[ChainId, ChainId], direction)
+            self._setup_direction(direction, rollups, l1)
 
     def start(self) -> None:
         assert self._stopped.is_set()
-        beamer.metrics.init(self._config)
-        self._event_processor.start()
+        for event_processor in self._event_processors.values():
+            request_manager = event_processor._context.request_manager
+            fill_manager = event_processor._context.fill_manager
```
istankovic, line 176: We should not access private variables like this (`._context`).

beamer/metrics.py
```diff
@@ -19,8 +19,8 @@ def init(config: Any) -> None:
     info.info(
         dict(
             account=config.account.address,
-            l2a_rpc_url=config.l2a_rpc_url,
-            l2b_rpc_url=config.l2b_rpc_url,
+            l2a_rpc_url=source_rpc_url,
+            l2b_rpc_url=target_rpc_url,
```
istankovic, line 23: We should also rename `l2a_rpc_url` and `l2b_rpc_url` to `source_rpc_url` and `target_rpc_url`, respectively.

beamer/state_machine.py
```diff
@@ -454,9 +454,9 @@ def _handle_initiate_l1_resolution(
     if _l1_resolution_threshold_reached(claim, context):
         future = context.task_pool.submit(
             run_relayer_for_tx,
-            context.config.l1_rpc_url,
-            context.config.l2b_rpc_url,
-            context.config.l2a_rpc_url,
+            context.config.rpc_urls["l1"],
+            context.fill_manager.web3.provider.endpoint_uri,  # type: ignore
+            context.request_manager.web3.provider.endpoint_uri,  # type: ignore
```
istankovic, line 460: Why do we need `type: ignore` here?

bilbeyt: provider is type BaseProvider which has no endpoint_uri normally. either cast it to HTTPProvider which actually it is or just ignore so i ignored

beamer/agent.py
```diff
@@ -31,107 +37,154 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
+    id: ChainId
+    rpc_url: URL
+
+
+@dataclass
+class Rollup(Chain):
+    contracts_info: dict[str, ContractInfo]
+    contracts: dict[str, Contract]
+    request_manager: RequestManager
+    fill_manager: FillManager
```
istankovic, line 52: Both of these two dataclass types should be private, because they won't be used outside of this module.

beamer/metrics.py
bilbeyt: this may create some problems in prometheus if we are using it with some calculations etc.

istankovic: > this may create some problems in prometheus if we are using it with some calculations etc.

Our Prometheus instance is no longer tracking beamer-agent so nothing will break.
I also seriously doubt that any of our agents use this.

bilbeyt: i guess it is not a problem as this is not posting data to a central prometheus or so

beamer/agent.py
fredo: nitpick: Do we really need to store the request_manager and fill_manager on the class level? A property annotation could be enough as they are stored in contracts, aren't they?

beamer/agent.py
fredo: would it make more sense to explicitly allow same chain pairs by an option for the agent?

bilbeyt: that would be another feature actually adding an option, all i have done here is exceptional case if the chain is only 1, if there would be an option then A->A,  B->B, C->C should be added.

bilbeyt: Otherwise there wouldnt be any event processor and direction.

beamer/state_machine.py
```diff
@@ -61,6 +61,14 @@ class Context:
     l1_invalidations: dict[ClaimId, Future]
     finality_periods: dict[ChainId, int] = field(default_factory=dict)
 
+    @property
+    def source_rpc_url(self) -> URL:
+        return URL(self.request_manager.web3.provider.endpoint_uri)  # type: ignore
```
istankovic, line 67: Why not cast the provider to `HTTPProvider` and remove the `# type: ignore` thing?

beamer/state_machine.py
```diff
@@ -61,6 +61,14 @@ class Context:
...
+
+    @property
+    def target_rpc_url(self) -> URL:
+        return URL(self.fill_manager.web3.provider.endpoint_uri)  # type: ignore
```
istankovic, line 71: Same here.

beamer/tests/agent/test_request.py
```diff
@@ -106,20 +106,26 @@ def test_challenge_own_claim(config, request_manager, token):
         int(time.time()),
     )
     # Add context so that maybe_challenge verifies that the claim is not expired
-    agent.context.requests.add(request.id, request)
-    agent.context.latest_blocks[brownie.chain.id] = {"timestamp": Timestamp(0)}
+    agent._event_processors[direction]._context.requests.add(request.id, request)
+    agent._event_processors[direction]._context.latest_blocks[brownie.chain.id] = {
```
istankovic, line 110: Access to private variables.

beamer/tests/agent/test_request.py
```diff
@@ -106,20 +106,26 @@ def test_challenge_own_claim(config, request_manager, token):
...
+        "timestamp": Timestamp(0)
+    }
 
-    assert not maybe_challenge(claim, agent.context), "Tried to challenge own claim"
+    assert not maybe_challenge(
+        claim, agent._event_processors[direction]._context
```
istankovic, line 115: Private variable access.

beamer/tests/agent/test_request.py
```diff
@@ -147,12 +153,14 @@ def test_fill_and_claim(request_manager, token, agent, allow_unlisted_pairs):
...
+def test_withdraw(request_manager, token, agent, direction):
     requester, target = alloc_accounts(2)
     request_id = make_request(request_manager, token, requester, target, 1)
 
     with Sleeper(10) as sleeper:
-        while (request := agent.context.requests.get(request_id)) is None:
+        while (
+            request := agent._event_processors[direction]._context.requests.get(request_id)
+        ) is None:
             sleeper.sleep(0.1)
```
istankovic, line 164: This is not nice, in addition to accessing private variables.
How about something like

```python
   context = agent.get_context(direction)
   while (request := context.requests.get(request_id)) is None:
      sleeper.sleep(0.1)
```
?

beamer/agent.py
```diff
@@ -31,107 +37,165 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
             latest_blocks={},
             config=self._config,
-            web3_l1=w3_l1,
+            web3_l1=l1.w3,
             task_pool=self._task_pool,
             claim_request_extension=claim_request_extension,
             l1_resolutions={},
             l1_invalidations={},
         )
+        direction_identifier = f"{direction[0]}->{direction[1]}"
```
istankovic, line 152: I think we should introduce more type-safety around the direction concept. For example, how about this:

```python

class TransferDirection(typing.NamedTuple):
     source: ChainID
     target: ChainID
```

If we had something like that in `beamer.typing`, it could be nicely used in functions currently taking plain `str` arguments. Additional benefit of having such a type is not having to use indices, but readable fields.

Note: we could also add a custom `__repr__` like

```python
def __repr__(self):
    return f"{self.source} -> {self.target}"
```
if we deem it necessary.

beamer/agent.py
```diff
@@ -31,107 +37,165 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
+    def _init_chains(self) -> dict[ChainId, _Chain]:
+        chains: dict[ChainId, _Chain] = {}
+        for chain_name, rpc_url in self._config.rpc_urls.items():
+            if chain_name == "l1":
+                continue
+            w3 = make_web3(rpc_url, self._config.account)
+            chain_id = ChainId(w3.eth.chain_id)
+            if chain_id in chains:
+                continue
+            contracts_info = _get_contracts_info(self._config, w3)
```
istankovic, line 83: Here we are passing `w3` to `_get_contracts_info`, which uses it just to get the chain ID.
However, we have just obtained the chain ID just a few lines above, so we can, instead of passing a `Web3` instance, give `_get_contracts_info` the chain ID directly and thereby save one RPC call.

beamer/agent.py
```diff
@@ -31,107 +37,165 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
-                chain_related_data = {
-                    "web3": w3_target,
-                    "contracts": (target_request_manager, target_fill_manager),
-                    "deployment_block": _get_deployment_block(target_contracts_info),
-                }
-            event_monitor_data.update(chain_related_data)
-            self._event_monitors[chain_id] = EventMonitor(**event_monitor_data)  # type: ignore
+        for direction in set(directions):
+            direction = cast(tuple[ChainId, ChainId], direction)
+            self._setup_direction(direction, chains, l1)
```
istankovic, line 176: We are using `set` here to filter duplicates of the form `(x, x)` in case we only have one L2 chain, right?

Would it be better to use `itertools.product` instead of `permutations`?

```python
>>> import itertools as it
>>> list(it.product([1], repeat=2))
[(1, 1)]
>>> list(it.product([1,2], repeat=2))
[(1, 1), (1, 2), (2, 1), (2, 2)]
```
This way, we don't get duplicates and there is no need to filter.

beamer/agent.py
```diff
@@ -31,107 +37,165 @@ def _get_deployment_block(contract_info: dict[str, ContractInfo]) -> BlockNumber
...
             address=self._config.account.address,
             latest_blocks={},
             config=self._config,
-            web3_l1=w3_l1,
+            web3_l1=l1.w3,
             task_pool=self._task_pool,
             claim_request_extension=claim_request_extension,
             l1_resolutions={},
             l1_invalidations={},
         )
```
istankovic, line 151: The context is slowly becoming a monster, but that was easily predictable. :woozy_face: 
Nothing to be done about it here, but we should keep it in mind in the future, to revisit it at some point.

beamer/chain.py
```diff
@@ -77,6 +77,10 @@ def stop(self) -> None:
         self._stop = True
         self._thread.join(_STOP_TIMEOUT)
 
+    def subscribe(self, event_processor: "EventProcessor") -> None:
+        self._on_new_events.append(event_processor.add_events)
+        self._on_sync_done.append(event_processor.mark_sync_done)
```
istankovic, line 80: This is not ideal because it introduces coupling between `EventMonitor` and `EventProcessor`, but let's go with it for now. We can think of a nice way to get rid of the coupling after the PR is merged.

beamer/agent.py
```diff
@@ -1,8 +1,14 @@
 import threading
 from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
+from itertools import permutations
+from typing import cast
+
 import structlog
+from brownie import FillManager, RequestManager
```
istankovic, line 8: `brownie` must absolutely not be imported by the agent.

beamer/chain.py
```diff
@@ -49,8 +49,8 @@ def __init__(
         contracts: tuple[Contract, ...],
         deployment_block: BlockNumber,
         on_new_events: list[Callable[[list[Event]], None]],
-        on_sync_done: Callable[[], None],
-        poll_period: int,
+        on_sync_done: list[Callable[[], None]],
```
istankovic, line 50: I would suggest making a separate type for the callable, e.g.

```python
SyncDoneCallback = Callable[[], None]
```
which would then allow us to be have readable

```python
        on_sync_done: list[SyncDoneCallback],
```
(not sure whether the type needs to be private, but if it can be made, then we should make it private).

beamer/chain.py
```diff
@@ -49,8 +49,8 @@ def __init__(
         contracts: tuple[Contract, ...],
         deployment_block: BlockNumber,
         on_new_events: list[Callable[[list[Event]], None]],
```
istankovic, line 49: Similarly to the comment for sync done, we could do 

```python
NewEventsCallback = Callable[[list[Event]], None]
```
and then have
```python
        on_new_events: list[NewEventsCallback],
```

beamer/state_machine.py
```diff
@@ -61,6 +61,16 @@ class Context:
     l1_invalidations: dict[ClaimId, Future]
     finality_periods: dict[ChainId, int] = field(default_factory=dict)
 
+    @property
+    def source_rpc_url(self) -> URL:
+        provider = cast(HTTPProvider, self.request_manager.web3.provider)
+        return URL(str(provider.endpoint_uri))
```
istankovic, line 68: Why do we need `str` here? Wouldn't just `URL(provider.endpoint_uri)` be enough?

docs/source/agent.rst
```diff
@@ -25,33 +25,35 @@ In order to satisfy above requirements, we opted for a simple thread-based appro
...
 EventProcessor
 --------------
 
 ``EventProcessor`` implements the Beamer protocol logic. It receives events from event monitors and
-stores them into a list. It is important to note that events are not separated based on the chain
-they came from -- all events are stored in a single list, in the order they arrived, regardless of
-the originating chain.
+stores them into a list. Events are filtered by source and target chains and events are stored in a single list, in the order they arrived.
 
 .. note::
```
istankovic, line 59: Commenting on the new image here because I could not find a way to comment on the image itself:

- we should fix the names (there is no more `ContractEventMonitor`)
- we should remove `(RequestManager & FillManager)` from the boxes, they just  make the boxes seem cluttered
- we should make the image a bit wider and the vertical lines connecting event monitors and event processors could be grouped such that the two vertical lines going into event processor A->B are closer together, then some bigger space, then two vertical lines going into event processor A->C closer together, and so on.
- would be nice to have the "Thread N" labels aligned horizontally with respect to boxes

docs/source/agent.rst
```diff
@@ -25,33 +25,35 @@ In order to satisfy above requirements, we opted for a simple thread-based appro
 all software components that could potentially block into their own threads. In particular, these
 are ``ContractEventMonitor`` and ``EventProcessor``.
 
+With the current implementation, we have a ``EventProcessor`` for every direction in their own thread running for the agent.
+This helps to scale well, as if there is a new L2 chain added, then agent would just create another directions which has the new
+chain and there is no need to run seperate processes for this manner. Also, we have L2 chain based ``ContractEventMonitors`` that 
+event processors can subscribe and get events from. This prevents us to create redundant event monitors for event processors.
+
 
 ContractEventMonitor
```
istankovic, line 34: This needs to be changed,  everywhere. There is no `ContractEventMonitor` anymore. It's not been here for some time, but seems we forgot to update the docs.

docs/source/agent.rst
```diff
@@ -25,33 +25,35 @@ In order to satisfy above requirements, we opted for a simple thread-based appro
 all software components that could potentially block into their own threads. In particular, these
 are ``ContractEventMonitor`` and ``EventProcessor``.
 
+With the current implementation, we have a ``EventProcessor`` for every direction in their own thread running for the agent.
```
istankovic, line 28: I would reformulate this as:
```
For every transfer direction, we have an ``EventProcessor`` running in its own thread.
```

docs/source/agent.rst
```diff
@@ -25,33 +25,35 @@ In order to satisfy above requirements, we opted for a simple thread-based appro
...
+This helps to scale well, as if there is a new L2 chain added, then agent would just create another directions which has the new
+chain and there is no need to run seperate processes for this manner. Also, we have L2 chain based ``ContractEventMonitors`` that 
+event processors can subscribe and get events from. This prevents us to create redundant event monitors for event processors.
+
```
istankovic, line 32: I don't think we need to mention the older, multiple-agent-processes implementation here, it's just something that we had and is now gone.

Please also don't make the lines in paragraph too long and try to follow the rest of the document.

docs/source/agent.rst
```diff
@@ -25,33 +25,35 @@ In order to satisfy above requirements, we opted for a simple thread-based appro
...
+JSON-RPC server and can block. However, even if the event fetcher for, say, source chain, blocks, since
 it is only being invoked by the contract event monitor, which runs inside a thread, it won't block
 the entire agent. Therefore, even if a JSON-RPC server is very slow or the connection is otherwise
 unreliable, the agent as a whole will remain responsive.
 
-With the current contract implementation, we have one contract, ``RequestManager``, deployed on the
-source chain (L2a), and one contract, ``FillManager``, deployed on the target chain (L2b). Therefore
-we have two event monitors, one for each contract, and each event monitor has its own event fetcher,
-as can be seen in the figure above.  Each pair ``(EventFetcher, ContractEventMonitor)`` works
+With the current contract implementation, we have both ``RequestManager`` and ``FillManager`` for L2 chain.
```
istankovic, line 46: This sentence looks weird, I am not sure I understand what we want to say here.

docs/source/agent.rst
```diff
@@ -25,33 +25,35 @@ In order to satisfy above requirements, we opted for a simple thread-based appro
...
+Therefore we have event monitor for every L2 chain, and each event monitor has its own event fetcher,
```
istankovic, line 47: Please use an indefinite article here: `have event monitor` -> `have an event monitor`

docs/source/agent.rst
```diff
@@ -25,33 +25,35 @@ In order to satisfy above requirements, we opted for a simple thread-based appro
...
+as can be seen in the figure above. Each pair ``(EventFetcher, ContractEventMonitor)`` works
```
istankovic, line 48: Please don't add new mentions of `ContractEventMonitor` -- it's gone.

docs/source/agent.rst
```diff
@@ -25,33 +25,35 @@ In order to satisfy above requirements, we opted for a simple thread-based appro
...
 
 
 EventProcessor
 --------------
 
 ``EventProcessor`` implements the Beamer protocol logic. It receives events from event monitors and
-stores them into a list. It is important to note that events are not separated based on the chain
-they came from -- all events are stored in a single list, in the order they arrived, regardless of
-the originating chain.
+stores them into a list. Events are filtered by source and target chains and events are stored in a single list, in the order they arrived.
```
istankovic, line 56: Please try to keep the line lengths consistent with the rest of the document.

istankovic, line 56: > Events are filtered by source and target chains and events are stored in a single list

There is no need to mention `events` a second time. This would read better:

`Events are filtered by source and target chains and are stored in a single list`

beamer/agent.py
```diff
@@ -145,6 +145,9 @@ def _setup_direction(
             claim_request_extension=claim_request_extension,
             l1_resolutions={},
             l1_invalidations={},
+            logger=structlog.get_logger("EventProcessor").bind(
+                source_chain_id=source_chain.id, target_chain_id=target_chain.id
+            ),
```
istankovic, line 150: Few things here:

- nitpick: this would look nicer if we were to only have `logger=logger` here and assign to `logger` above the context creation line
- I don't think we should use `EventProcessor` here because this logger is going to be owned by the context; so maybe `Context`? It's not ideal either, but it's a result of the whole messy split of the event processor funcs....

istankovic (changes requested):
I think this has been progressing nicely, but there are still some changes to be done.

beamer/state_machine.py
istankovic: So why you don't need a call to `str` in `conftest.py`?

beamer/agent.py
bilbeyt: just looked this one more time, as i am using chain_id as key in chains, chain_ids should be a list of unique chain ids. if the length is only one then it is adding only in that case a duplicate chain_id to the chain_ids, then permutations should have no duplicate direction in my opinion. before using chain_id i was using name and probably i needed a set here. will just remove set.

bilbeyt: in normal cases we are not adding same direction like A->A like product is doing in your example it is just exception now if there is really only one chain.

docs/source/agent.rst
bilbeyt: Trying to say our eventfetcher now looking both. RequestManager and FillManager :D

beamer/state_machine.py
bilbeyt: imo `brownie.web3.provider.endpoint_uri` is also wrong as provider is BaseProvider but we dont have ignore there or we dont cast it. as endpoint_uri in conftest is Any i guess it is not raising anything but here we cast it and endpoint_uri is URI | None and it is raising.

bilbeyt: and without casting it is raising here for source_rpc_url

bilbeyt: only allowed to use `URL(provider.endpoint_uri or "")` if it is ok or need to use str directly.

beamer/agent.py
fredo: +1 prio point for state machine rework ;)

beamer/agent.py
istankovic: > just looked this one more time, as i am using chain_id as key in chains, chain_ids should be a list of unique chain ids. if the length is only one then it is adding only in that case a duplicate chain_id to the chain_ids, then permutations should have no duplicate direction in my opinion.

I do not understand.
Suppose we have this in `chains`: `[1]`. Because the length is `1`, you will duplicate the single chain ID: `[1, 1]`.
And now you give this list to `permutations`, which will generate the following sequence: `[(1, 1), (1, 1)]`.

Could you please explain why that is correct and why it is desired to call `_setup_direction` two times with the same arguments?

bilbeyt: ok i looked another time :D you are right about permutations, but still product is generating (1,1) and (2,2) as well. Should i keep it ?  We dont have something like this normally in run-your-own-repo

istankovic (approved):
Looking good, thanks!

PR #1493, opened 2023-02-15
Agent ratelimit

istankovic:


beamer/middleware.py
```diff
@@ -41,3 +51,221 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
...
+#
+# This effectively caps the request rate at 1 / _RATE_LIMIT_REQUEST_DELAY with
+# the hope that the RPC will eventually notice that we slowed down and accept
+# the requests. If the RPC rejects new requests even after the rate limiting
+# period is over, we simply raise a RuntimeError, which will cause the agent to
+# shut down, as there is nothing more that could be done.
+#
+# Tapering mode begins immediately after the end of rate limiting mode, with
+# the idea that we ramp up to the original normal operation request rate (i.e.
+# without any delays). Currently,t the ramp up is implemented in the following
```
bilbeyt, line 82: small typo

beamer/middleware.py
```diff
@@ -41,3 +51,221 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
...
+#
+# The lock protects all state fields except num_waiting_on_lock, which is an
+# atomic used to keep track of threads waiting on the lock.
+@dataclass(slots=True)
+class _RateLimiterState:
+    lock: threading.Lock = field(default_factory=threading.Lock)
+    num_waiting_on_lock: atomics.base.AtomicUint = field(
+        default_factory=lambda: atomics.atomic(width=4, atype=atomics.UINT)
+    )
+    rate_limit_end: None | float = None
```
bilbeyt, line 122: isnt this equal to Optional[float] ?

beamer/middleware.py
```diff
@@ -41,3 +51,221 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
...
+
+def _try_make_request(
+    make_request: _MakeRequest, method: RPCEndpoint, params: Any
+) -> tuple[bool, None | RPCResponse]:
+    try:
+        response = make_request(method, params)
+    except requests.exceptions.HTTPError as exc:
+        if exc.response.status_code == 429:
+            return True, None
+        raise exc
```
bilbeyt, line 152: In which case this will raise ? I understood it is HTTPError but do we have an example for this? Also, if this raises somehow then other parts will break

bilbeyt:
Other than those small things, it looks very good!

bilbeyt (approved):

beamer/middleware.py
istankovic: Indeed it is, this is by the new union types feature in newer Python: https://docs.python.org/3/library/stdtypes.html#types-union so `Optional` isn't really needed anymore.

beamer/middleware.py
```diff
@@ -41,3 +51,221 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
...
+#
+# In cases where an RPC returns HTTP 429 (Too many requests), we need to reduce
+# the rate at which we issue requests to that RPC. In order to do that, we
+# define the following operating modes:
+#
+# 1) normal operation
+# 2) rate limiting mode
+# 3) tapering mode
+#
+# In normal operation, the RPC accepts our every request and we issue requests
```
fredo, line 66: `RPC accepts every request`

beamer/middleware.py
```diff
@@ -41,3 +51,221 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
...
+#
+# One instance per web3 rate limiter middleware, i.e. one instance per
+# chain/RPC if there is exactly one Web3 instance per chain/RPC.
+#
+# Note: it is important to ensure only one Web3 instance per chain/RPC,
+# otherwise, it could happen that two parallel rate limiter middleware talk to
+# the same RPC, which would rate limit both middlewares.
+#
+# The lock protects all state fields except num_waiting_on_lock, which is an
+# atomic used to keep track of threads waiting on the lock.
```
fredo, line 115: It would be nice to mention somewhere that with the introduction of an overall lock, we effectively prohibit concurrent requests to the rpc

beamer/middleware.py
```diff
@@ -41,3 +51,221 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
...
+    make_request: _MakeRequest,
+    w3: Web3,
+    state: _RateLimiterState,
+) -> RPCResponse:
+    if hasattr(_RATE_LIMITER_TLD, "entered"):
+        # We already entered this function once and likely took the lock,
+        # so just make a request immediately to avoid deadlocks.
+        # The cause of a second entry into this function is most probably
+        # an eth_chainId call made as part of our call to make_request.
+        return make_request(method, params)
```
fredo, line 235: just to clarify: We can call `make_request()` here because we are sure, that this is call within another call which comes from `_try_make_request()` and thus an exception will bubble up and be caught in `_try_make_request()`?

fredo (approved):
Thanks for this beautiful PR. 
I left some comments and questions, but otherwise looks very good.

pyproject.toml
```diff
@@ -18,6 +18,7 @@ python-statemachine = "^0.9.0"
 prometheus-client = "^0.16.0"
 lru-dict = "^1.1.7"
 toml = "^0.10.2"
+atomics = "^1.0.2"
```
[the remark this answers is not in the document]
istankovic: The `atomics` package (https://github.com/doodspav/atomics) provides atomic primitives that allow you to safely modify data from multiple threads, without the need for a lack. That relies on hardware supporting these primitives, which may differ between platforms.

In the rate limiter middleware we are using this to implement a counter of the number of threads waiting on the rate limiter lock. We cannot use the lock to protect the counter so we use an atomic 4-byte wide integer to safely update the counter from multiple threads (synchronization is ensured on the hardware level).

beamer/middleware.py
istankovic: > `RPC accepts every request`

I prefer to leave the `our` in there to make it more explicit that this is about RPC rate limiting us, not other possible clients. In other words, we may be experiencing rate limits, but other clients of the same RPC may issue requests just fine.

beamer/middleware.py
istankovic: Done.

beamer/middleware.py
istankovic: Yes.

beamer/middleware.py
fredo: the grammar sounds weird to me, do you mean `the RPC accepts every request of ours` or something like that?

istankovic: > the grammar sounds weird to me, do you mean `the RPC accepts every request of ours` or something like that?

Yes, that is the intended meaning. I've just tried several gramamr checkers and none reported an error.
Consider this:

- our every request
- your every request

I'm guessing you'd consider the second example (with "your") less strange, but from the grammar point of view, they are looking the same.

Issue #355, opened 2022-03-16
agent: enable unlimited allowance for FillManager


fredo:
following #353, we can introduce the same change to the agent, saving on transaction per fill by approving an unlimited amount to the fill manager contract.

istankovic:
@fredo this one should be trivial, right?
Check if the token allowance for the agent's address is UINT256_MAX and, if not, approve the max amount.

istankovic:
@fredo ping... :-)

czepluch:
I'd say we should make sure that this is optional. If agents wants to enable it and do it, then fine, but it's a bit of bad practice, so maybe giving the agent / users the choice is the best practice.

fredo:
please note, that every transfer deducts the amount from the contract's allowance. If we make this value configurable with any real number (which IMO is desirable), we need to check the current allowance before any call. 

It then gets a bit tricky. We must decide how the agent reacts when the allowance decreases towards 0. When should the agent re-update the allowance and call `approve()`? It involves a strategy which is not quite trivial. But as always we should go ahead with an easy approach and optimize later.

One suggestion would be: 
Upon every fill check whether current allowance is sufficient (RPC call). If not, call approve to update allowance. I think this is a suboptimal solution but good for now. Note, that the agent will most likely lose the race for that particular fill against someone where allowance is high enough.

fredo:
Additionally, please note that agents which have an unlimited allowance have an advantage over other agents which do not, because such agents don't need the RPC call to check the allowance before the fill transaction. 
This might give them a couple of hundred milliseconds advantage and are more likely to win the "fill race". 
If it becomes financially unviable to run an agent with limits set, must yet to be seen.

istankovic:
> Upon every fill check whether current allowance is sufficient (RPC call). If not, call approve to update allowance. I think this is a suboptimal solution but good for now. Note, that the agent will most likely lose the race for that particular fill against someone where allowance is high enough.

I do not perceive that as a problem, because it arises from the very same thing that would be the solution in this case: configurable allowance.

The extra call is not nice in general, however, it is good enough for now.

fredo:
Not saying that this is a problem. But it needs to be handled by the agent correctly. And again, tehre are various ways to handle it.

fredo:
Picking this up again, as we see now inefficiency in the agents' performances on mainnet.

[...]

Issue #1346, opened 2023-01-11
Agent config validation


fredo:
Currently we check in the token match checker if the configured rpc have the same base chain.

Instead we should do either: 

a) a static analysis of the config
b) analyse the config upon agent's start up and either raise or give a warning that the config does not seem fine.

Here, we could gradually enhance the check around the config if values configured makes sense.

fredo:
We realized that no direct values in the config are checked syntactically. 
Especially in values which are typed (i.e. int) but have a smaller range of valid values.
For example: time periods can only be > 0 values.

Additionally the same sanity check should for for values we load from the deployment files (i.e. BlockNumber can only be > 0)

@istankovic will research for some generic toml and json validation approach and posts his findings here.

fredo:
Additionally, when implementing this issue, we should revise the values' types in the config such as 

chain id -> integer
allowance -> integer

in order to safe conversions and fail upon loading and/or validation directly

PR #1510, opened 2023-02-20
Allow configurable approval for tokens

fredo:
implements: #355

To each tuple of tokens, a third field is optional.
It can be: 

- a value, representing the allowance to be approved.
- -1, which is highest possible allowance
- If no value is given (tuple size == 2) the request amount is taken as a dynamic allowance value.

-1 was chosen for the maximum allowance as it represents `0xff....ff` in K2 complementary integer representation. Additionally, other tools, like etherscan allow setting maximum allowance with -1, so it is a common representation.


beamer/data/agent.conf.example
```diff
@@ -25,6 +25,13 @@ rpc-url = "GOERLI_OPTIMISM_RPC_URL"
 # Each token is represented by a pair [chain-id, token-address].
 # All tokens within the same list are considered equivalent and
 # transfers between them are allowed.
+# A third value per token representation is optional, which defines
+# the allowance amount, the agent will approve to the fill manager contract.
+# [chain-id, token-address, allowance]
+# Allowed values are:
+# - Any value > 0, which defines the exact allowance given
+# - -1, the agent will approve type(uint256).max
```
istankovic, line 33: How about using TOML-supported infinity value, `inf` e.g.
```toml
        ["11", "0x2644292EE5aed5c17BDcc6EDF1696ba802351cf6", inf]
```
which would in Python be converted to `math.inf`?

fredo: I gave my explanation above why I chose -1, and actually inf would also semantically not be correct as more recent versions of ERC20 don't support unlimited allowance, only set it to the max value, but handle it as any other value (subtracting the amounts on each transfer)

beamer/util.py
```diff
@@ -136,6 +137,13 @@ def make_web3(
 
 _Token = tuple[ChainId, ChecksumAddress]
 
+
+@dataclass(frozen=True)
+class _TokenData:
+    equivalent_classes: frozenset[_Token]
+    allowance: Optional[int]
```
istankovic, line 144: Love that you factored this out!

Nitpick: I would rename `equivalent_classes` to `equivalence_class` because it really refers to the equivalence class of this token (note the singular; https://en.wikipedia.org/wiki/Equivalence_class); it is the equivalence class (just a fancy name for a set) which this token is a member of. 

Also note you can freely use union type syntax instead of `Optional` because we are on a fresh enough Python version: `allowance: None | int`.

beamer/util.py
```diff
@@ -165,9 +172,18 @@ def __init__(self, tokens: List[List[List[str]]]) -> None:
...
-            for token in equiv_class:
+            for token in token_mapping:
+                chain_id = ChainId(int(token[0]))
+                token_address = to_checksum_address(token[1])
+                if len(token) == 2:
+                    allowance = None
+                elif int(token[2]) == -1:
+                    allowance = 2**256 - 1
+                else:
+                    allowance = int(token[2])
```
istankovic, line 183: I always wanted to try out the new structural pattern matching feature of recent Python and this looks like a perfect place for it! :smile: 

Consider
```python
match token:
   case [_, _]:
    allowance = None
   case [_, _, -1]:
    allowance = 2**256 - 1
   case [_, _, allowance]:
    pass # not sure whether int conversion is really needed here -- won't toml convert it to an int?
   case _:
    raise ValueError('unexpected token data: %r', token)
```

beamer/util.py
```diff
@@ -153,7 +153,7 @@ class _TokenData:
 }
 
 
-class TokenMatchChecker:
+class TokenChecker:
```
istankovic, line 156: You deserve a beer just for this change! It's definitely a better name now.

beamer/chain.py
```diff
@@ -303,15 +303,28 @@ def fill_request(request: Request, context: Context) -> None:
...
         token.functions.allowance(context.address, context.fill_manager.address).call()
         < request.amount
     ):
-        func = token.functions.approve(context.fill_manager.address, request.amount)
+        func = token.functions.approve(context.fill_manager.address, allowance)
         try:
             transact(func)
         except TransactionFailed as exc:
-            log.error("approve failed", request_id=request.id, exc=exc)
+            log.error("Approve failed", request_id=request.id, exc=exc)
```
istankovic, line 327: Nitpick: I prefer the previous spelling because it referred to the function name, `approve`, which I think is consistent with all the other logging calls in the agent.

beamer/tests/agent/test_request.py
```diff
@@ -147,6 +147,47 @@ def test_fill_and_claim(request_manager, token, agent, allow_unlisted_pairs):
...
+
+    else:
+        allowance_after = token.allowance(agent.address, fill_manager)
+
+        # Some token contracts implement either an unlimited allowances at type(uint256).max
+        # other start decreasing the allowance from the highest amount.
+        if actual_allowance == 2**256 - 1:
+            assert allowance_after in [actual_allowance - amount, actual_allowance]
+        else:
+            assert token.allowance(agent.address, fill_manager) == actual_allowance - amount
```
istankovic, line 188: Can we just use `allowance_after` here instead of making another call to `allowance`?

beamer/tests/agent/test_request.py
```diff
@@ -147,6 +147,47 @@ def test_fill_and_claim(request_manager, token, agent, allow_unlisted_pairs):
...
+                sleeper.sleep(0.1)
+    except Timeout:
+        # Agent did not fill
+        assert actual_allowance < amount
+
+    else:
+        allowance_after = token.allowance(agent.address, fill_manager)
+
+        # Some token contracts implement either an unlimited allowances at type(uint256).max
+        # other start decreasing the allowance from the highest amount.
```
istankovic, line 184: Some nitpicks in the comment:
```
        # Some token contracts implement an unlimited allowance as type(uint256).max,
        # others start decreasing the allowance from the highest amount.
```

beamer/tests/agent/test_request.py
```diff
@@ -147,6 +147,47 @@ def test_fill_and_claim(request_manager, token, agent, allow_unlisted_pairs):
...
+        with Sleeper(5) as sleeper:
+            while (
+                request := agent.context.requests.get(request_id)
+            ) is None or request.filler != agent.address:
+                sleeper.sleep(0.1)
+    except Timeout:
+        # Agent did not fill
+        assert actual_allowance < amount
+
+    else:
```
istankovic, line 180: Nitpick: extraneous empty line.

beamer/tests/agent/test_request.py
```diff
@@ -147,6 +147,47 @@ def test_fill_and_claim(request_manager, token, agent, allow_unlisted_pairs):
...
+        allowance_after = token.allowance(agent.address, fill_manager)
+
+        # Some token contracts implement either an unlimited allowances at type(uint256).max
+        # other start decreasing the allowance from the highest amount.
+        if actual_allowance == 2**256 - 1:
+            assert allowance_after in [actual_allowance - amount, actual_allowance]
```
istankovic, line 186: I feel like we should hardcode the expected behaviour for our token, because we know how it behaves and what the value is supposed to be. Since we are using only one ERC20 token implementation, I'd rather close the door here.

beamer/data/agent.conf.example
```diff
@@ -25,6 +25,13 @@ rpc-url = "GOERLI_OPTIMISM_RPC_URL"
...
+# If no value is given, the requested amount will be approved
```
istankovic, line 34: Nitpick: missing end-of-sentence dot.

docs/source/configuration.rst
```diff
@@ -78,21 +85,28 @@ transferring tokens that are not in the set will be ignored.
...
+If no value is given, the requested amount will be approved 
 
 Example::
 
     [tokens]
     TST = [
-        ["11", "0x2644292EE5aed5c17BDcc6EDF1696ba802351cf6"],
-        ["22", "0xAcF5e964b76773166F69d6E53C1f7A9114a8E01D"]
+        ["11", "0x2644292EE5aed5c17BDcc6EDF1696ba802351cf6", "-1"],
+        ["22", "0xAcF5e964b76773166F69d6E53C1f7A9114a8E01D", "100000000000"]
```
istankovic, line 109: I wonder, why don't we just use integers, i.e. without quotes?

beamer/util.py
fredo: >Also, not sure whether we check it already somewhere, but it would be good if we check that the allowance is >= 0 (if it is not -1).

I think it is a general problem that we do not really check the configured values. So we should address this as a whole IMO. ( I think there is already an issue to that)

docs/source/configuration.rst
fredo: Well, the same thing you could argue for chain id. I just wanted to stay consistent. I'm open to change it, but then we should also do it for all integers in this list

beamer/util.py
fredo: 🍺

fredo:
I decided to leave the allowance values as strings for now, and added the revision of those types to #1346. 
In the process of validation we should revise the types of the values.

beamer/tests/agent/test_request.py
```diff
@@ -113,23 +113,15 @@ def test_challenge_own_claim(config, request_manager, token, direction):
     assert not maybe_challenge(claim, context), "Tried to challenge own claim"
 
 
-@pytest.mark.parametrize("allow_unlisted_pairs", (True, False))
-def test_fill_and_claim(request_manager, token, agent, allow_unlisted_pairs, direction):
+def test_fill_and_claim(request_manager, token, agent, direction):
```
istankovic, line 116: Why was the `allow_unlisted_pairs` fixture and parameterization removed?

beamer/util.py
```diff
@@ -136,6 +137,13 @@ def make_web3(
 
 _Token = tuple[ChainId, ChecksumAddress]
 
+
+@dataclass(frozen=True)
+class _TokenData:
+    equivalence_classes: frozenset[_Token]
```
istankovic, line 143: This should be `equivalence_class` (singular, not plural).

beamer/util.py
```diff
@@ -165,9 +172,21 @@ def __init__(self, tokens: List[List[List[str]]]) -> None:
...
+                token_address = to_checksum_address(token[1])
+                match token:
+                    case [_, _]:
+                        allowance = None
+                    case [_, _, "-1"]:
+                        allowance = 2**256 - 1
+                    case [_, _, _allowance]:
+                        allowance = int(_allowance)
+                    case _:
+                        raise ValueError("unexpected token data: %r", token)
```
istankovic, line 186: Nice!

beamer/chain.py
```diff
@@ -318,11 +318,24 @@ def fill_request(request: Request, context: Context) -> None:
...
+            request_amount=request.amount,
+        )
+        return
+
     if (
         token.functions.allowance(context.address, context.fill_manager.address).call()
         < request.amount
     ):
-        func = token.functions.approve(context.fill_manager.address, request.amount)
+        func = token.functions.approve(context.fill_manager.address, allowance)
```
istankovic, line 338: Just to check whether my understanding is correct: if `allowance` is `-1`, when preparing tx data it will be converted internally by `web3` to the maximum representable value of `uint256` (`0xff...ff`), which is what we want?

beamer/tests/agent/test_request.py
fredo: The reason for it was that with the `token_list` fixture, the token is allowed either way.
So the test becomes useless. Additionally, it is not essential for what the test should do. the test is about filling and claiming, not about filtering requests.

beamer/chain.py
fredo: good idea! unfortunately, web3.py cannot do the special casing. But moving this into token_checker.allowance is the right thing to do.

fredo: so actually we could remove the comparison to -1 above.

beamer/tests/agent/test_request.py
istankovic: Then we should probably also remove the `allow_unlisted_pairs` fixture from `conftest.py` as it does not appear to be used anymore.

istankovic (approved):
Looks good. :+1:

beamer/tests/agent/test_request.py
fredo: it is actually used in the agent via `set_allow_unlisted_pairs` and thus in a lot of tests. I test if we can get rid of this at all.

fredo: tests passed smoothly by removing the ENV. will remove it

PR #1519, opened 2023-02-23
tests: implement test for multiple event processors

bilbeyt:


beamer/tests/agent/multiple_chain/conftest.py
```diff
@@ -0,0 +1,217 @@
...
+        )
+        chain_contracts[rollup_info] = _RollupContracts(
+            l1_messenger=l1_messenger,
+            l2_messenger=l2_messenger,
+            resolver=resolver,
+            fill_manager=fill_manager,
+            request_manager=request_manager,
+            token=token,
+        )
+    return chain_contracts
```
fredo, line 163: It looks very similar to the deploy function in the conftest, one dir up. could you  outsource the similiar code into a utility function?

fredo (changes requested):

fredo:
General question: I see a lot of copying fixtures to the lower hierarchy conftest.
Could we go for a more generic approach with less redundancy?

I'm thinking of reducing redundancy by creating utility functions for one. 

The other question is, wouldn't it be possible to just adapt whats needed in the nested folder (like adapting the config) and other fixtures like `agent` would be from the upper conftest with the new config? I'm sure pytest would support something like this.

beamer/tests/agent/multiple_chain/conftest.py
bilbeyt: it is not actually, this one is deploying things by looking at the rpc and returning web3 contracts while the other one directly using brownie rpc only and returning Brownie type ContractInstances

bilbeyt: and we have token here as an extra

beamer/agent/agent.py
```diff
@@ -188,6 +188,9 @@ def start(self) -> None:
             event_monitor.start()
         self._stopped.clear()
 
+    def get_directions(self) -> list[TransferDirection]:
```
istankovic, line 191: We should return `tuple[TransferDirection]` if we don't need mutability.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+from beamer.agent.agent import Agent
+from beamer.agent.config import Config
+from beamer.agent.contracts import ContractInfo
+from beamer.agent.typing import BlockNumber, ChainId
+from beamer.agent.util import TokenChecker, make_web3, transact
+from beamer.tests.util import Sleeper
+
+SLAVE_TEST_PATH = Path(__file__).parent / "test_slave.py"
+CONFIG_PATH = os.path.join(os.getcwd(), "ape-config.yaml")
+BACKUP_PATH = os.path.join(os.getcwd(), "ape-config.yaml.2")
```
istankovic, line 27: Why do we assume `ape-config.yaml` will be located at `os.getcwd()`?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+        "--wallet.mnemonic",
+        "brownie",
+        "--chain.chainId",
+        str(chain_id),
+    ]
+    return psutil.Popen(cmd, stdin=DEVNULL, stdout=PIPE, stderr=PIPE)
+
+
+def _start_ganache(chain_id: ChainId, port: str) -> psutil.Popen:
+    return _start_ganache_process(chain_id, port)
```
istankovic, line 80: Why not inline `_start_ganache_process` here?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+    current_config["ganache"]["server"]["port"] = port
+    with open(CONFIG_PATH, "w") as _f:
+        yaml.dump(current_config, _f, sort_keys=False, default_flow_style=False)
+
+
+def _restore_backup_config(config: dict[str, Any]):
+    os.remove(CONFIG_PATH)
+    os.remove(BACKUP_PATH)
+    with open(CONFIG_PATH, "w") as _f:
+        yaml.dump(config, _f, sort_keys=False, default_flow_style=False)
```
istankovic, line 55: I suggest we replace

- `_create_and_return_backup_config` 
- `_create_new_config`
- `_restore_backup_config`

with just a single single function, a context manager that overwrites ape config, e.g.:

```python
   with _configure_ganache_port(port):
         slave = _start_slave_test()
         ...
```
where the context manager would take care to restore the original config.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+            processes[chain_id] = _start_ganache(chain_id, port)
+        yield current_config
+    finally:
+        for process in processes.values():
+            _kill_ganache(process)
+
+
+def _get_chain_map() -> dict[ChainId, int]:
+    number_of_chains = 4
+    return {ChainId(i): 8545 + i for i in range(1, number_of_chains + 1)}
```
istankovic, line 115: I would avoid using low chain IDs, especially chain ID 1.

Why not just use port numbers as chain IDs?
```python
    return {ChainId(i): i for i in range(8546, 8546 + number_of_chains)}
```

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def _get_config(
+    chain_map: dict[ChainId, int],
+    contracts,
+    allowance,
+    slave_contract_addresses: dict[ChainId, dict[str, str]],
```
istankovic, line 122: It's weird that we have types for some arguments, but for some not.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+        rpc_urls=rpc_urls,
+        deployment_info=deployment_info,
+        token_checker=TokenChecker(token_list),
+        account=account,
+        fill_wait_time=0,
+        unsafe_fill_time=600,
+        prometheus_metrics_port=None,
+        log_level="debug",
+    )
+    beamer.agent.metrics.init(config=config, source_rpc_url=url, target_rpc_url=url)
```
istankovic, line 160: Why do we need this here?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+    directions = agent.get_directions()
+
+    for direction in directions:
+        with Sleeper(5) as sleeper:
+            while len(agent.get_context(direction).requests) != 1:
+                sleeper.sleep(0.1)
+
+            assert len(agent.get_context(direction).requests) == 1
+
+            request = list(agent.get_context(direction).requests)[0]
```
istankovic, line 177: Why do you use `list` here?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+            with Sleeper(5) as sleeper:
+                while not request.is_claimed:
+                    sleeper.sleep(0.1)
+    agent.stop()
+
+
+def _get_slave_contract_addresses(proc: psutil.Popen):
+    while True:
+        out = proc.stdout.readline().decode()
+        print(out, flush=True)
```
istankovic, line 188: Is this `print` intentional?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+        if "Chain is ready" in out:
+            proc.stdin.write("get_contracts\n".encode())
+            proc.stdin.flush()
```
istankovic, line 191: Is this `flush` really necessary if you write a buffer ending with `\n`?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+            out = proc.stdout.readline().decode()
+            contract_addresses = json.loads(out.replace("\n", ""))
```
istankovic, line 193: Why not instead
```python
line = proc.stdout.readline().decode().strip()
contract_addresses = json.loads(line)
```
?

Since you are reading only one line and only stripping one newline.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+        proc.wait()
+
+
+def _get_token_list(chain_map, allowance, slave_contract_addresses):
+    tokens = []
+    for chain_id in chain_map:
+        if allowance is None:
+            tokens.append([chain_id, slave_contract_addresses[chain_id]["token"]])
+        else:
+            tokens.append([chain_id, slave_contract_addresses[chain_id]["token"], allowance])
```
istankovic, line 209: Can we turn those inner lists `[chain_id, slave_con....]` into tuples?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+        while not w3.is_connected():
+            time.sleep(1)
+        l2_token = w3.eth.contract(
+            address=slave_contract_addresses[int(chain_name)]["token"],
+            abi=[abi.dict() for abi in token.contract_type.abi],
+        )
+        transact(l2_token.functions.mint(config.account.address, 300))
+
+
+def test_multiple_event_processor(contracts, token, allowance):
```
istankovic, line 227: Nitpick: `test_multiple_event_processor` is a weird singular, I'd suggest either `test_multiple_event_processors` or `test_multiple_chain_pairs`.

istankovic, line 227: Why do you need `allowance` fixture here?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+    chain_map = _get_chain_map()
+    with _new_networks(chain_map) as ape_config:
+        slave_test_procs = []
+        slave_contract_addresses = {}
+        for chain_id, port in chain_map.items():
+            slave = _start_slave_test(port, len(chain_map.values()))
```
istankovic, line 233: `len(chain_map.values())` is the same as `len(chain_map)` (only less efficient).

beamer/tests/agent/multiple_processors/test_slave.py
```diff
@@ -0,0 +1,32 @@
...
+    for i in range(1, request_count + 1):
+        if i == chain_id:
+            continue
+        contracts.request_manager.setFinalityPeriod(i, finality_period)
+        make_request(contracts.request_manager, token, requester, target, 1, target_chain_id=i)
+    print("Chain is ready", flush=True)
+
+    while True:
+        command = sys.stdin.readline()
+        if command == "get_contracts\n":
```
istankovic, line 23: Nitpick: I suggest to just use 
```python
        command = sys.stdin.readline().strip()
```
to avoid dealing with newlines.

beamer/tests/agent/multiple_processors/test_slave.py
```diff
@@ -0,0 +1,32 @@
...
+            contract_addresses = {
+                "request_manager": contracts.request_manager.address,
+                "fill_manager": contracts.fill_manager.address,
+                "token": token.address,
+                "l2_messenger": contracts.l2_messenger.address,
+            }
+            sys.stdout.write(json.dumps(contract_addresses) + "\n")
+            sys.stdout.flush()
```
istankovic, line 31: Why write `\n` if you're going to call `flush`?

beamer/tests/agent/multiple_processors/test_slave.py
```diff
@@ -0,0 +1,32 @@
...
+import time
+import ape
+
+from beamer.tests.util import alloc_accounts, make_request
+
+
+def test_sending_request_slave(contracts, token, finality_period):
+    requester, target = alloc_accounts(2)
+    chain_id = ape.chain.chain_id
+    request_count = int(os.environ.get("REQUEST_COUNT", 0))
```
istankovic, line 13: This environment variable must be present, right?
In that case, I would not call `get` -- let it crash the test.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,241 @@
...
+                address=to_canonical_address(
+                    slave_contract_addresses[chain_id]["request_manager"]
+                ),
+                abi=[abi.dict() for abi in contracts.request_manager.contract_type.abi],
+            ),
+            FillManager=ContractInfo(
+                deployment_block=BlockNumber(1),
+                address=to_canonical_address(slave_contract_addresses[chain_id]["fill_manager"]),
+                abi=[abi.dict() for abi in contracts.fill_manager.contract_type.abi],
+            ),
```
istankovic, line 141: I wish there was a nicer way to do this... :woozy_face:

.circleci/workflows.yml
```diff
@@ -177,7 +177,7 @@ jobs:
     executor: vm
     steps:
       - initialize-environment
-      - run: poetry run ape test beamer/tests/{agent,health} -s --gas --cov beamer --cov-report=term
+      - run: poetry run ape test beamer/tests/{agent,health} -s --gas -k "not slave" --cov beamer --cov-report=term
```
istankovic, line 180: This is not necessary, assuming you rename the test, as I suggested in my comment.

pyproject.toml
```diff
@@ -71,9 +72,14 @@ multi_line_output = 3
 ignore_whitespace = true
 
 [[tool.mypy.overrides]]
-module = ["ape.*", "twisted.*"]
+module = ["ape.*", "twisted.*", "yaml.*"]
```
istankovic, line 75: Why this `yaml.*` thing?

istankovic (changes requested):
The approach looks quite good, but there are many details that need to be taken care of.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
bilbeyt: because it is :) getcwd() basically the project root as we can say and ape-config.yaml is there.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
bilbeyt: it is there also for conftest config fixture. i can just delete it

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
bilbeyt: it is for me to see slave progress actually, i will try deleting this

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
bilbeyt: as long as i read regarding usage of it, yes it is necessary.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
bilbeyt: also as long as i remember, token_list is defined like this in agent as well so probably mypy will complain

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
bilbeyt: again for token_list fixture also defined in conftest.

beamer/tests/agent/multiple_processors/test_slave.py
bilbeyt: first iteration was like that and without \n flush didnt work.

pyproject.toml
bilbeyt: mypy complains yaml. stub is missing.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
istankovic: > it is there also for conftest config fixture. i can just delete it

I seem to remember that the metrics are being used internally so that may be the reason we first need to initialize that module. Could you please check that?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
istankovic: > also as long as i remember, token_list is defined like this in agent as well so probably mypy will complain

Oh well, if mypy complains about it, then ok, but it's... meh. It really should be a `Sequence` and we shouldn't use mutability where it is not needed.

beamer/tests/agent/multiple_processors/test_slave.py
istankovic: > first iteration was like that and without \n flush didnt work.

What is the buffering mode of `sys.stdout` here?

pyproject.toml
istankovic: > mypy complains yaml. stub is missing.

How does the error look like exactly? Aren't there typing stubs for yaml?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
bilbeyt: in agent.py we are initializing metrics already in start, maybe this is not necessary.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
bilbeyt: changed a little bit and now deleted

pyproject.toml
bilbeyt: beamer/tests/agent/multiple_processors/test_multiple_event_processors.py:13: error: Library stubs not installed for "yaml"  [import]
beamer/tests/agent/multiple_processors/test_multiple_event_processors.py:15: error: Library stubs not installed for "yaml.loader"  [import]
beamer/tests/agent/multiple_processors/test_multiple_event_processors.py:15: note: Hint: "python3 -m pip install types-PyYAML"
beamer/tests/agent/multiple_processors/test_multiple_event_processors.py:15: note: (or run "mypy --install-types" to install all missing stub packages)
beamer/tests/agent/multiple_processors/test_multiple_event_processors.py:15: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports

istankovic: So why don't we install the required typing packages? `types-PyYAML`?

bilbeyt: yes, now it is done.

pyproject.toml
```diff
@@ -20,6 +20,7 @@ lru-dict = "^1.1.7"
 toml = "^0.10.2"
 atomics = "^1.0.2"
 typing-extensions = "^4.5.0"
+types-pyyaml = "^6.0.0"
```
istankovic, line 23: Why is `types-pyyaml` not a dev-dependency? It is not needed outside of development.

I now notice that `typing-extensions` is also a regular dependency. We should make it a dev-dependency.

bilbeyt: moved

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,224 @@
...
+from beamer.agent.agent import Agent
+from beamer.agent.config import Config
+from beamer.agent.contracts import ContractInfo
+from beamer.agent.typing import BlockNumber, ChainId
+from beamer.agent.util import TokenChecker, make_web3, transact
+from beamer.tests.conftest import Contracts
+from beamer.tests.util import Sleeper
+
+SLAVE_TEST_PATH = Path(__file__).parent / "_test_slave.py"
+CONFIG_PATH = ape.project.local_project.path / ape.project.local_project.config_file_name
```
istankovic, line 25: Nice, this is much better, thanks for fixing it.
Both of those should be private, though.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,224 @@
...
+
+@contextlib.contextmanager
+def _configure_ganache_port(port: int):
+    try:
+        with open(CONFIG_PATH) as _f:
+            current_config = yaml.load(_f, Loader=SafeLoader)
+        current_port = current_config["ganache"]["server"]["port"]
+        current_config["ganache"]["server"]["port"] = port
+        with open(CONFIG_PATH, "w") as _f:
+            yaml.dump(current_config, _f, sort_keys=False, default_flow_style=False)
```
istankovic, line 63: This part should be moved outside the `try` block. We only want to do the revert to the original config if
we replaced it _and_ an exception happened during `yield`.

Also, please rename `_f` to just `f` or `fp` (think we have that elsewhere) -- I don't see a reason for the underscore here.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,224 @@
...
+    finally:
+        current_config["ganache"]["server"]["port"] = current_port
+        with open(CONFIG_PATH, "w") as _f:
+            yaml.dump(current_config, _f, sort_keys=False, default_flow_style=False)
+
+
+def _start_slave_test(port: int, request_count: int) -> psutil.Popen:
+    return psutil.Popen(
+        ["ape", "test", str(SLAVE_TEST_PATH), "-s"],
+        bufsize=0,
```
istankovic, line 74: Why `bufsize` of zero?

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,224 @@
...
+                while len(agent.get_context(direction).requests) != 1:
+                    sleeper.sleep(0.1)
+
+            request = tuple(agent.get_context(direction).requests)[0]
+
+            with Sleeper(20) as sleeper:
+                while not request.is_claimed:
+                    sleeper.sleep(0.1)
+    finally:
+        agent.stop()
```
istankovic, line 161: `agent.wait` is not necessary here because `agent.stop` will anyway block due to `event_processor.stop` and `event_monitor.stop` both being blocking, right?

bilbeyt: yes.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
bilbeyt: was trying something regarding \n but it doesnt work and forgot

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,223 @@
...
+    current_port = current_config["ganache"]["server"]["port"]
+    current_config["ganache"]["server"]["port"] = port
+    with open(_CONFIG_PATH, "w") as f:
+        yaml.dump(current_config, f, sort_keys=False, default_flow_style=False)
+    try:
+        yield
+    finally:
+        current_config["ganache"]["server"]["port"] = current_port
+        with open(_CONFIG_PATH, "w") as f:
+            yaml.dump(current_config, f, sort_keys=False, default_flow_style=False)
```
istankovic, line 68: Instead of serializing the YAML content again, we should instead
- load the original config
- temporarily move the original config file to a temporary dir (e.g. by way of `tmp_path` fixture)
- serialize the modified config
- yield
- move the original config back to its previous location

ape-config.yaml
```diff
@@ -1,45 +1,38 @@
...
   mnemonic: brownie
   number_of_accounts: 10
   hd_path: m/44'/60'/0'/0/{}
-
 ethereum:
   local:
     default_provider: ganache
-    default_transaction_type: 0
\ No newline at end of file
+    default_transaction_type: 0
```
istankovic, line 38: We should drop this commit -- it is a consequence of dumping YAML from a previous commit, but that should be changed.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -92,7 +92,7 @@ def _new_networks(chain_port_map: dict[ChainId, int]):
 
 
 def _get_chain_map() -> dict[ChainId, int]:
-    number_of_chains = 3
+    number_of_chains = 4
```
istankovic, line 95: This should be squashed in the commit that introduces that functions.

beamer/agent/agent.py
```diff
@@ -160,6 +161,7 @@ def _init(self) -> None:
         # This is necessary as we use the account for all resolutions and
         # would otherwise run into nonce problems
         self._task_pool = ThreadPoolExecutor(max_workers=1)
+        self._fill_mutexes: Tracker[tuple[ChainId, ChecksumAddress], threading.Lock] = Tracker()
```
istankovic, line 164: As mentioned before, we should not use `Tracker` here. Not only is it not necessary, but it is actively harmful because it has an implicit lock, which we do not want.

beamer/agent/chain.py
```diff
@@ -327,38 +327,47 @@ def fill_request(request: Request, context: Context) -> None:
...
-    if (
-        token.functions.allowance(context.address, context.fill_manager.address).call()
-        < request.amount
-    ):
-        func = token.functions.approve(context.fill_manager.address, allowance)
+    chain_id = ChainId(w3.eth.chain_id)
+    token_address = token.address
+    if (chain_id, token_address) not in context.fill_mutexes:
+        mutex = threading.Lock()
+        context.fill_mutexes.add((chain_id, token_address), mutex)
```
istankovic, line 334: This has not been addressed, I'm not sure whether github ate my comment, but anyway... we should create these locks during context creation, not here. Also, we should _not_ remove these locks -- they should be alive and reachable for as longer their contexts are.

istankovic (changes requested):

beamer/agent/agent.py
```diff
@@ -188,6 +188,9 @@ def start(self) -> None:
             event_monitor.start()
         self._stopped.clear()
 
+    def get_directions(self) -> tuple[TransferDirection, ...]:
+        return tuple(self._event_processors.keys())
```
istankovic, line 194: You should omit the `.keys()` part -- iteration over a `dict` is iteration over its keys and is the preferred way to do it.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -0,0 +1,223 @@
...
+    backup_path = _CONFIG_PATH.parent / "ape-config.yaml.original"
+    os.rename(_CONFIG_PATH, backup_path)
+    current_config["ganache"]["server"]["port"] = port
+    with open(_CONFIG_PATH, "w") as f:
+        yaml.dump(current_config, f, sort_keys=False, default_flow_style=False)
+    try:
+        yield
+    finally:
+        os.remove(_CONFIG_PATH)
+        os.rename(backup_path, _CONFIG_PATH)
```
istankovic, line 68: Nitpick: you can omit the call to `os.remove`, `os.rename` will conveniently overwrite it.

beamer/agent/agent.py
```diff
@@ -114,6 +116,12 @@ def _check_target_chain(self, target_chain: _Chain) -> None:
         if not target_chain.fill_manager.functions.allowedLps(self._config.account.address).call():
             raise RuntimeError("Agent address is not whitelisted on FillManager")
 
+    def _init_fill_mutexes(self, chains: dict[ChainId, _Chain]) -> None:
+        self._fill_mutexes: dict[tuple[ChainId, ChecksumAddress], threading.Lock] = {}
+        for chain in chains.values():
+            for chain_id, address in chain.tokens:
+                self._fill_mutexes[(chain_id, address)] = threading.Lock()
```
istankovic, line 123: Yup, this is it. :heart:

istankovic, line 123: I am wondering about the lifetime of these locks. You create them here, which is called from `_setup_direction`. When are these locks destroyed?

Note that an agent could be stopped and started again.

beamer/agent/chain.py
```diff
@@ -327,38 +327,41 @@ def fill_request(request: Request, context: Context) -> None:
...
-
-    if (
-        token.functions.allowance(context.address, context.fill_manager.address).call()
-        < request.amount
-    ):
-        func = token.functions.approve(context.fill_manager.address, allowance)
+    chain_id = ChainId(w3.eth.chain_id)
+    token_address = token.address
+    retrieved_mutex = context.fill_mutexes[(chain_id, token_address)]
+    with retrieved_mutex:
```
istankovic, line 333: Nitpick: I would just call this variable `mutex`, the `retrieved_` part is unnecessary.

beamer/agent/agent.py
bilbeyt: shouldnt they be destroyed when agent stopped and reinitialized?

bilbeyt: do you want to keep them all the time and dont replace them?

istankovic: > do you want to keep them all the time and dont replace them?

It depends. The question is: for how long should those locks live?

Issue #1395, opened 2023-01-19
Write L1 messenger contract


andfletcher:

fredo:
Needs to be part of the trusted call chain, and will directly call the encoded call on the RequestManager

andfletcher:
Done

PR #1532, opened 2023-02-28
Ethereum messenger contracts

### LINKED ISSUE #1494: Write an L2 Messenger contract for Ethereum

GabrielBuragev:
Closes #1494, #1395 

Didn't do any manual tests as we lack automated deployment support for Ethereum as another rollup.
This will be tackled in https://github.com/beamer-bridge/beamer/issues/1398

### DIFF
--- a/contracts/contracts/chains/ethereum/EthereumMessengers.sol
+++ b/contracts/contracts/chains/ethereum/EthereumMessengers.sol
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.12;
 
-import "OpenZeppelin/openzeppelin-contracts@4.8.0/contracts/access/Ownable.sol";
-
 import "../../../interfaces/IMessenger.sol";
 import "../../RestrictedCalls.sol";
 import "../../Resolver.sol";
@@ -28,7 +26,7 @@ contract EthereumL1Messenger is IMessenger, RestrictedCalls {
 // In order to keep naming consistent between our rollup specific contracts
 // we are using `EthereumL2Messenger` here as a name even if this contract
 // is supposed to be deployed on L1.
-contract EthereumL2Messenger is IMessenger, Ownable, RestrictedCalls {
+contract EthereumL2Messenger is IMessenger, RestrictedCalls {
     enum MessageStatus {
         UNDEFINED,
         DEFINED,
--- a/contracts/contracts/chains/ethereum/EthereumMessengers.sol
+++ b/contracts/contracts/chains/ethereum/EthereumMessengers.sol
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.12;
+
+import "OpenZeppelin/openzeppelin-contracts@4.8.0/contracts/access/Ownable.sol";
+
+import "../../../interfaces/IMessenger.sol";
+import "../../RestrictedCalls.sol";
+import "../../Resolver.sol";
+
+contract EthereumL1Messenger is IMessenger, RestrictedCalls {
+    function callAllowed(address, address) external pure returns (bool) {
+        // The call must be delivered by the Resolver
+        // Since we are on the same chain as the Resolver
+        // we should never see an execution of this function.
+        revert("Unexpected call to callAllowed");
+    }
+
+    // Simply forward the message to the target on L1
+    function sendMessage(address target, bytes calldata message)
+        external
+        restricted(block.chainid)
+    {
+        (bool sent, ) = target.call(message);
+        require(sent, "sending message failed");
+    }
+}
+
+// In order to keep naming consistent between our rollup specific contracts
+// we are using `EthereumL2Messenger` here as a name even if this contract
+// is supposed to be deployed on L1.
+contract EthereumL2Messenger is IMessenger, Ownable, RestrictedCalls {
+    enum MessageStatus {
+        UNDEFINED,
+        DEFINED,
+        RELAYED
+    }
+
+    Resolver public immutable resolver;
+
+    /// Maps message hashes to their status.
+    mapping(bytes32 => MessageStatus) public messageHashes;
+
+    constructor(address resolver_) {
+        resolver = Resolver(resolver_);
+    }
+
+    function callAllowed(address, address) external pure returns (bool) {
+        // Since we are on the same chain as the FillManager & RequestManager
+        // we should never see an execution of this function.
+        revert("Unexpected call to callAllowed");
+    }
+
+    // Store message instead of forwarding it.
+    // Can be forwarded, if ever needed, by calling `relayMessage`.
+    //
+    // .. seealso:: :sol:func:`relayMessage`
+    function sendMessage(address, bytes calldata message)
+        external
+        restricted(block.chainid)
+    {
+        bytes32 messageHash = keccak256(message);
+
+        require(
+            messageHashes[messageHash] != MessageStatus.DEFINED,
+            "Message is already registered"
+        );
+        require(
+            messageHashes[messageHash] != MessageStatus.RELAYED,
+            "Message has already been relayed"
+        );
+
+        messageHashes[messageHash] = MessageStatus.DEFINED;
+    }
+
+    function relayMessage(
+        bytes32 requestId,
+        bytes32 fillId,
+        uint256 sourceChainId,
+        address filler
+    ) external {
+        bytes memory message = abi.encodeCall(
+            Resolver.resolve,
+            (requestId, fillId, block.chainid, sourceChainId, filler)
+        );
+        bytes32 messageHash = keccak256(message);
+
+        require(
+            messageHashes[messageHash] != MessageStatus.UNDEFINED,
+            "Message not yet registered"
+        );
+        require(
+            messageHashes[messageHash] != MessageStatus.RELAYED,
+            "Message has already been relayed"
+        );
+
+        messageHashes[messageHash] = MessageStatus.RELAYED;
+
+        (bool sent, ) = address(resolver).call(message);
+        require(sent, "relaying message failed");
+    }
+}
--- a/contracts/contracts/chains/arbitrum/ArbitrumMessengers.sol
+++ b/contracts/contracts/chains/arbitrum/ArbitrumMessengers.sol
@@ -5,7 +5,6 @@ import "OpenZeppelin/openzeppelin-contracts@4.8.0/contracts/vendor/arbitrum/IArb
 import "OpenZeppelin/openzeppelin-contracts@4.8.0/contracts/vendor/arbitrum/IBridge.sol";
 import "OpenZeppelin/openzeppelin-contracts@4.8.0/contracts/vendor/arbitrum/IInbox.sol";
 import "OpenZeppelin/openzeppelin-contracts@4.8.0/contracts/vendor/arbitrum/IOutbox.sol";
-import "OpenZeppelin/openzeppelin-contracts@4.8.0/contracts/access/Ownable.sol";
 
 import "../../../interfaces/IMessenger.sol";
 import "../../RestrictedCalls.sol";
@@ -80,7 +79,7 @@ contract ArbitrumL1Messenger is IMessenger, RestrictedCalls {
     /* solhint-enable avoid-tx-origin */
 }
 
-contract ArbitrumL2Messenger is IMessenger, Ownable, RestrictedCalls {
+contract ArbitrumL2Messenger is IMessenger, RestrictedCalls {
     function callAllowed(address caller, address courier)
         external
         view


contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,112 @@
...
+        restricted(block.chainid)
+    {
+        target.call(message);
+    }
+}
+
+// In order to keep naming consistent between our rollup specific contracts
+// we are using `EthereumL2Messenger` here as a name even if this contract
+// is supposed to be deployed on L1.
+contract EthereumL2Messenger is IMessenger, Ownable, RestrictedCalls {
```
bilbeyt, line 31: you dont need Ownable again here as it is already there with RestrictedCalls

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    // Store message instead of forwarding it.
+    // Can be forwarded, if ever needed, by calling `relayMessage`.
+    //
+    // .. seealso:: :sol:func:`relayMessage`
+    function sendMessage(address, bytes calldata message)
+        external
+        restricted(block.chainid)
+    {
+        storeMessage(message);
```
istankovic, line 77: Is this inlined?

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,112 @@
...
+        DEFINED,
+        RELAYED
+    }
+
+    Resolver public immutable resolver;
+
+    /// Maps message hashes with their status.
+    mapping(bytes32 => MessageStatus) public messageHashes;
+
+    constructor(address resolver_) {
```
bilbeyt, line 43: why dont we use _resolver instead of resolver_ here ?

contracts/contracts/chains/ethereum/EthereumMessengers.sol
istankovic: > What does inlined mean here?

Does `solc`, the Solidity compiler, produce inlined code in this case, i.e. will it actually omit the `storeMessage` function and move its body into `sendMessage` in-line, as if it were written there?

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: No idea, saw this pattern being used in `ArbitrumL1Messenger`'s constructor.

@istankovic ?

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,112 @@
...
+    function callAllowed(address, address) external view returns (bool) {
+        // The call must be delivered by the Resolver
+        // Since we are on the same chain as the Resolver
+        // we should never see an execution of this function.
+        revert("Unexpected state.");
+    }
+
+    // Simply forward the message to the target on L1
+    /* solhint-disable avoid-tx-origin */
+    function sendMessage(address target, bytes calldata message)
```
istankovic, line 20: Do we need `    /* solhint-disable avoid-tx-origin */` here even though we don't mention `tx.origin` explicitly?

contracts/contracts/chains/ethereum/EthereumMessengers.sol
istankovic: `resolver_` is correct style here, per https://docs.soliditylang.org/en/v0.8.19/style-guide.html#avoiding-naming-collisions. So this is fine.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,112 @@
...
+        );
+
+        require(
+            messageHashes[messageHash] != MessageStatus.UNDEFINED,
+            "Message not yet registered."
+        );
+        require(
+            messageHashes[messageHash] != MessageStatus.RELAYED,
+            "Message has already been relayed."
+        );
```
bilbeyt, line 100: saw these two times maybe you can introduce a small check function for this.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,112 @@
...
+    Resolver public immutable resolver;
+
+    /// Maps message hashes with their status.
+    mapping(bytes32 => MessageStatus) public messageHashes;
+
+    constructor(address resolver_) {
+        resolver = Resolver(resolver_);
+    }
+
+    function callAllowed(address, address) external view returns (bool) {
```
bilbeyt, line 47: not sure if we need to write this again as restrictedcalls seemed enough to me

bilbeyt: also true for L1Messenger part

contracts/contracts/chains/ethereum/EthereumMessengers.sol
bilbeyt: ok.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: I guess. I am wondering if it has some meaning though since it is present in the `ArbitrumL2Messenger` too.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: You are right, forgot to remove this line.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: This is required as we are implementing the `IMessenger` interface.
The `callAllowed` plays an important role in the determining the trusted call chain.

In this case it just reverts because we are on the same chain and a call to this function should not be seen under normal operation of the protocol. 

This function is expected to be called here: https://github.com/beamer-bridge/beamer/blob/ethereum-messenger-contracts/contracts/contracts/RestrictedCalls.sol#L71

If you see the conditional statements you will notice that we should never reach that part of the code due to the fact that we are on the same chain (EthereumL1Messenger, EthereumL2Messenger, Resolver, FillManager are all on same chain in the "Ethereum as a rollup" case)

The code will always stop here: https://github.com/beamer-bridge/beamer/blob/ethereum-messenger-contracts/contracts/contracts/RestrictedCalls.sol#L63

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: They are not the same.

1. In the first case we are checking that a message is not yet defined or relayed. If it is, we should not allow overwriting same messages.
2. In the second case we check if the message is not undefined or relayed. If it is undefined or already relayed then we should not try to relay that message.

I did it like this so we can provide different error messages for these cases for better understanding of whats going on.

bilbeyt (approved):

contracts/contracts/chains/ethereum/EthereumMessengers.sol
bilbeyt: @istankovic ?

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: So, i tried running the solc compiler with `--ir-optimized` and i found out that the `storeMessage` is treated as another function that is being called inside `sendMessage`.

Should i move the body of the `storeMessage` inside of `sendMessage`? 
I only splitted this into its own function for better readability/testability of this contract.

I guess i can just squash the one into the other since we wont write any unit tests for this contract and instead include it in our E2E tests.

GabrielBuragev: I am currently assessing the difference in gas costs for the inlined an un-inlined version of this function to see if it really makes sense to inline the code.

GabrielBuragev: The result from the gas estimation made using `solc --gas` was `infinite gas` in both cases :D 
Any idea how i can get an absolute number here?

contracts/contracts/chains/ethereum/EthereumMessengers.sol
istankovic: Omitting `Ownable` should result in the same functional behaviour as having it explicitly present as I believe solc does The Right Thing(tm) during type resolution. Specifically, per https://docs.soliditylang.org/en/v0.8.19/contracts.html#multi-inheritance, Solidity uses C3 linearization (similar to Python), but note the order is reverse to the one used by Python: types that are "more-derived" should come later in the sequence.

I have no preference one way or the other, but if you're going to omit `Ownable` from this one, then I think we should also omit it from all other messenger contracts where it is mentioned explicitly.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
istankovic: > The result from the gas estimation made using `solc --gas` was `infinite gas` in both cases :D Any idea how i can get an absolute number here?

This is a know limitation in `solc` static gas analysis, please see e.g. https://ethereum.stackexchange.com/questions/29725/infinite-gas-estimation-from-solc-for-simple-function.

istankovic:
@GabrielBuragev could you please squash your fixup commits?

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,106 @@
...
+        // we should never see an execution of this function.
+        revert("Unexpected state.");
+    }
+
+    // Simply forward the message to the target on L1
+    function sendMessage(address target, bytes calldata message)
+        external
+        restricted(block.chainid)
+    {
+        target.call(message);
```
istankovic, line 23: You're not checking the return value of the call.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,106 @@
...
+        messageHashes[messageHash] = MessageStatus.RELAYED;
+
+        resolver.resolve(
+            requestId,
+            fillId,
+            block.chainid,
+            sourceChainId,
+            filler
+        );
+    }
```
istankovic, line 105: Why not do

```
address(resolver).call(data)
```
where `data` is the result of `abi.encodeCall` above, which you have to do anyway? We would save unnecessary stack operations that way and make it more robust against potential breakages where we change the arguments in one place, but not the other.

Of course, similar to a previous comment, `call` requires proper return value checks.

istankovic (changes requested):
There are still some things to fix.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,106 @@
...
+import "../../../interfaces/IMessenger.sol";
+import "../../RestrictedCalls.sol";
+import "../../Resolver.sol";
+
+contract EthereumL1Messenger is IMessenger, RestrictedCalls {
+    function callAllowed(address, address) external pure returns (bool) {
+        // The call must be delivered by the Resolver
+        // Since we are on the same chain as the Resolver
+        // we should never see an execution of this function.
+        revert("Unexpected state.");
```
istankovic, line 15: Nitpick: it's a bit strange to say that the state is unexpected (might be confused with contract state). I think something like `Unexpected call to callAllowed` would be more precise...

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: Thanks for sharing this great source of info @istankovic. 
Since we are not explicitly overriding methods of `Ownable` i guess its safe to proceed with omitting it.

Will do the same for `ArbitrumMessengers` as well.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: What should be the ideal behavior here?

```
(bool sent, ) = target.call(message);
require(sent, "failed to send Message.");
``` 

Is this sufficient?

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: Good idea 👍

contracts/contracts/chains/ethereum/EthereumMessengers.sol
GabrielBuragev: Indeed. 👍

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,101 @@
...
+contract EthereumL2Messenger is IMessenger, RestrictedCalls {
+    enum MessageStatus {
+        UNDEFINED,
+        DEFINED,
+        RELAYED
+    }
+
+    Resolver public immutable resolver;
+
+    /// Maps message hashes with their status.
```
istankovic, line 40: Nitpick:
```suggestion
    /// Maps message hashes to their status.
```

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,101 @@
...
+import "../../../interfaces/IMessenger.sol";
+import "../../RestrictedCalls.sol";
+import "../../Resolver.sol";
+
+contract EthereumL1Messenger is IMessenger, RestrictedCalls {
+    function callAllowed(address, address) external pure returns (bool) {
+        // The call must be delivered by the Resolver
+        // Since we are on the same chain as the Resolver
+        // we should never see an execution of this function.
+        revert("Unexpected call to `callAllowed`.");
```
istankovic, line 15: Nitpick: I would omit the backticks and the full stop at the end, to be consistent with the rest of our messages.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    // Simply forward the message to the target on L1
+    function sendMessage(address target, bytes calldata message)
+        external
+        restricted(block.chainid)
+    {
+        (bool sent, ) = target.call(message);
+        require(sent, "sending message failed.");
```
istankovic, line 24: Nitpick: I would omit full stop at the end, to be consistent with the rest of our messages.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,101 @@
...
+    mapping(bytes32 => MessageStatus) public messageHashes;
+
+    constructor(address resolver_) {
+        resolver = Resolver(resolver_);
+    }
+
+    function callAllowed(address, address) external pure returns (bool) {
+        // Since we are on the same chain as the FillManager & RequestManager
+        // we should never see an execution of this function.
+        revert("Unexpected call to `callAllowed`.");
```
istankovic, line 50: Nitpick: I would omit the backticks and the full stop at the end, to be consistent with the rest of our messages.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,101 @@
...
+        bytes32 messageHash = keccak256(message);
+
+        require(
+            messageHashes[messageHash] != MessageStatus.DEFINED,
+            "Message is already registered."
+        );
+        require(
+            messageHashes[messageHash] != MessageStatus.RELAYED,
+            "Message has already been relayed."
+        );
```
istankovic, line 70: Nitpick: I would omit full stop at the ends of both these messages, to be consistent with the rest of our messages.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,101 @@
...
+
+        messageHashes[messageHash] = MessageStatus.RELAYED;
+
+        (bool sent, ) = address(resolver).call(message);
+        require(sent, "relaying message failed.");
```
istankovic, line 99: Nitpick: I would omit the full stop at the end, to be consistent with the rest of our messages.

contracts/contracts/chains/ethereum/EthereumMessengers.sol
```diff
@@ -0,0 +1,101 @@
...
+        bytes32 messageHash = keccak256(message);
+
+        require(
+            messageHashes[messageHash] != MessageStatus.UNDEFINED,
+            "Message not yet registered."
+        );
+        require(
+            messageHashes[messageHash] != MessageStatus.RELAYED,
+            "Message has already been relayed."
+        );
```
istankovic, line 94: Nitpick: I would omit full stop at the ends of both these messages, to be consistent with the rest of our messages.

istankovic (approved):
Looks good!

contracts/contracts/chains/ethereum/EthereumMessengers.sol
istankovic: Yup.

PR #1539, opened 2023-03-02
beamer: tests: make HTTPProxy more generic

istankovic:
This is to allow arbitrary behaviour when making POST requests.

### DIFF
--- a/beamer/tests/agent/test_challenge.py
+++ b/beamer/tests/agent/test_challenge.py
@@ -9,7 +9,6 @@ import beamer.agent.agent
 from beamer.tests.constants import FILL_ID
 from beamer.tests.util import (
     EventCollector,
-    HTTPProxy,
     alloc_accounts,
     alloc_whitelisted_accounts,
     earnings,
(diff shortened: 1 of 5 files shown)


beamer/tests/agent/test_challenge.py
```diff
@@ -250,16 +249,6 @@ def test_challenge_4(request_manager, fill_manager, token, config):
...
-    proxy_l2b = HTTPProxy(config.rpc_urls["l2b"])
-    proxy_l2b.delay_rpc({"eth_getLogs": 3})
-    proxy_l2b.start()
-
-    l2b_rpc_url = "http://%s:%s" % (
-        proxy_l2b.server_address[0],
-        proxy_l2b.server_address[1],
-    )
-
-    config.rpc_urls["l2b"] = l2b_rpc_url
```
compojoom, line 262: Why is this no longer necessary?

istankovic: > Why is this no longer necessary?

I see no reason why that particular test should depend on the fact that `eth_getLogs` calls are delayed.
The test checks whether the agent correctly waits until (constant) fill wait time has passed, which is something determined by the state machine and must not be affected by the RPC performance.

compojoom (approved):
Looks awesome! Can't wait to write my new tests on top of that PR :)

PR #1541, opened 2023-03-02
.circleci: also include health in CI tests

istankovic:


.circleci/workflows.yml
```diff
@@ -173,11 +173,11 @@ jobs:
             - /home/circleci/.brownie
             - /home/circleci/.solcx
 
-  test-agent:
+  test-beamer:
     executor: vm
     steps:
       - initialize-environment
-      - run: poetry run brownie test beamer/tests/agent -v -G --cov beamer --cov-report=term
+      - run: poetry run brownie test beamer/tests/{agent,health} -v -G --cov beamer --cov-report=term
```
compojoom, line 180: Why does it have to be agent & health? Can't we just run with `beamer/tests`?

compojoom: The health tests just run in few seconds - I don't think that a new workflow will speed up things. Let's leave it like this.

PR #1551, opened 2023-03-06
Ethereum relayer service

### LINKED ISSUE #1496: Relayer for Ethereum
The relayer needs to get another feature to be able to trigger L1 resolution coming from Ethereum.

It would be nice to stay with the same interface.
The contract could store a mapping from request_id to filler. currently the relayer receives tx_hashes as keys for a particular l1_resolution.
We could go with the same approach or try to use the request_id which is also bytes32. 

on the receiving side there is nothing to do (unlike Arbitrum)

GabrielBuragev:
Closes #1496

### DIFF
--- a/relayer/.gitignore
+++ b/relayer/.gitignore
@@ -1,3 +1,4 @@
 node_modules/
 build/
 coverage/
+types-gen/
--- a/relayer/jest.config.ts
+++ b/relayer/jest.config.ts
@@ -8,6 +8,7 @@ export default async (): Promise<Config> => {
       "^@/(.*)$": "<rootDir>/src/$1",
       "^~/(.*)$": "<rootDir>/tests/$1",
     },
+    testPathIgnorePatterns: ["build"],
     resetMocks: true,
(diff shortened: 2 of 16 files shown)


manuelwedler (approved):
Looks perfect! Nothing to add. Thanks!

Just to be sure, I got right: The issues says you would need to adapt the CLI arguments. In the implementation instead, you retrieve the necessary parameters directly in the relayer from the event logs, correct?

GabrielBuragev:
@manuelwedler Thanks for the review!

That was the initial idea, but then we found a better approach in order to keep the existing relayer interface and CLI as can be seen in this comment:
> As discussed, we have a solution to maintain the interface between the agent and the relayer.
The agent will pass the tx_hash to the relayer.
The relayer will read the events emitted by the tx. From there the relayer is able to rebuild the encoded call and will call the EthereumL2Messenger for resolution

Hopefully this answers your question. I will update the issues initial description so it is easier to find this info.

Issue #1533, opened 2023-02-28
Implement new fee calculation in Request manager


fredo:
The new fee calculation is: 

`Fee = ((1 - chains[source].targetWeightingFactor) * chains[source].normalizingFactor + chains[target].targetWeightingFactor * chains[target].normalizingFactor) * tokens[token].referenceCost * LP_MARGIN + chains[target].inefficiencyCost`

## Chain struct
We now have an increasing number of values per chain. Besides the existing `finalityPeriods`, we will additionally have `normalizingFactor`, `TargetWeightingFactor`. 
Similar to how we did with `Token`, we should create a chain struct containing these values.

## function minLpFee(source_chain_id, target_chain_id, token)
minLpFee is now a view function calculating the minLpFee based on the above equation.

fredo:
Updated the initial comment. 

If nobody disagrees, I would go for incorporating the inefficiency cost into the transfer cost by adding the average amount of fill transactions, the system produces (successful and failing). This means, nothing to be done here.

PR #1560, opened 2023-03-09
feat: implement new minLpFee for RequestManager

bilbeyt:
Closes: #1533


istankovic:
Could you please post the diff of the gas costs for the affected contracts?

contracts/contracts/RequestManager.sol
```diff
@@ -59,12 +59,18 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
 
     struct Token {
         uint256 transferLimit;
-        uint96 minLpFee;
+        uint256 price;
```
fredo, line 62: same question as above?

contracts/contracts/RequestManager.sol
```diff
@@ -59,12 +59,18 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
         uint32 lpFeePPM;
         uint32 protocolFeePPM;
         uint96 collectedProtocolFees;
     }
 
+    struct Chain {
+        uint256 finalityPeriod;
```
fredo, line 69: this struct could be optimized to fit into one uint256 slot.

Suggestions:

finality period is denominated in seconds and will likely not be higher than 1M. So it would fit well in `uint32`

targetWeightPPM is < 1000000.

and the transferCost could be chosen to fill a uint256 slot but could probably also go down to `uint96`

contracts/contracts/RequestManager.sol
```diff
@@ -198,14 +214,36 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+        Chain storage targetChain = chains[targetChainId];
+        return
+            ((((1 - sourceChain.targetWeightPPM / 1_000_000) *
+                sourceChain.transferCost) /
+                1 ether +
+                ((targetChain.targetWeightPPM / 1_000_000) *
+                    targetChain.transferCost) /
+                1 ether) *
+                token.price *
+                (lpMargin + 100)) / 100;
```
fredo, line 234: Am I assuming correctly that you consider the token price to be without decimals? The token price should be represented in the lowest unit as there is no float type in solidity.

If you divide the costs by 1 ether wouldnt the outcome be 0, if `transferCost` < 1 ether?

fredo: I would probably consider lpMargin also to be PPM.

beamer/agent/state_machine.py
```diff
@@ -163,9 +161,8 @@ def _timestamp_is_l1_finalized(
     # FIXME: Remove this for the above once the contracts with the event are deployed
     target_chain_finality = context.finality_periods.get(target_chain_id)
     if target_chain_finality is None:
-        target_chain_finality = context.request_manager.functions.finalityPeriods(
-            target_chain_id
-        ).call()
+        target_chain = context.request_manager.functions.chains(target_chain_id).call()
```
fredo, line 164: This looks like an inefficiency in the old code, but since we store and update the finality period in memory, we can reuse `context.finality_periods[target_chain_id]` instead of fetching it with another network call here

beamer/tests/conftest.py
```diff
@@ -79,6 +79,21 @@ def challenge_period_extension():
     return 50
 
 
+@pytest.fixture
+def lp_margin():
+    return 30
```
fredo, line 84: as suggested I would probably use PPM so that we stay consistent.

scripts/deployment/arbitrum-local-template.json
```diff
@@ -9,25 +9,33 @@
...
                 "claim_period": 3600,
-                "challenge_period_extension": 3600
+                "challenge_period_extension": 3600,
+                "lp_margin": 30
             },
             "tokens":
                 [
                     {
                         "token_address": "mintable_token",
+                        "price": 1000,
```
fredo, line 25: even though it's an test config, I think it looks a bit misleading that we choose 1000 here but in the tests we use ..e18 representation. 
I know it doesn't really matter here but we should somehow state clearly what price means in this context.

fredo:
The more i read through the PR the more I have my troubles with the parameter `price`. 
Although in general it is an acceptable term, what we are really interested in is the ETH/token conversion rate. IMO price brings this more into the context of buying/selling where of course you also need a conversion rate to determine a price but I kind of have the feeling that something like `ethRate` or `rate` would fit more. Especially very often prices are converted into FIAT and we will always have a `eth/token` rate no matter what the token is.

`ethRate` would probably very explicit of what it is supposed to mean because we apply this parameter in the context of a token.

scripts/call_contracts.py
```diff
@@ -349,7 +355,7 @@ def whitelist(
 @cli.command("update-token")
 @click.argument("token-address", type=str, metavar="ADDRESS", callback=validate_address)
 @click.option("--transfer-limit", type=int, help="New transfer limit of token")
-@click.option("--min-lp-fee", type=int, help="New minimum lp fee")
+@click.option("--rate", type=int, help="ETH/token conversion rate")
```
istankovic, line 358: Nitpick: I'd prefer a more explicit option name and help text, e.g.

```python
@click.option("--eth-per-token", type=int, help="ETH per token, in wei")
```

fredo: wouldnt it be token per eth then?

fredo: Unfortunately finance languange describes price feeds the other way around (ETH/USD means 1800 USD per ETH)

istankovic: Perfectly illustrates why I am asking this. :-)

istankovic: I don't really care whichever way, but making it more explicit in terms of `--X-per-Y` would be much more obvious, at least for me. I don't really care what X and Y are.

fredo: or I call it simply `eth-token-rate`?

fredo: ok I will write `token-per-eth` but writing it with `/` should be `ETH/token`

istankovic: > Unfortunately finance languange describes price rates the other way around (ETH/USD means 1800 USD per ETH)

I have no idea how relevant the source is, but [this](https://saylordotorg.github.io/text_international-finance-theory-and-policy/s07-02-exchange-rate-definitions.html) says:

> It is important to note that the value of a currency is always given in terms of another currency. Thus the value of a U.S. dollar in terms of British pounds is the £/$ exchange rate. The value of the Japanese yen in terms of dollar is the $/¥ exchange rate.

Given that, I would interpret `ETH/USD` as `the value of 1 USD, in terms of ETH`. Which is how I have always been interpreting ratios.

fredo: > I'm really lacking a financial background so I can't really say what rate is supposed to mean, but from the discussion so far it seems to mean exactly the inverse of what a regular mathematical ration (X / Y) means.

yes exactly, it is the exact opposite of the mathematical ratio.

fredo: > Alternatively, we could go with just rate as in your original approach, if that is what you'd prefer, but I am afraid we would need to be crystal clear (i.e. by documenting) what it actually means.

This is actually where I think, most people would completely understand the conversion rate as done financial lingo. Or rather I would assume most people could follow this approach. 
Knowing that it is the exact opposite of how it is expressed in mathematical terms. 

To be more precise, I think for `eth-token-rate` most people would understand `how many token are equivalent to 1 ETH`. I also do understand that this is completely counter intuitive for you, if you only have a mathematical background.
I think it's a dilemma here

fredo: I'm actually leaning towards using the above mentioned rate, so the opposite to the mathematical term, as I think it will be understood by the broader set of people intuitively.
Also if people want to compare the configured rate to the current rate somewhere online, they will find the same value under the ticker `ETH/Token`.

istankovic: Yeah, after some searching online, I think most places interpret `ETH/USD` as `ETH in USD` or `ETH to USD`, so I guess it makes much more sense to interpret the `/` as meaning `in` or `to` in a financial context.

In that light, I'm fine with your suggestion `eth-token-rate`.

(We could also consider e.g. `--eth-to-token-rate` for a tiny bit more explicitness, but I don't mind.)

BTW, I'm curious what you think about my comment [here](https://github.com/beamer-bridge/beamer/pull/1560#discussion_r1147326658).

fredo: > BTW, I'm curious what you think about my comment [here](https://github.com/beamer-bridge/beamer/pull/1560#discussion_r1147326658).

Ok, interesting. BTW, I completely understand where you are coming from and mathematically spoken I completely agree.
According to your link it seems to me that forex (foreign exchange) markets again have their own common lingo, which is again how ratios are interpreted mathematically. (Also it makes mathematically more sense because we end up with the unit we want to end in). 
However, as you also have pointed out, I think in the general financial context it is more common sense to interpret it as `ETH to token`.
Given the fact that it seems that even within the financial markets there is different interpretation I would suggest to go with the more explicit `--eth-to-token` naming.
Do you agree?

istankovic: > `--eth-to-token-rate` or just `eth-in-token`?

Whichever you prefer, I don't have a strong preference here.

contracts/contracts/RequestManager.sol
```diff
@@ -119,19 +125,27 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
-    event FinalityPeriodUpdated(uint256 targetChainId, uint256 finalityPeriod);
-
     /// Emitted when token object of a token address is updated.
     ///
     /// .. seealso:: :sol:func:`updateToken`
     event TokenUpdated(
         address tokenAddress,
         uint256 transferLimit,
-        uint96 minLpFee,
+        uint256 rate,
```
istankovic, line 134: Should we also consider renaming this field, in light of the other "ETH/USD" discussion?

fredo: yes, I would rename all occurences of rate throughout the code base

bilbeyt:
i think there is one missing function in contract for max transferable amount or so for fe as we discussed before

fredo:
> i think there is one missing function in contract for max transferable amount or so for fe as we discussed before

ah right good point !

contracts/contracts/RequestManager.sol
istankovic: I'm curious, what was the outcome of this discussion?
I imagine both `finalityPeriod` and `targetWeightPPM`could be smaller, e.g. `uint128`, which would reduce the number of slots required by 1.

On the other hand, I'm also ok with this, but would like to know the reasoning of the previous discussion.

contracts/contracts/RequestManager.sol
```diff
@@ -326,7 +378,7 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
         newRequest.targetChainId = targetChainId;
         newRequest.amount = amount;
         newRequest.validUntil = uint32(block.timestamp + validityPeriod);
-        newRequest.lpFee = uint96(lpFeeTokenAmount);
+        newRequest.lpFee = lpFeeTokenAmount;
         newRequest.protocolFee = uint96(protocolFeeTokenAmount);
```
istankovic, line 411: Shouldn't we also remove the cast to `uint96` for `protocolFee`?

contracts/contracts/RequestManager.sol
```diff
@@ -247,16 +291,19 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
     /// @param _claimPeriod Claim period, in seconds.
     /// @param _challengePeriodExtension Challenge period extension, in seconds.
+    /// @param _lpMarginPPM Liquidity provider profit margin.
     constructor(
         uint96 _claimStake,
         uint256 _claimRequestExtension,
         uint256 _claimPeriod,
-        uint256 _challengePeriodExtension
+        uint256 _challengePeriodExtension,
+        uint256 _lpMarginPPM
```
istankovic, line 329: I'm wondering about whether we need this in the constructor, or rather, whether lpMarginPPM needs to be immutable. For instance, we have
```
        uint32 lpFeePPM;
        uint32 protocolFeePPM;
```
as part of `Token`, which I was wondering about above.
If we don't really need these two values to be per-token, or per-chain, then we can consider all three

```
        uint32 lpFeePPM;
        uint32 protocolFeePPM;
        uint32 lpMarginPPM
```
as a separate contract-level part of our fee model. (I don't have a nice name for such a struct, but we probably can also go without the struct too...) Also, this would allow `lpMarginPPM` to be mutable the same way the other two fields are.

fredo: Good point, I'm thinking about why we decided to put it into the token struct.
At the end, it's the users demand which will decide if different lpFees are necessary for different tokens. 
For now, I'm actually happy to move them out of the struct and have it protocol wide.

Regarding mutability, I do see that the lpFee and protocolFee are more subject to be changed regularly than the lpMargin on the minLpFee, but this also may be just my subjective view.
I have nothing against making it mutable.

istankovic: > Good point, I'm thinking about why we decided to put it into the token struct, It probably just moved their because of the minLpFee which was token dependent.
> At the end, it's the users demand which will decide if different lpFees are necessary for different tokens.
> For now, I'm actually happy to move them out of the struct and have it protocol wide.

Cool, then let's have all three of these
```
        uint32 lpFeePPM;
        uint32 protocolFeePPM;
        uint32 lpMarginPPM
```
on the contract level and move `lpMarginPPM` out of the constructor.
We would then need another function to update these, e.g. `updateFees`, right?
So we end up with
```
function updateToken(address tokenAddress, uint256 transferLimit, uint256 ethInToken)
function updateChain(uint256 chainId, uint256 finalityPeriod, uint256 transferCost, uint256 targetWeightPPM)
function updateFees(uint32 lpFeePPM, uint32 protocolFeePPM, uint32 lpMarginPPM)
```
where that last one would be called the least often.

So after deployment we call all three to configure the contract, and then only the first two(?) on a, say, weekly basis.

beamer/agent/events.py
```diff
@@ -67,9 +67,11 @@ class TxEvent(Event):
 
 
 @dataclass(frozen=True)
-class FinalityPeriodUpdated(TxEvent, SourceChainEvent):
-    target_chain_id: ChainId
+class ChainUpdated(TxEvent, SourceChainEvent):
+    chain_id: ChainId
     finality_period: int
+    target_weight_p_p_m: int
```
istankovic, line 73: Nitpick: I think this would be more readable as `target_weight_ppm`

beamer/agent/state_machine.py
```diff
@@ -157,14 +157,11 @@ def _proof_ready_for_l1_relay(request: Request) -> bool:
...
-    # target_chain_finality = context.finality_periods[target_chain_id]
 
-    # FIXME: Remove this for the above once the contracts with the event are deployed
     target_chain_finality = context.finality_periods.get(target_chain_id)
-    if target_chain_finality is None:
-        target_chain = context.request_manager.functions.chains(target_chain_id).call()
-        target_chain_finality = target_chain[0]
+    # The internal memory should not be none because at this point
+    # the ChainUpdated event must have arrived, otherwise there would be no request
+    assert target_chain_finality is not None
```
istankovic, line 164: Nice!

beamer/tests/contracts/test_request_manager_fees.py
```diff
@@ -200,19 +212,17 @@ def test_different_fees(request_manager, token, claim_period, claim_stake):
...
     with temp_fee_data(request_manager, token, *fee_data_2):
         request_id_2 = make_request(
             request_manager, token, requester, requester, amount, fee_data="standard"
         )
-        lp_fee_2 = request_manager.lpFee(token.address, amount)
+        lp_fee_2 = request_manager.lpFee(source_chain_id, target_chain_id, token.address, amount)
         protocol_fee_2 = request_manager.protocolFee(token.address, amount)
-        assert lp_fee_2 == 36e15
-        assert protocol_fee_2 == 279e15
+        assert protocol_fee_2 > lp_fee_2
```
istankovic, line 225: Here, as above, we are checking for inequality, whereas before we checked for equality.
I am not sure whether we have another test that makes sure that the fees that are computed by the contract are exactly what they are supposed to be (i.e. the test that does the same calculation as the contract). But if we don't have such a test, we should have one, in order to catch changes to the fee model. I am not sure if this is such a test, but it feels weird that we are moving toward a more relaxed check here. Note: elsewhere it'd probably be fine because it's just a detail, but here we are in `test_request_manager_fees`, which is probably the only place that cares about fees.

scripts/deployment/arbitrum-local-template.json
```diff
@@ -9,21 +9,24 @@
             "name": "Arbitrum",
             "rpc": "http://localhost:8547",
             "chain_id": 412346,
-            "finality_period": 3600,
             "l1_messenger": ["ArbitrumL1Messenger", ${l1_messenger_args}],
             "l2_messenger": "ArbitrumL2Messenger",
+            "finality_period": 3600,
+            "transfer_cost": 0,
+            "target_weight_ppm": 0,
```
istankovic, line 16: Is there are reason why we set both `transfer_cost` and `target_weight_ppm` to zero?
I think nonzero values for these should work and would be more realistic (this is being used in our Arbitrum E2E test).

scripts/deployment/ethereum-local.json
```diff
@@ -9,20 +9,23 @@
             "name": "Ethereum on ganache L1",
             "rpc": "http://localhost:8545",
             "chain_id": 1337,
-            "finality_period": 3600,
+            "finality_period": 1,
+            "transfer_cost": 0,
+            "target_weight_ppm": 0,
```
istankovic, line 14: Similarly here.

scripts/deployment/goerli-testnets.json
```diff
@@ -68,17 +74,20 @@
       ],
       "l2_messenger": "OptimismL2Messenger",
       "finality_period": 1800,
+      "transfer_cost": "",
+      "target_weight_ppm": "",
```
istankovic, line 78: Here we use a string, elsewhere an integer `0`.

scripts/deployment/goerli-testnets.json
```diff
@@ -16,17 +16,20 @@
       ],
       "l2_messenger": "ArbitrumL2Messenger",
       "finality_period": 18000,
+      "transfer_cost": "",
+      "target_weight_ppm": "",
```
istankovic, line 20: Here we use a string, elsewhere an integer `0`.

scripts/deployment/mainnet.json
```diff
@@ -16,24 +16,27 @@
             ],
             "l2_messenger": "ArbitrumL2Messenger",
             "finality_period": 691200,
+            "transfer_cost": "",
+            "target_weight_ppm": "",
```
istankovic, line 20: Here we use a string, elsewhere an integer `0`.

scripts/deployment/mainnet.json
```diff
@@ -46,25 +49,28 @@
             "l1_messenger": ["OptimismL1Messenger", "0x25ace71c97B33Cc4729CF772ae268934F7ab5fA1"],
             "l2_messenger": "OptimismL2Messenger",
             "finality_period": 604800,
+            "transfer_cost": "",
+            "target_weight_ppm": "",
```
istankovic, line 53: Here we use a string, elsewhere an integer `0`.

scripts/deployment/mainnet.json
```diff
@@ -77,25 +83,28 @@
             "l1_messenger": ["OptimismL1Messenger", "0x6D4528d192dB72E282265D6092F4B872f9Dff69e"],
             "l2_messenger": "OptimismL2Messenger",
             "finality_period": 604800,
+            "transfer_cost": "",
+            "target_weight_ppm": "",
```
istankovic, line 87: Here we use a string, elsewhere an integer `0`.

beamer/tests/contracts/test_request_manager.py
```diff
@@ -1418,3 +1418,19 @@ def test_transfer_limit_requests(request_manager, token):
...
+    source_chain_cost = (1_000_000 - chain_params[2]) * chain_params[1]
+    target_chain_cost = target_chain_params[2] * target_chain_params[1]
+    min_lp_fee = (
+        (source_chain_cost + target_chain_cost)
+        * token_params[1]
+        * (1_000_000 + lp_margin_ppm)
+        / 1e30
+    )
+
+    assert min_lp_fee == request_manager.minLpFee(ape.chain.chain_id, 999, token.address)
```
istankovic, line 1436: Nitpick: since this is concerned with the fees and this file is already too big, we should move this to `test_request_manager_fees.py`.

beamer/agent/state_machine.py
fredo: I just followed other examples of our codebase, but I can change that.

beamer/tests/contracts/test_request_manager_fees.py
fredo: I think it wasn't me who changed this check. I have to look into this function.

scripts/deployment/arbitrum-local-template.json
fredo: The reason for it is, that our e2e tests only work with `0` fees. I used the transfer cost to nail down the fees to 0.
With the changes about the fee structure in the contract, I'm going to pin the `lpMarginPPM` to 0 so we can choose arbitrary values for the other parameters.
In general we could choose any value, but they influence when minFee or when lpFeePPM is applied, so we need to be careful whenever we apply fees (which isn't the case for e2e tests)

fredo: regarding mainnet.json, I left it open on purpose, because we did this for all keys which are not clear yet, or have to be added upon deployment. Let me know if you think otherwise. But I'd like to let the mainnet.json fail if we did not modify the necessary values

scripts/deployment/mainnet.json
fredo: I changed it to fictitious values for all test configs and left it to `""` for `mainnet.json`, so that we are forced to put in numbers upon deployment.

beamer/tests/contracts/test_request_manager_fees.py
bilbeyt: i changed it back to use exact values, i didnt want to calculate exact values before but it is ok now.

contracts/contracts/RequestManager.sol
```diff
@@ -268,6 +268,29 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+            lpFee(sourceChainId, targetChainId, tokenAddress, amount) >= minFee
+        ) {
+            return transferableAmount;
+        }
+
+        return
+            (amount * 1_000_000) /
+            (1_000_000 + token.lpFeePPM + token.protocolFeePPM);
+    }
+
```
istankovic, line 289: Do we have a test for this function?

contracts/contracts/RequestManager.sol
```diff
@@ -764,42 +833,64 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+    /// @param _lpFeePPM LP percentage fee applied on transfer amount denominated in parts per million
+    /// @param _protocolFeePPM Protocol fee applied on transfer amount denominated in parts per million
+    /// @param _protocolFeePPM Protocol fee applied on transfer amount denominated in parts per million
+    function updateFees(
+        uint32 _minFeePPM,
+        uint32 _lpFeePPM,
+        uint32 _protocolFeePPM
+    ) external onlyOwner {
+        require(_lpFeePPM <= 999_999, "Maximum PPM of 999999 exceeded");
+        require(_protocolFeePPM <= 999_999, "Maximum PPM of 999999 exceeded");
```
istankovic, line 889: We're missing another `require` in the same way for `minFeePPM`.

contracts/contracts/RequestManager.sol
```diff
@@ -198,30 +216,80 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+    }
+
+    function transferableAmount(
+        uint256 sourceChainId,
+        uint256 targetChainId,
         address tokenAddress,
         uint256 amount
     ) public view returns (uint256) {
-        return lpFee(tokenAddress, amount) + protocolFee(tokenAddress, amount);
+        Token storage token = tokens[tokenAddress];
```
istankovic, line 275: `token` is unused in this function so we can remove this line.

contracts/contracts/RequestManager.sol
fredo: I left it open on purpose. The reason for it is that it is very unlikely that we will apply a percentage fee higher than 100%. But as for the minFee I see situations where we want to apply a fee greater than 100%. Not sure that it will ever happen, but naturally the minFeePPM is much larger than lpFeePPM and protocolFeePPM just to begin with.

contracts/contracts/RequestManager.sol
fredo: will add tests for our view functions similar to minLpFee. I think we didnt have any.

contracts/contracts/RequestManager.sol
```diff
@@ -119,17 +123,23 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
     ///
     /// .. seealso:: :sol:func:`updateToken`
     event TokenUpdated(
         address tokenAddress,
         uint256 transferLimit,
-        uint96 minLpFee,
-        uint32 lpFeePPM,
-        uint32 protocolFeePPM
+        uint256 ethInToken
+    );
```
istankovic, line 133: While we are here, I think it would be nice to make `address tokenAddress` `indexed`.

contracts/contracts/RequestManager.sol
```diff
@@ -119,17 +123,23 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+
+    /// Emitted when chain object of a chain id is updated.
+    ///
+    /// .. seealso:: :sol:func:`updateChain`
+    event ChainUpdated(
+        uint256 chainId,
+        uint256 finalityPeriod,
+        uint256 transferCost,
+        uint256 targetWeightPPM
     );
```
istankovic, line 143: Similar to above, I think it would be nice to make `uint256 chainId` `indexed`.

beamer/tests/contracts/test_request_manager_fees.py
```diff
@@ -189,28 +185,31 @@ def test_different_fees(request_manager, token, claim_period, claim_stake):
...
             request_manager, token, requester, requester, amount, fee_data="standard"
         )
-        lp_fee_1 = request_manager.lpFee(token.address, amount)
-        protocol_fee_1 = request_manager.protocolFee(token.address, amount)
-        assert lp_fee_1 == 6e18
-        assert protocol_fee_1 == 63e15
+        lp_fee_1 = request_manager.lpFee(target_chain_id, token.address, amount)
+        protocol_fee_1 = request_manager.protocolFee(amount)
+
+        assert lp_fee_1 > protocol_fee_1
```
bilbeyt, line 204: note to myself: fix this

scripts/deployment/mainnet.json
```diff
@@ -46,25 +49,28 @@
...
+                        "transfer_limit": 10000,
+                        "eth_in_token": "",
                         "lp_fee_ppm": 3000,
                         "protocol_fee_ppm": 0
                     },
                     {
                         "token_address": "0xDA10009cBd5D07dd0CeCc66161FC93D7c9000da1",
                         "transfer_limit": 100000000000000000000,
-                        "min_lp_fee": 1000000000000000000,
+                        "rate": 1000000000000000000,
```
bilbeyt: should be wrong there is no rate anymore

scripts/deployment/mainnet.json
```diff
@@ -46,25 +49,28 @@
             "l1_messenger": ["OptimismL1Messenger", "0x25ace71c97B33Cc4729CF772ae268934F7ab5fA1"],
             "l2_messenger": "OptimismL2Messenger",
             "finality_period": 604800,
+            "transfer_cost": "",
```
bilbeyt, line 52: is this working like this? it is written as int in apischema validation

bilbeyt: also true for some other string representations here

beamer/agent/events.py
istankovic: We can have this field named `target_weight_ppm` more nicely by improving the `_camel_to_snake` function in the same module:

```python
def _camel_to_snake(s: str) -> str:                                                 
    snake = "".join(                                                                
        "%c_" % current if current.islower() and next.isupper() else current.lower()
        for current, next in pairwise(s)                                            
    )                                                                               
    # The last character isn't included in snake due to how pairwise() works,       
    # so add it manually.                                                           
    return snake.lstrip("_") + s[-1].lower()                                        
```

(Needs `from itertools import pairwise` at the top.)

contracts/contracts/RequestManager.sol
```diff
@@ -764,42 +818,64 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+    }
+
+    /// Update fees
+    ///
+    /// .. note:: This function can only be called by the contract owner.
+    ///
+    /// @param _minFeePPM Margin which is going to be applied to the minLpFee
+    /// @param _lpFeePPM LP percentage fee applied on transfer amount denominated in parts per million
+    /// @param _protocolFeePPM Protocol fee applied on transfer amount denominated in parts per million
+    /// @param _protocolFeePPM Protocol fee applied on transfer amount denominated in parts per million
```
bilbeyt, line 867: there is a duplicate param

contracts/contracts/RequestManager.sol
```diff
@@ -198,30 +216,66 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+        uint256 minFee = minLpFee(targetChainId, tokenAddress);
+        require(amount > minFee, "Amount not high enough to cover the fees");
+        uint256 transferableAmount = (amount * 1_000_000) /
+            (1_000_000 + lpFeePPM + protocolFeePPM);
+
+        if (lpFee(targetChainId, tokenAddress, amount) >= minFee) {
+            return transferableAmount;
+        }
+
+        return (amount * 1_000_000) / (1_000_000 + lpFeePPM + protocolFeePPM);
```
bilbeyt, line 278: this also equals to the transferableAmount then why we have if above?

contracts/contracts/RequestManager.sol
```diff
@@ -198,30 +216,66 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
         uint256 amount
     ) public view returns (uint256) {
-        return lpFee(tokenAddress, amount) + protocolFee(tokenAddress, amount);
+        return lpFee(targetChainId, tokenAddress, amount) + protocolFee(amount);
+    }
+
+    function transferableAmount(
+        uint256 targetChainId,
+        address tokenAddress,
+        uint256 amount
```
bilbeyt, line 267: is this amount will be actually user total balance ?

beamer/tests/contracts/test_request_manager_fees.py
```diff
@@ -246,16 +246,73 @@ def test_different_fees(request_manager, token, claim_period, claim_stake):
...
     with ape.reverts("Maximum PPM of 999999 exceeded"):
-        request_manager.updateToken(token.address, 0, 0, ppm, 0)
+        request_manager.updateFees(0, ppm, 0)
 
     with ape.reverts("Maximum PPM of 999999 exceeded"):
-        request_manager.updateToken(token.address, 0, 0, 0, ppm)
+        request_manager.updateFees(0, 0, ppm)
+
+
+@pytest.mark.parametrize("transfer_cost", [int(400e12)])
```
bilbeyt, line 263: i think we didnt use this right ?

beamer/tests/contracts/test_request_manager_fees.py
```diff
@@ -246,16 +246,73 @@ def test_different_fees(request_manager, token, claim_period, claim_stake):
...
+def test_total_fee(request_manager, token, chain_params):
+    target_chain_params = chain_params[0], int(3e12), 750_000
+    request_manager.updateChain(999, *target_chain_params)
+
+    min_lp_fee = request_manager.minLpFee(999, token.address)
+    # find the value where lpFee == minLpFee to test with values lower and higher
+    conjunction_amount = int(min_lp_fee * 1_000_000 / (request_manager.lpFeePPM()))
+    test_amounts = [int(conjunction_amount / 2), conjunction_amount * 2]
```
istankovic, line 288: Nitpick: instead of doing `int(...)` conversion, you can use the floor division operator `//`: `conjunction_amount // 2`.

beamer/tests/contracts/test_request_manager_fees.py
```diff
@@ -246,16 +246,73 @@ def test_different_fees(request_manager, token, claim_period, claim_stake):
...
+
+    min_lp_fee = request_manager.minLpFee(target_chain_id, token.address)
+    # find the value where lpFee == minLpFee to test with values lower and higher
+    conjunction_amount = int(min_lp_fee * 1_000_000 / (request_manager.lpFeePPM()))
+
+    test_amounts = [
+        int(conjunction_amount / 2),
+        conjunction_amount * 2,
+        randint(1, 2**63),
+    ]
```
istankovic, line 309: I would also include just `conjunction_amount` as a test point.

contracts/contracts/RequestManager.sol
istankovic: > is this amount will be actually user total balance ?

Yes, I assume our frontend will use this function when the user presses the "max" button.

contracts/contracts/RequestManager.sol
```diff
@@ -764,42 +821,64 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
     ///
     /// .. note:: This function can only be called by the contract owner.
     ///
-    /// @param targetChainId The target chain ID.
-    /// @param finalityPeriod Finality period in seconds.
-    function setFinalityPeriod(
-        uint256 targetChainId,
-        uint256 finalityPeriod
+    /// @param chainId The chain ID of the chain.
+    /// @param finalityPeriod The finality period of the chain.
```
istankovic, line 838: Nitpick: I would keep the `in seconds` part that was here before, in order to make it clear that the period is in seconds.

contracts/contracts/RequestManager.sol
fredo: totalFee can be applied with any value, namely the value the user chooses to send

beamer/agent/events.py
fredo: can you please create its own PR for it once we have it merged?

contracts/contracts/RequestManager.sol
```diff
@@ -764,42 +821,64 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
...
+        uint32 _protocolFeePPM
+    ) external onlyOwner {
+        require(_lpFeePPM <= 999_999, "Maximum PPM of 999999 exceeded");
+        require(_protocolFeePPM <= 999_999, "Maximum PPM of 999999 exceeded");
 
-        emit FinalityPeriodUpdated(targetChainId, finalityPeriod);
+        minFeePPM = _minFeePPM;
+        lpFeePPM = _lpFeePPM;
+        protocolFeePPM = _protocolFeePPM;
     }
```
istankovic, line 882: Is it intentional that we don't emit a `FeesUpdated` event?

scripts/deployment/config.py
```diff
@@ -30,13 +30,17 @@ class _RequestManagerArgs:
     challenge_period_extension: int = field(metadata=schema(min=0))
 
 
+class _Fees:
```
istankovic, line 33: Missing `@dataclass` decorator.

contracts/contracts/RequestManager.sol
fredo: I just saw it as well, will add

beamer/tests/contracts/test_request_manager_fees.py
```diff
@@ -246,16 +246,73 @@ def test_different_fees(request_manager, token, claim_period, claim_stake):
...
+
+    min_lp_fee = request_manager.minLpFee(999, token.address)
+    # find the value where lpFee == minLpFee to test with values lower and higher
+    conjunction_amount = int(min_lp_fee * 1_000_000 / (request_manager.lpFeePPM()))
+    test_amounts = [int(conjunction_amount / 2), conjunction_amount * 2]
+
+    for amount in test_amounts:
+        assert request_manager.totalFee(999, token.address, amount) == request_manager.lpFee(
+            999, token.address, amount
+        ) + request_manager.protocolFee(amount)
```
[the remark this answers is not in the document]
fredo: BTW, I chose /2 and * 2 because I saw rounding problems around the conjunction amount

beamer/agent/events.py
istankovic: > can you please create its own PR for it once we have it merged?

No problem, man. :+1:

beamer/tests/contracts/test_request_manager_fees.py
fredo: it is used by other fixtures (it will be used during deployment)

beamer/tests/contracts/test_request_manager_fees.py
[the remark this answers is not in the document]
istankovic: > BTW, I chose /2 and * 2 because I saw rounding problems around the conjunction amount

The reason why I am suggesting this is to make sure that the contract computation really switches between the "minLpFee" mode and normal mode. I'm guessing if you choose params suitably, we should be able to tell the difference between `conjunction_amount` and `conjunction_amount + 1`.

istankovic (approved):

istankovic:
Epic effort. :heart:

PR #1578, opened 2023-03-15
Add multi-token support to monitoring script

### LINKED ISSUE #1525: Add DAI to agent monitoring script
## Description
We have / will support DAI bridges. Since we want to ensure that Agents are also informed about their DAI positions, we must add it to the monitoring script.

## Acceptance Criteria
- [ ] We also show DAI and ChainID on which the DAI is in the output of the monitoring script for the agent address

GabrielBuragev:
closes #1525 

Note: Any future token will be easily supported by simply adding it to the list of tokens in the config file

### DIFF
--- a/beamer/health/check.py
+++ b/beamer/health/check.py
@@ -11,15 +11,20 @@ from eth_utils import to_checksum_address
 from typing_extensions import NotRequired
 from web3 import Web3
 from web3.constants import ADDRESS_ZERO
-from web3.middleware import geth_poa_middleware
 
 import beamer.agent.contracts
 import beamer.agent.events
 from beamer.agent.config import _merge_dicts
 from beamer.agent.events import ClaimMade, DepositWithdrawn, RequestCreated, RequestFilled
 from beamer.agent.typing import ChainId
-from beamer.agent.util import load_ERC20_abi
 from beamer.health.notify import Message, NotificationConfig, NotificationState, Notify
+from beamer.health.util import (
+    TokenDetails,
+    get_token_amount_in_decimals,
+    get_token_balance,
+    get_token_details,
+    make_web3,
+)
 
 
 class NotificationTypes:
@@ -30,7 +35,8 @@ class NotificationTypes:
     CHALLENGE_GAME_CLAIMED_BY_SOMEONE_ELSE = "ChallengeGameClaimedBySomeoneElse"
 
 
-TokenMap = dict[str, list[tuple[str, str]]]
+TokenMap = dict[str, list[list[str]]]
+TokenDetailsMap = dict[str, TokenDetails]
 
 
 class HealthConfig(TypedDict):
@@ -55,6 +61,8 @@ TransferMap = DefaultDict[str, Transfer]
 
 ChainEventMap = dict[ChainId, list[beamer.agent.events.Event]]
 
+TokenVolume = dict[str, int]
+
 
 class NotificationMeta(TypedDict):
     request_id: str
@@ -77,13 +85,13 @@ class Stats:
 
 @dataclass
 class TransferStats:
-    volume: int = 0
+    volume: TokenVolume = field(default_factory=TokenVolume)
     fills: int = 0
     requests: int = 0
     withdrawals: int = 0
     claims: int = 0
     expired_requests: int = 0
-    expired_requests_volume: int = 0
+    expired_requests_volume: TokenVolume = field(default_factory=TokenVolume)
 
 
 @dataclass
@@ -91,11 +99,26 @@ class Context:
     stats: TransferStats = field(default_factory=TransferStats)
     notifications: list = field(default_factory=list)
     agent_address: str = ""
-
+    token_deployments: TokenMap = field(default_factory=TokenMap)
+    tokens: TokenDetailsMap = field(default_factory=dict)
     notification_state: NotificationState = NotificationState(Path("chain/"))
 
     def add_notification(self, message: Notification) -> None:
         self.notifications.append(message)
+
+    def parse_tokens(self, token_deployments: TokenMap, rpcs: dict[int, str]) -> None:
+        self.token_deployments = token_deployments
+        for token_symbol, deployments in token_deployments.items():
+            chain_id = int(deployments[0][0])
+            token_address = deployments[0][1]
+            rpc = rpcs[chain_id]
+            # we only need to store one TokenDetails
+            self.tokens[token_symbol] = get_token_details(token_address, rpc)
+
+    def initialize_volumes(self) -> None:
+        for token_symbol in self.tokens:
+            self.stats.volume[token_symbol] = 0
+            self.stats.expired_requests_volume[token_symbol] = 0
 
 
 claim_request_extension = 86400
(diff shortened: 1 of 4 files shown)


beamer/health/check.py
```diff
@@ -398,8 +470,8 @@ def analyze_transfers(transfers: TransferMap, ctx: Context) -> None:
 Total fills in network: {ctx.stats.fills}
 Total claims in network: {ctx.stats.claims}
 Total withdrawals in network: {ctx.stats.withdrawals}
-Total total volume in network: {ctx.stats.volume * 1e-6} USDC
-Total expired requests volume in network: {ctx.stats.expired_requests_volume * 1e-6} USDC
+Total total volume in network: {volume_per_token}
```
bilbeyt, line 473: Why this is not just Total volume but Total total volume :D

GabrielBuragev: probably a typo, will fix it :D

beamer/health/check.py
```diff
@@ -192,27 +210,54 @@ def fetch_events() -> ChainEventMap:
...
+    token = None
+    transfer_value = "0"
+
+    for token_symbol, deployments in token_deployments.items():
+        if deployments.count([str(token_source_chain), token_source_address]):
+            token = token_details[token_symbol]
+            break
+
+    if token is not None:
+        decimal_value = get_token_amount_in_decimals(request.amount, token)
```
istankovic, line 239: When will `token` be `None` here?

beamer/health/check.py
```diff
@@ -192,27 +210,54 @@ def fetch_events() -> ChainEventMap:
...
 
-def make_web3(rpc: str) -> Web3:
-    web3 = Web3(Web3.HTTPProvider(rpc, request_kwargs=dict(timeout=5)))
-    web3.middleware_onion.inject(geth_poa_middleware, layer=0)
-    return web3
+def get_transfer_token_symbol(transfer: Transfer, token_deployments: TokenMap) -> str | None:
+    for token_symbol, deployments in token_deployments.items():
+        if any(
+            transfer["created"].source_token_address.lower() == address.lower()
+            and transfer["created"].chain_id == int(chain_id)
```
istankovic, line 217: You can move both `transfer["created"].source_token_address.lower()` and ` transfer["created"].chain_id` out of the loop and into local variables since they are invariant with respect to the loop and thereby save unnecessary operations. As a bonus, the code inside `any(...)` will be shorter and more readable.

beamer/health/check.py
```diff
@@ -389,6 +441,26 @@ def analyze_transfers(transfers: TransferMap, ctx: Context) -> None:
         else "Everything is calm in Beamerland."
     )
 
+    def format_token_volume(token_amount: int, token_details: TokenDetails) -> str:
+        return (
+            str(get_token_amount_in_decimals(token_amount, token_details))
+            + " "
+            + token_details["symbol"]
+        )
```
istankovic, line 449: I would move this out to module level. It does not depend on anything around it.

beamer/health/check.py
```diff
@@ -389,6 +441,26 @@ def analyze_transfers(transfers: TransferMap, ctx: Context) -> None:
...
+            format_token_volume(token_amount, ctx.tokens[token_symbol])
+            for token_symbol, token_amount in ctx.stats.volume.items()
+        ]
+    )
+    expired_request_volume_per_token = ", ".join(
+        [
+            format_token_volume(token_amount, ctx.tokens[token_symbol])
+            for token_symbol, token_amount in ctx.stats.expired_requests_volume.items()
+        ]
+    )
```
istankovic, line 462: Similar to above, I would use a generator expression here instead of a list comprehension.

beamer/health/check.py
```diff
@@ -192,27 +210,54 @@ def fetch_events() -> ChainEventMap:
...
+def get_transfer_value_formatted(
+    request: RequestCreated, token_deployments: TokenMap, token_details: TokenDetailsMap
+) -> str:
+    token_source_chain = request.chain_id
+    token_source_address = request.source_token_address
+    token = None
+    transfer_value = "0"
+
+    for token_symbol, deployments in token_deployments.items():
+        if deployments.count([str(token_source_chain), token_source_address]):
```
istankovic, line 234: Why do we construct a list as the argument to `count`?

istankovic:
Could you please post the relevant coverage before and after this PR?

beamer/health/check.py
GabrielBuragev: This will happen when we are trying to get the transfer value for transfers of tokens that are not yet defined inside `health-check-mainnet.conf` under the token definitions.

GabrielBuragev:
> Could you please post the relevant coverage before and after this PR?

image

beamer/health/check.py
```diff
@@ -389,6 +441,26 @@ def analyze_transfers(transfers: TransferMap, ctx: Context) -> None:
...
+            + " "
+            + token_details["symbol"]
+        )
+
+    volume_per_token = ", ".join(
+        [
+            format_token_volume(token_amount, ctx.tokens[token_symbol])
+            for token_symbol, token_amount in ctx.stats.volume.items()
+        ]
+    )
```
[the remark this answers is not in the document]
GabrielBuragev: So, if i understood it right, after applying the changes like so:
```
expired_request_volume_per_token = ", ".join(
      format_token_volume(token_amount, ctx.tokens[token_symbol])
      for token_symbol, token_amount in ctx.stats.expired_requests_volume.items()
  )
```
we are basically passing a generator to the join function which will internally exhaust it to the point where it concatenates the strings with the delimiter ", " in between?
The difference is that instead of constructing the array from the generator we are passing the generator as argument?

Thanks for the awesome summary of the differences between the two and how they can be applied in different scenarios!

beamer/health/check.py
istankovic: Ah, I see. Then I suggest to pull the list construction out of the loop, since it is loop invariant so you create the list only once. And then you could use the more Python `x in sequence` expression like
```python
    token = [str(token_source_chain), token_source_address]
    for token_symbol, deployments in token_deployments.items():
        if token in deployments:
            ...
```

beamer/health/check.py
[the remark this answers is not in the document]
istankovic: > > So, if i understood it right, after applying the changes like so:
> 
> ```
> expired_request_volume_per_token = ", ".join(
>       format_token_volume(token_amount, ctx.tokens[token_symbol])
>       for token_symbol, token_amount in ctx.stats.expired_requests_volume.items()
>   )
> ```
> 
> we are basically passing a generator to the join function which will internally exhaust it to the point where it concatenates the strings with the delimiter ", " in between? The difference is that instead of constructing the array from the generator we are passing the generator as argument?

Yes, exactly. Pulling values out of the generator will be driven by `join`, and the generator itself will then ask for the next item of `ctx.stats.expired_requests_volume.items()`, transform it according to the expression you wrote and return it back to `join`.

istankovic (approved):
A nitpick here and there, but otherwise looks good, quite readable!

GabrielBuragev:
Thanks for the awesome input @istankovic !

PR #1584, opened 2023-03-17
Add automatic withdrawal inside ArbitrumL1Messenger

### LINKED ISSUE #1314: contracts: Add automatic withdraw inside ArbitrumL1Messenger
# Rationale
In order to avoid confusion & forgotten funds inside our Messenger contracts by our agents, we should add support for automatic withdrawing of the deposited funds at the time of a `sendMessage` execution.
Currently, we expect the agents to have a decent amount of funds deposited inside the ArbitrumL1Messenger in order to relay an L1 resolution message. 

After discussing this with @fredo we agreed that by invoking the `withdraw` function inside the `sendMessage` function we will be able to reduce the costs for this operation from `2x 0.00021 ETH (gas limit) + opcode fees` to `0.00021 ETH + opcode fees`.

# Acceptance criteria
- [ ] Contracts return the deposited funds right after propagating the message that is being executed on L1
## Additional information
- [ ] Manual withdraw will be kept the way it is

GabrielBuragev:
closes #1314

### DIFF
--- a/contracts/contracts/chains/arbitrum/ArbitrumMessengers.sol
+++ b/contracts/contracts/chains/arbitrum/ArbitrumMessengers.sol
@@ -38,14 +38,18 @@ contract ArbitrumL1Messenger is IMessenger, RestrictedCalls {
         deposits[msg.sender] += msg.value;
     }
 
-    function withdraw() external {
-        uint256 amount = deposits[msg.sender];
+    function _withdraw(address target) private {
+        uint256 amount = deposits[target];
         require(amount > 0, "nothing to withdraw");
 
-        deposits[msg.sender] = 0;
+        deposits[target] = 0;
 
-        (bool sent, ) = msg.sender.call{value: amount}("");
+        (bool sent, ) = target.call{value: amount}("");
         require(sent, "failed to send Ether");
+    }
+
+    function withdraw() public {
+        _withdraw(msg.sender);
     }
 
     /* solhint-disable avoid-tx-origin */
@@ -75,6 +79,8 @@ contract ArbitrumL1Messenger is IMessenger, RestrictedCalls {
             0,
             message
         );
+
+        if (deposits[tx.origin] > 0) _withdraw(tx.origin);
     }
     /* solhint-enable avoid-tx-origin */
 }
--- a/relayer/src/services/arbitrum.ts
+++ b/relayer/src/services/arbitrum.ts
@@ -51,9 +51,9 @@ export class ArbitrumRelayerService extends BaseRelayerService {
 
     const deltaSubmissionFee = submissionFee.mul(110).div(100); // We add 10% on top to ensure that the creation of a retryable ticket doesn't fail
 
-    // We want to make sure that the deposited amount is always in the range [deltaSubmissionFee(MAX_MESSAGE_LENGTH_BYTES), 2 * delateSubmissionFee(MAX_MESSAGE_LENGTH_BYTES)]
-    // in order to avoid depositing often
     if (currentDeposit.lte(deltaSubmissionFee)) {
+      // ArbitrumL1Messenger refunds the rest of the deposit automatically
+      // after relaying a message by calling `ArbitrumL1Messenger.sendMessage`.
       const estimatedGasLimit = await arbitrumL1Messenger.estimateGas.deposit({
         value: deltaSubmissionFee,
       });
@@ -67,9 +67,14 @@ export class ArbitrumRelayerService extends BaseRelayerService {
 
       const receipt = await arbitrumL1Messenger.provider.waitForTransaction(transaction.hash, 1);
 
-      console.log(
-        `Successfully deposited funds on ArbitrumL1Messenger: ${receipt.transactionHash}`,
-      );
+      if (receipt && receipt.status) {
+        console.log(
+          `Successfully deposited funds on ArbitrumL1Messenger: ${receipt.transactionHash}`,
+        );
+        return true;
+      } else {
+        throw new Error("Arbitrum L1 deposit transaction reverted on chain.");
+      }
     } else {
       console.log("ArbitrumL1Messenger has enough funds to proceed!");
     }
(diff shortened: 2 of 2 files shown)


istankovic (dismissed):

GabrielBuragev:
@istankovic the `e2e-test-arbitrum` CI job is failing with timeouts and weird errors like: `nothing to withdraw`. 
Any idea what could be the reason for this?

Tried debugging but i couldn't figure out how to connect to the local node and investigate further. Maybe you can take a look?

fredo:
> @istankovic the `e2e-test-arbitrum` CI job is failing with timeouts and weird errors like: `nothing to withdraw`. Any idea what could be the reason for this?
> 
> Tried debugging but i couldn't figure out how to connect to the local node and investigate further. Maybe you can take a look?

I had a quick look at the code (without really looking into the node's state). I saw that you unconditionally call `withdraw()` inside of `sendMessage()`. I don't know if this is the cause for the current failure but if `depositOrigin == submissionFee` which is possible, then the new deposit would be 0.
In this case `withdraw()` would revert but the deposit would actually have been high enough. So in any case we should check for this condition before calling `withdraw()`.

contracts/contracts/chains/arbitrum/ArbitrumMessengers.sol
```diff
@@ -75,6 +75,8 @@ contract ArbitrumL1Messenger is IMessenger, RestrictedCalls {
             0,
             message
         );
+
+        this.withdraw();
```
fredo, line 79: is there a reason why you chose to make an external function call to withdraw?
Why not changing the visibility of withdraw to `public`?

see https://docs.soliditylang.org/en/v0.4.21/contracts.html#visibility-and-getters

fredo:
Found a typo in https://github.com/beamer-bridge/beamer/blob/cf6c805c126991f85df9cadb0fc7eebab8878fed/relayer/src/services/arbitrum.ts#L120

contracts/contracts/chains/arbitrum/ArbitrumMessengers.sol
GabrielBuragev: It was just lack of knowledge tbh. Will change it to a public one

GabrielBuragev:
> > @istankovic the `e2e-test-arbitrum` CI job is failing with timeouts and weird errors like: `nothing to withdraw`. Any idea what could be the reason for this?
> > Tried debugging but i couldn't figure out how to connect to the local node and investigate further. Maybe you can take a look?
> 
> I had a quick look at the code (without really looking into the node's state). It fails when estimating the l1 resolution transaction.
> 
> I saw that `withdraw()` is called unconditionally inside of `sendMessage()`. I don't know if this is the cause for the current failure but if `depositOrigin == submissionFee` which is possible and valid, then the new deposit would be 0. In this case `withdraw()` would revert but the deposit would actually have been high enough. So in any case we should check for this condition before calling `withdraw()`.

Good point. Will adjust the code

GabrielBuragev:
I found out why its reverting with `nothing to withdraw`.

Since we are calling the `withdraw` function from within `sendMessage` and `sendMessage` is being called by the resolver, we end up with the resolvers address as `msg.sender` address.
Inside the `withdraw` function, we use `msg.sender` to get the deposit amount that needs to be returned. 

Is it safe to use `tx.origin` instead of `msg.sender` inside the `withdraw` function?

fredo:
> Is it safe to use `tx.origin` instead of `msg.sender` inside the `withdraw` function?

First of all I saw that we rely on `tx.origin` anyway, to know from whom to take the fees. So in this regard we are not changing any security risk, but let me get back to you regarding the general usage of `tx.origin`.

GabrielBuragev:
I just extended the `withdraw` function to support passing an address as an argument instead of using `tx.origin` internally. Check the code and let me know what you think. Thanks!

contracts/contracts/chains/arbitrum/ArbitrumMessengers.sol
```diff
@@ -38,13 +38,13 @@ contract ArbitrumL1Messenger is IMessenger, RestrictedCalls {
         deposits[msg.sender] += msg.value;
     }
 
-    function withdraw() external {
-        uint256 amount = deposits[msg.sender];
+    function withdraw(address target) public {
```
fredo, line 41: passing the address for whom to withdraw is problematic as it is a potential DoS attack vector.
Whenever someone wants or needs to do an L1 resolution, an attacker could frontrun and withdraw his deposit first. The L1 resolution would fail.
A potential prevention could be an atomic execution of deposit and l1 resolution via a proxy (Not saying that this is the best solution, but I think the only one to withdraw on behalf here).

fredo: Alternatively, we could use tx.origin to authorize like (`require target == target.origin,"not authorized"`) but this is exactly what is known as a potential vulnerability in the contracts, I will outline in a separate comment.

fredo:
The general problem with `tx.origin` which has occured several times, is when doing authorization with `tx.origin`.

If one has a `require(tx.origin == owner)` in his contract this is considered a vulnerability. The reason for that is that a malicious contract could trick the owner to call his contract and then the malicious contract could call the vulnerable function internally which grants the malicious contract authorization.

That's why it is considered, that it's best practices to not use `tx.origin` (especially for authorization).

We have at least not such a "big problem" with it as we already have a trusted call chain in place, so malicious contracts need to sneak in at the entry point of the trusted call chain (which would be triggering L1 resolution). 

Actually what we can already do is, having a malicious contract which calls L1 resolution. Trying to trick an agent which has some deposit left on the contract, to call the malicious contract. Unnoticed the agent would call the malicious contract and actually pays for the L1 resolution. 
With the automatic withdraw, we keep the balance sheet clean, which is a nice prevention, but the vulnerability remains. So actually I already consider this a vulnerability in the current version of `ArbitrumL1Messenger`. So let's find a solution to also remove this

fredo (approved):

PR #1587, opened 2023-03-17
fix: ape: use _send_call to not isolate calls

bilbeyt:


beamer/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+from ape.api.providers import Web3Provider
+
+
+Web3Provider.send_call = Web3Provider._send_call  # type: ignore
```
istankovic, line 4: We should add a comment in here as to why we're doing this. E.g.
```
# Make ape do the call directly, instead of using chain_manager.isolate() to
# do a snapshot, make a transaction and then revert to the original chain state.
```

beamer/tests/contracts/test_fill_manager.py
```diff
@@ -152,3 +152,10 @@ def test_invalidate_previous_block_hash(fill_manager):
 
     # We proceeded one block so invalidating the fill id should work now
     fill_manager.invalidateFill(b"1", current_block.hash, chain_id)
+
+
+def test_call_without_new_block(fill_manager):
+    current_block_number = ape.chain.blocks[-1].number
+    fill_manager.l1Resolver()
+    after_call_block_number = ape.chain.blocks[-1].number
+    assert current_block_number == after_call_block_number
```
istankovic, line 161: Is this test always going to detect state reverts caused by ape?

istankovic (approved):

PR #1596, opened 2023-03-22
agent: Configurable poll period

### LINKED ISSUE #1495: agent: allow poll period to be configured per-chain
We should allows LPs to configure poll period per chain in the agent's configuration file.
Example:
```
[chains.foo]
poll-period = 1.37
```
This would set the poll period for chain `foo` to `1.37` seconds. Note that the value is a floating point number.
Poll period should also be configurable on the global level. The value to be used should be determined by the first successful lookup, in order:

- chain-specific value in the configuration file
- global value in the configuration file
- built-in default

Tasks:
- [x] introduce the new config variable
- [x] update `beamer/data/agent.conf.example` to show off the usage
- [x] document the new config variable
- [x] add a test

manuelwedler:
Closes #1495 

This allows to configure the poll period globally and per-chain via the config file. I additionally added the option to configure the global value via the cli.


beamer/agent/config.py
```diff
@@ -110,9 +113,15 @@ def load(config_path: Path, options: dict[str, Any]) -> Config:
         raise ConfigError(f"missing settings: {missing}")
 
     rpc_urls = {}
+    poll_period_per_chain = {}
 
     for chain_name, chain_info in config["chains"].items():
-        rpc_urls[chain_name] = URL(chain_info["rpc-url"])
+        rpc_url = chain_info.get("rpc-url")
+        if rpc_url is not None:
+            rpc_urls[chain_name] = URL(rpc_url)
```
istankovic, line 121: I would omit the check for existence of `rpc_url` here, let us make it crash here, rather than later in the process, when the agent tries to start.

This will anyway be addressed later by proper config validity checks.

docs/source/configuration.rst
```diff
@@ -197,6 +199,23 @@ Options reference
         [chains.foo]
         rpc-url = "http://foo.bar:8545"
 
+   * - ::
+
+        poll-period = TIME
+
+     - Time in seconds which is waited before new events are fetched from the chains after 
+       the last fetch. If a value for a specific chain is provided, it takes precedence for
+       this chain. Default: ``5.0``.
```
istankovic, line 208: Nitpick: a bit more specific variant that also highlights this as a global value could look like

```
     - Time in seconds to wait between two consecutive RPC requests for new events.
       The value applies to all chains that don't have the chain-specific poll period defined.
       Default: ``5.0``.
```

docs/source/configuration.rst
```diff
@@ -197,6 +199,23 @@ Options reference
...
+
+   * - ::
+
+        [chains.NAME]
+        poll-period = TIME
+
+     - Time in seconds which is waited before new events are fetched from the chain NAME 
+       after the last fetch. If this value is provided for the chain NAME, it takes 
+       precedence over the global poll period.
```
istankovic, line 217: Nitpick: similar to above:

```
     - Time in seconds to wait between two consecutive RPC requests for new events.
       The value applies only to chain NAME, taking precedence over the global poll period.
```

beamer/agent/agent.py
```diff
@@ -83,11 +83,14 @@ def _init_chains(self) -> dict[ChainId, _Chain]:
             contracts = make_contracts(w3, contracts_info)
             request_manager = contracts["RequestManager"]
             fill_manager = contracts["FillManager"]
+            poll_period = self._config.poll_period_per_chain.get(chain_name)
+            if poll_period is None:
+                poll_period = self._config.poll_period
```
istankovic, line 88: Could be shorter:
```python
            poll_period = self._config.poll_period_per_chain.get(chain_name, self._config.poll_period)
```

istankovic:
@manuelwedler could you please get rid of the fixup commits? They are blocked by CI.
You can just squash them into appropriate existing commits.

manuelwedler:
I usually push the fixup commits to github to let the reviewer have another look, what seems to me easier if you can just see what changed by the commits. Then, I squash them once I get an approval. Seems like we don't follow the same process in the team?

istankovic:
> I usually push the fixup commits to github to let the reviewer have another look, what seems to me easier if you can just see what changed by the commits. Then, I squash them once I get an approval. Seems like we don't follow the same process in the team?

Ah, I remember now that you do it slightly differently. I'd be ok with it, but the thing is, I like for the CI to get through because we don't know whether the changes will work or not, and that is not possible if the whole PR is blocked by fixup commits. There might be other failures which might get in the way, which may require additional changes, which should also then be covered by the review.

istankovic (approved):
Looks good!

manuelwedler:
Aren't the other pipelines already running even if the fixup commit one fails?

GabrielBuragev:
I am almost certain that the fixup commit check currently blocks the circle CI pipeline.

PR #1602, opened 2023-03-23
Update to solidity 0.8.19

istankovic:
This PR
- adds mapping key and value names, where appropriate
- updates the minimum support solidity version 0.8.19


fredo (approved):
Looks good 👍

Issue #1570, opened 2023-03-13
be able to handle reorgs


andfletcher:
# Rationale
Ethereum has reorgs regularly. In contrast to rollups this means that state can be reverted and changed. The agents have take this into account. Imagine that an agent will fill a request of a block which will be replaced, then the nonce and thus the request id could completely change because of reordering of requests.

Since the merge we haven't seen reorgs greater than 1 block. In general the agent should only act on state which he feels confident on. 

# Solution for now
We will need to introduce an confirmation block value into the event fetcher. Similar to the poll period it can be exposed in the config in the future.

The eventfetcher will only accept the latest block number and fetch events for 

`web3.eth.blockNumber - confirmations`

andfletcher:
@fredo please elaborate a bit more.

PR #1607, opened 2023-03-27
Handle reorgs for agent

bilbeyt:
Closes: #1570

### DIFF
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -51,6 +51,7 @@ class EventMonitor:
         on_new_events: list[_NewEventsCallback],
         on_sync_done: list[_SyncDoneCallback],
         poll_period: float,
+        confirmation_blocks: int,
     ):
         self._web3 = web3
         self._chain_id = ChainId(self._web3.eth.chain_id)
@@ -60,6 +61,7 @@ class EventMonitor:
         self._on_new_events = on_new_events
         self._on_sync_done = on_sync_done
         self._poll_period = poll_period
+        self._confirmation_blocks = confirmation_blocks
         self._log = structlog.get_logger(type(self).__name__).bind(chain_id=self._chain_id)
 
         for contract in contracts:
@@ -84,7 +86,9 @@ class EventMonitor:
             "EventMonitor started",
             addresses=[c.address for c in self._contracts],
         )
-        fetcher = EventFetcher(self._web3, self._contracts, self._deployment_block)
+        fetcher = EventFetcher(
+            self._web3, self._contracts, self._deployment_block, self._confirmation_blocks
+        )
         current_block = self._web3.eth.block_number
         events = []
         while fetcher.synced_block < current_block:
--- a/beamer/agent/config.py
+++ b/beamer/agent/config.py
@@ -25,6 +25,7 @@ class Config:
     token_checker: TokenChecker
     fill_wait_time: int
     unsafe_fill_time: int
+    confirmation_blocks: dict[str, int]
     prometheus_metrics_port: Optional[int]
     log_level: str
     poll_period: float
@@ -87,10 +88,12 @@ def _default_config() -> dict:
         "metrics": {},
         "tokens": {},
         "poll-period": 5.0,
+        "confirmation-blocks": 0,
     }
 
 
 _REQUIRED_KEYS = (
+    "confirmation-blocks",
     "deployment-dir",
     "fill-wait-time",
     "unsafe-fill-time",
@@ -115,12 +118,16 @@ def load(config_path: Path, options: dict[str, Any]) -> Config:
 
     rpc_urls = {}
     poll_period_per_chain = {}
+    confirmation_blocks = {}
 
     for chain_name, chain_info in config["chains"].items():
         rpc_urls[chain_name] = URL(chain_info["rpc-url"])
         poll_period = chain_info.get("poll-period")
         if poll_period is not None:
             poll_period_per_chain[chain_name] = float(poll_period)
+        confirmation_blocks[chain_name] = chain_info.get(
+            "confirmation-blocks", config["confirmation-blocks"]
+        )
 
     path = Path(_get_value(config, "account.path"))
     password = _get_value(config, "account.password")
@@ -135,6 +142,7 @@ def load(config_path: Path, options: dict[str, Any]) -> Config:
         rpc_urls=rpc_urls,
         token_checker=token_checker,
         base_chain_rpc_url=config["base-chain"]["rpc-url"],
+        confirmation_blocks=confirmation_blocks,
         fill_wait_time=config["fill-wait-time"],
         unsafe_fill_time=config["unsafe-fill-time"],
         prometheus_metrics_port=_lookup_value(config, "metrics.prometheus-port"),
--- a/beamer/agent/data/agent.conf.example
+++ b/beamer/agent/data/agent.conf.example
@@ -3,6 +3,7 @@ deployment-dir = "deployments/goerli"
 fill-wait-time = 120
 unsafe-fill-time = 600
 poll-period = 5.0
+confirmation-blocks = 0
 
 [account]
 path = "account.json"
@@ -17,6 +18,7 @@ poll-period = 60.0
 
 [chains.goerli-arbitrum]
 rpc-url = "GOERLI_ARBITRUM_RPC_URL"
+confirmation-blocks = 1
 
 [chains.goerli-optimism]
 rpc-url = "GOERLI_OPTIMISM_RPC_URL"
--- a/beamer/agent/events.py
+++ b/beamer/agent/events.py
@@ -225,13 +225,20 @@ class EventFetcher:
     _ETH_GET_LOGS_THRESHOLD_FAST = 2
     _ETH_GET_LOGS_THRESHOLD_SLOW = 5
 
-    def __init__(self, web3: Web3, contracts: tuple[Contract, ...], start_block: BlockNumber):
+    def __init__(
+        self,
+        web3: Web3,
+        contracts: tuple[Contract, ...],
+        start_block: BlockNumber,
+        confirmation_blocks: int,
+    ):
         self._web3 = web3
         self._chain_id = ChainId(web3.eth.chain_id)
         self._contract_addresses = [c.address for c in contracts]
         self._next_block_number = start_block
         self._blocks_to_fetch = EventFetcher._DEFAULT_BLOCKS
         self._event_abis = _make_topics_abi_mapping_for_contracts(contracts)
+        self._confirmation_blocks = confirmation_blocks
         self._log = structlog.get_logger(type(self).__name__).bind(chain_id=self._chain_id)
 
         for contract in contracts:
@@ -302,15 +309,17 @@ class EventFetcher:
     def fetch(self) -> list[Event]:
         try:
             block_data = self._web3.eth.get_block("latest")
-            block_number = BlockNumber(block_data["number"])
         except RequestException:
             return []
+
+        block_number = BlockNumber(block_data["number"] - self._confirmation_blocks)
 
         if block_number < self._next_block_number:
             return []
 
         result = []
         from_block = self._next_block_number
+
         while from_block <= block_number:
             to_block = min(block_number, BlockNumber(from_block + self._blocks_to_fetch))
             try:
--- a/beamer/tests/agent/test_request.py
+++ b/beamer/tests/agent/test_request.py
@@ -377,3 +377,39 @@ def test_event_fetcher_progress_without_events(agent, direction):
                 != ape.chain.blocks[-1].number
             ):
                 sleeper.sleep(0.1)
+
+
+def test_handling_reorgs(config, direction, token, request_manager):
+    requester, target = alloc_accounts(2)
+    confirmation_blocks = 2
+    config.confirmation_blocks["l2a"] = confirmation_blocks
+    agent = Agent(config)
+    agent.start()
+
+    snapshot = ape.chain.snapshot()
+    ape.chain.mine(confirmation_blocks + 1)
+    request_id_1 = make_request(request_manager, token, requester, target, 1)
+
+    assert ape.chain.blocks[-1].number is not None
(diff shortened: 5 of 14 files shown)


beamer/agent/chain.py
```diff
@@ -26,6 +26,8 @@
 
 POLL_PERIOD: float = 5
 
+CONFIRMATION_BLOCKS: int = 1
```
istankovic, line 29: Could you please make this configurable, per-chain, similar to what Manuel has done for the poll period?
We should have one global default which is used in case the chain-specific value is not defined.

beamer/agent/events.py
```diff
@@ -307,10 +314,19 @@ def fetch(self) -> list[Event]:
...
+        if block_number - self._confirmation_blocks == 0:
+            return []
+
+        if block_number - self._confirmation_blocks < from_block:
+            return []
+
+        while from_block <= block_number - self._confirmation_blocks:
+            to_block = (
+                min(block_number, from_block + self._blocks_to_fetch) - self._confirmation_blocks
+            )
```
istankovic, line 327: I feel like the entire block above could be simplified by just subtracting the number of confirmation blocks from `block_number` after getting the block number from the RPC, i.e. after that `try...except` block above. Then you would not need to change anything here, right?

beamer/agent/config.py
```diff
@@ -110,9 +113,13 @@ def load(config_path: Path, options: dict[str, Any]) -> Config:
         raise ConfigError(f"missing settings: {missing}")
 
     rpc_urls = {}
+    chain_confirmation_blocks = {}
 
     for chain_name, chain_info in config["chains"].items():
         rpc_urls[chain_name] = URL(chain_info["rpc-url"])
+        chain_confirmation_blocks[chain_name] = int(
+            chain_info.get("confirmation-blocks", config["confirmation-blocks"])
+        )
```
istankovic, line 122: I'm guessing this `int(...)` here is only because of mypy... if so, :woozy_face:

beamer/agent/events.py
```diff
@@ -307,10 +314,11 @@ def fetch(self) -> list[Event]:
 
         result = []
         from_block = self._next_block_number
+
         while from_block <= block_number:
             to_block = min(block_number, BlockNumber(from_block + self._blocks_to_fetch))
             try:
-                events = self._fetch_range(from_block, to_block)
+                events = self._fetch_range(from_block, BlockNumber(to_block))
```
istankovic, line 321: Why do we need to cast `to_block` to `BlockNumber`? It was without it before...

beamer/agent/events.py
```diff
@@ -298,7 +305,7 @@ def _fetch_range(
     def fetch(self) -> list[Event]:
         try:
             block_data = self._web3.eth.get_block("latest")
-            block_number = BlockNumber(block_data["number"])
+            block_number = BlockNumber(block_data["number"] - self._confirmation_blocks)
```
istankovic, line 308: Nitpick: this is totally fine, but I'd prefer to have an `assert block_number > 0` after this anyway...
Also, only the `get_block` call should be inside of this `try...except` block, the rest does not need to be there so let's move it out of the block.

beamer/agent/config.py
```diff
@@ -24,6 +24,7 @@ class Config:
     token_checker: TokenChecker
     fill_wait_time: int
     unsafe_fill_time: int
+    chain_confirmation_blocks: dict[str, int]
```
istankovic, line 27: Nitpick: just `confirmation_blocks` would be shorter and a bit nicer....

beamer/tests/agent/test_request.py
```diff
@@ -378,3 +378,20 @@ def test_event_fetcher_progress_without_events(agent, direction):
...
+    agent.start()
+
+    request_id = make_request(request_manager, token, requester, target, 1)
+    ape.chain.mine(confirmation_blocks)
+
+    with Sleeper(1) as sleeper:
+        while (agent.get_context(direction).requests.get(request_id)) is None:
+            sleeper.sleep(0.1)
+
+    agent.stop()
```
istankovic, line 397: How does this test work, i.e. how does a reorg happen and how do we check that we handled it correctly?

bilbeyt: so basically without extra ape.chain.mine(1) there we dont get the request at all, it is working in this way.

bilbeyt: so if request in block number 23 normally we can only reach it in block number 24 iterationn which we want

beamer/agent/events.py
bilbeyt: my bad :D forgot it

docs/source/commands.rst
```diff
@@ -32,6 +32,10 @@ The ``agent`` command will run a Beamer agent and provide liquidity for the brid
      - Time in seconds to wait for a fill event before challenging a false claim.
        Default: ``120``.
 
+   * - ``--confirmation-blocks BLOCKS``
+     - Number of confirmation blocks to consider the block is ready to process.
```
istankovic, line 36: Nitpick: `block is ready to process` -> `block ready for processing`

docs/source/configuration.rst
```diff
@@ -150,6 +152,13 @@ Options reference
      - Time in seconds to wait for a fill event before challenging a false claim.
        Default: ``120``.
 
+   * - ::
+
+        confirmation-blocks = BLOCKS
+
+     - Number of confirmation blocks to consider the block is ready to process.
```
istankovic, line 159: Nitpick: `block is ready to process` -> `block ready for processing`

beamer/tests/agent/test_request.py
istankovic: I see, but the problem is, this test would be successful even without the confirmation blocks feature, right?. I.e. we are not handling a reorg _really_ because there is no reorg. We are just testing whether we wait for the specified number of confirmation blocks.

bilbeyt: isnt it enough? can you do a reorg in test?

istankovic: Well, one way to do it would be to 
1. make a snapshot
2. mine N blocks
3. roll back the chain to the snapshot
4. mine new blocks

I think that's as close as we could get to a real reorg with ape.
Now, for the actual test...

1. make a snapshot
2. mine N blocks, where `N > confirmation blocks`
3. make a request in another block, note the request ID, say `request1`
4. roll back the chain to the snapshot
5. mine N blocks
6. assert that the agent _did not_ see `request1`
7. make another request, note the request ID, say `request2`
8. mine `confirmation blocks` new blocks
9. assert that the agent _did_ see `request2`

istankovic (changes requested):
Looks good overall,  but I think we need to improve the test.

beamer/agent/data/agent.conf.example
```diff
@@ -2,6 +2,7 @@ log-level = "debug"
 deployment-dir = "deployments/goerli"
 fill-wait-time = 120
 unsafe-fill-time = 600
+confirmation-blocks = 0
```
fredo, line 5: What is the meaning of the confirmation-blocks on the top level?

beamer/agent/config.py
```diff
@@ -110,9 +113,13 @@ def load(config_path: Path, options: dict[str, Any]) -> Config:
         raise ConfigError(f"missing settings: {missing}")
 
     rpc_urls = {}
+    chain_confirmation_blocks = {}
 
     for chain_name, chain_info in config["chains"].items():
         rpc_urls[chain_name] = URL(chain_info["rpc-url"])
+        chain_confirmation_blocks[chain_name] = int(
```
fredo, line 120: general suggestion: Wouldn't it be nicer to have a "chain" dictionary in the config which stores chain related information. 
Since we sort our information per chain we could have a dataclass `ChainConfig`
and a `dict[chain_name, ChainConfig]` (kind of similar how we structured it with api schema in the deployment config

beamer/agent/data/agent.conf.example
fredo: is it meant to be a default?

beamer/agent/events.py
bilbeyt: not sure about assert here actually, lets say confirmation blocks 1 and block number is 1 then this makes 0 and assert will crash, i would have it as if and return [] instead so this wont crash

bilbeyt: not adding any assert would be doing it actually

beamer/agent/data/agent.conf.example
bilbeyt: yes it will be default. the first suggestion was using the structure like we are using for poll_period and chain based poll_period so i have added also this one as default.

beamer/agent/config.py
bilbeyt: i can handle this after merging this pr

istankovic: > and a `dict[chain_name, ChainConfig]` (kind of similar how we structured it with api schema in the deployment config

Exactly this has been my plan to do, just didn't want to block this PR on that. But yes, using the apischema approach for agent config is definitely the way to go.

beamer/agent/events.py
istankovic: > not adding any assert would be doing it actually

I'm afraid I don't understand this. :sweat_smile:

bilbeyt: so block_number < 0 will return [] in next if at line 317 because self._next_block_number would be bigger

istankovic: > so block_number < 0 will return [] in next if at line 317 because self._next_block_number would be bigger

Ah, yes, perfect! :+1:

beamer/tests/agent/test_request.py
```diff
@@ -377,3 +377,30 @@ def test_event_fetcher_progress_without_events(agent, direction):
...
+
+
+def test_handling_reorgs(config, direction, token, request_manager):
+    requester, target = alloc_accounts(2)
+    confirmation_blocks = 2
+    config.confirmation_blocks["l2a"] = confirmation_blocks
+    agent = Agent(config)
+    agent.start()
+
+    snap1 = ape.chain.snapshot()
```
istankovic, line 389: Nitpick: I would just call this `snapshot` instead of `snap1`.

beamer/tests/agent/test_request.py
```diff
@@ -377,3 +377,30 @@ def test_event_fetcher_progress_without_events(agent, direction):
...
+    ape.chain.mine(confirmation_blocks + 1)
+    request_id_1 = make_request(request_manager, token, requester, target, 1)
+    ape.chain.restore(snap1)
```
istankovic, line 392: Here it could happen that the agent does not get a chance to sync before we restore the chain state.
So before doing `restore`, we should wait until event fetcher's `synced_block` is `latest chain's block number - confirmation block`. Only then we should do a `restore`.

beamer/agent/config.py
fredo: ok then lets go for it after this PR, can you create a follow up please?

beamer/tests/agent/test_request.py
```diff
@@ -377,3 +377,39 @@ def test_event_fetcher_progress_without_events(agent, direction):
...
+    confirmation_blocks = 2
+    config.confirmation_blocks["l2a"] = confirmation_blocks
+    agent = Agent(config)
+    agent.start()
+
+    snapshot = ape.chain.snapshot()
+    ape.chain.mine(confirmation_blocks + 1)
+    request_id_1 = make_request(request_manager, token, requester, target, 1)
+
+    assert ape.chain.blocks[-1].number
```
istankovic, line 393: We're checking whether the block number is zero? Is that intentional?

beamer/agent/data/agent.conf.example
fredo: Just to clarify: 
The agent has the option to configure the global default, plus for chain specific values, which override the default? 
Additionally, we have an internal default which is used if no global default is set? 
And this counts both for poll period and confirmation blocks?

beamer/tests/agent/test_request.py
```diff
@@ -377,3 +377,39 @@ def test_event_fetcher_progress_without_events(agent, direction):
...
+    expected_block_number = ape.chain.blocks[-1].number - confirmation_blocks
+    with Sleeper(1) as sleeper:
+        while (
+            agent.get_context(direction).latest_blocks[ChainId(ape.chain.chain_id)].number
+            != expected_block_number
+        ):
+            sleeper.sleep(0.1)
```
istankovic, line 400: Very nice solution, that is exactly what we needed!

beamer/agent/data/agent.conf.example
fredo: (if true, sounds to me like an overkill)

beamer/tests/agent/test_request.py
bilbeyt: it is mypy again and the reason is actually this expression return type it is int | None so i am making sure that this is not None

beamer/agent/data/agent.conf.example
istankovic: > Just to clarify: The agent has the option to configure the global default, plus for chain specific values, which override the default? Additionally, we have an internal default which is used if no global default is set? And this counts both for poll period and confirmation blocks?

Yes, exactly.

I do admit it may seem as a bit of an overkill and most likely not needed (at least not right now), but OTOH, having it in this manner is more in line with how people generally expect configuration to behave, so I see it not as "we're definitely not going to need this" but rather "people expect config files to work a certain way, so let's make it intuitive for them, even though they may not have an immediate need for it".

beamer/tests/agent/test_request.py
istankovic: > it is mypy again and the reason is actually this expression return type it is int | None so i am making sure that this is not None

Ah, mypy again. :woozy_face: 
But then, please make it explicit that you are testing against `None`:

```python
    assert ape.chain.blocks[-1].number is not None
```

istankovic (approved):
Looks good from my side.

@fredo may have some reservations about this, though.

PR #1611, opened 2023-03-28
Change agent config to use base-chain instead of l1

bilbeyt:


beamer/tests/conftest.py
```diff
@@ -178,8 +178,8 @@ def config(request_manager, fill_manager, token, token_list):
     account = eth_account.Account.from_key(_LOCAL_ACCOUNT.private_key)
     token.mint(account.address, 300)
     url = URL(ape.config.provider.uri)
-    rpc_urls = {"l1": URL(url), "l2a": URL(url), "l2b": URL(url)}
-    poll_period_per_chain = {"l1": 2.0}
+    poll_period_per_chain = {"l2a": 1.1}
```
istankovic, line 181: Why are we making L2 poll period bigger than the one for L1?

docs/source/configuration.rst
```diff
@@ -188,6 +194,13 @@ Options reference
      - Name of the target chain. Deprecated and will be removed.
        No longer needed because the agent supports multiple chain pairs.
 
+   * - ::
+
+        [base-chain]
+        rpc-url = URL
+
+     - Associate a JSON-RPC endpoint URL with base chain. May be given multiple times.
```
istankovic, line 202: Why `May be given multiple times.`?

beamer/tests/conftest.py
bilbeyt: there is no l1 eventmonitor anyway ? so i was thinking we missed this in previous pr imo. that is why i have now l2a and previously tests were not failing as i explained but with 2 secs normally timeouts happening inn tests so i still wanted to have difference so changed it to become 1.1

docs/source/configuration.rst
bilbeyt: my bad :D

scripts/deployment/main.py
```diff
@@ -264,8 +264,8 @@ def _margin_gas_price_strategy(web3: Web3, transaction_params: TxParams) -> Wei:
     resolver = deploy_contract(web3_l1, "Resolver")
 
     deployment_data: dict = {"beamer_commit": get_commit_id(), "deployer": account.address}
-    deployment_data["L1"] = collect_contracts_info({"Resolver": resolver})
-    deployment_data["L2"] = {}
+    deployment_data["base_chain"] = collect_contracts_info({"Resolver": resolver})
```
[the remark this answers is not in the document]
bilbeyt: may have some problems in relayer/fe side but lets see

beamer/tests/conftest.py
istankovic: > there is no l1 eventmonitor anyway ?

What do you mean?
This conftest and the config here is common to all tests, since it is at the top of the pytest hierarchy, no?
Wherever we have an agent, we will have event monitors which would be affected by this, assuming they don't customise the poll period.

> so i was thinking we missed this in previous pr imo.

Which PR?

> that is why i have now l2a and previously tests were not failing as i explained but with 2 secs normally timeouts happening inn tests

Could you please expand on this part, please? 

Also, if we really need to specify a poll period, why not go into the other direction and use, say, `0.9` or `0.8`?

bilbeyt: https://github.com/beamer-bridge/beamer/pull/1596 this one. so the thing here if you look at the agent.py. if chain_name is l1 then it continues and actually there is no event monitor right? so that is why adding poll period 2.0 here in conftest for the l1 had no impact in tests at all, so there was no timeout error in agent tests

bilbeyt: i am ok with 0.9 or 0.8 no problem

istankovic: > #1596 this one. so the thing here if you look at the agent.py. if chain_name is l1 then it continues and actually there is no event monitor right? so that is why adding poll period 2.0 here in conftest for the l1 had no impact in tests at all, so there was no timeout error in agent tests

Ah, I see now, yes, having `l1` in there totally does not make sense, good catch!

istankovic: Actually, here I think we can even configure the global poll period to `0.5` and just remove `poll_period_per_chain`. What do you think?

bilbeyt: only for tests right? it is ok by me

scripts/deployment/main.py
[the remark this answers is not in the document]
GabrielBuragev: With `base-chain` as a property name (instead of `base_chain`) we would have to access it in the TS/JS code as `deployment['base-chain']` instead of `deployment.base_chain`.
That will be the only thing affected by this change.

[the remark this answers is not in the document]
GabrielBuragev: We rarely access properties with `['']` in TS/JS but if we want to be consistent with naming (we should) then i am fine with this small exception.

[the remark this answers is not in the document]
GabrielBuragev: Why don't we just use `base_chain` as a property name inside the deployment config files ??
If we want to be consistent with property name casing than we should apply this change.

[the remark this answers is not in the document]
bilbeyt: this was the same thinking when i first answer it like this :D

beamer/tests/conftest.py
istankovic: > only for tests right? it is ok by me

Yes, only for tests. Hopefully it makes them a bit faster too.

scripts/deployment/main.py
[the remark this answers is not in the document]
istankovic: Yeah, ok, if it makes the fronted part easier, let's leave this as-is and after this lands we can convert the deployment config files to be in sync with this.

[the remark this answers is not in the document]
GabrielBuragev: So its not forgotten: https://github.com/beamer-bridge/beamer/issues/1619

Issue #1371, opened 2023-01-17
Frontend: Optimize/reduce claim count listeners


GabrielBuragev:
# Rationale
Currently, when a request has been marked as expired, we subscribe to the `ClaimMade` && `ClaimStakeWithdrawn` events by using a single explicit topic like so:
```
contract.filters.ClaimStakeWithdrawn(undefined, options.requestIdentifier);
contract.filters.ClaimMade(options.requestIdentifier);
```

This means that for every expired request we will have new event listeners.
But this can be improved with `ethers >= 5` like so:
```
const reduceFilter = contract.filters.ClaimStakeWithdrawn(undefined, [requestId1, requestId2]);
const increaseFilter = contract.filters.ClaimMade([requestId1, requestId2]);
```

Why should this feature exist?
Optimizes our app & reduces the RPC calls

# Acceptance Criteria
- [ ] Listeners for `ClaimMade` & `ClaimStakeWithdrawn` events are grouped by using an OR topic-set

PR #1612, opened 2023-03-28
Frontend: add reorg safety mechanisms

### LINKED ISSUE #1572: Frontend: Handle reorgs
Same as https://github.com/beamer-bridge/beamer/issues/1570 but this one focuses on the frontend.

# Acceptance Criteria
- [ ] Frontend waits until transactions are confirmed. 
This means that there is at least +1 blocks mined after the request transaction or an reported event by ethers when listening with a filter.

GabrielBuragev:
Closes #1572


frontend/tests/unit/services/transactions/request-manager.spec.ts
```diff
@@ -42,6 +42,14 @@ const PROVIDER = new JsonRpcProvider();
 const SIGNER = new JsonRpcSigner(undefined, PROVIDER);
 const DEFAULT_REQUEST_IDENTIFIER = '1';
 
+function mockGetSafeEventHandler() {
```
manuelwedler, line 45: Can't we put such functions to a shared utility file, so we don't have to define them several times?

frontend/tests/unit/services/transactions/utils.spec.ts
```diff
@@ -2,18 +2,88 @@ import * as providers from '@ethersproject/providers';
...
 } from '@/services/transactions/utils';
 import { getRandomEthereumAddress, getRandomUrl } from '~/utils/data_generators';
+import { MockedEvent, MockedTransactionReceipt } from '~/utils/mocks/ethers';
 
 vi.mock('@ethersproject/providers');
 vi.mock('ethers');
 
 describe('utils', () => {
+  describe('getSafeEventHandler()', () => {
+    it('returns a re-org safe version of the passed handler', () => {
```
manuelwedler, line 21: Nitpick: I think this test name is a bit far-fetched. It just ensures that a function is returned.

frontend/tests/unit/services/transactions/utils.spec.ts
```diff
@@ -2,18 +2,88 @@ import * as providers from '@ethersproject/providers';
...
+        CONFIRMATION_TIME_BLOCKS[chainId] = confirmationTimeBlocks;
+
+        const provider = new providers.JsonRpcProvider();
+        provider.getNetwork = vi.fn().mockResolvedValue({ chainId });
+        provider.waitForTransaction = vi
+          .fn()
+          .mockResolvedValue(new MockedTransactionReceipt({ status: 1 }));
+
+        const safeHandler = getSafeEventHandler(handler, provider);
+        await safeHandler(event);
```
manuelwedler, line 46: The handler isn't async. The whole test function doesn't have to be.

frontend/tests/unit/services/transactions/utils.spec.ts
```diff
@@ -2,18 +2,88 @@ import * as providers from '@ethersproject/providers';
...
+
+          const provider = new providers.JsonRpcProvider();
+          provider.getNetwork = vi.fn().mockResolvedValue({ chainId });
+          provider.waitForTransaction = vi
+            .fn()
+            .mockResolvedValue(new MockedTransactionReceipt({ status: 1 }));
+
+          const safeHandler = getSafeEventHandler(handler, provider);
+
+          await safeHandler(event);
```
manuelwedler, line 75: Again not needed `await`

manuelwedler (changes requested):
I think it's perfect how you did it. I'm not sure if it is really necessary to have another abstraction around ether's provider. Actually, we already have our own abstraction, the `IEthereumProvider` interface. It was once the idea to just use that in all the transaction related functions. What was missing was to have also our own interface for the signer, which could be used in the transaction functions. But later these rpc urls got introduced as parameters and we went a step back by coupling more with the ethers library again. Overall, it seems a bit much to me to have another abstraction level for the provider.

So the changes look good. Some things just need some clarification.

frontend/src/services/transactions/utils.ts
```diff
@@ -1,10 +1,46 @@
...
+import type { ContractInterface, Event } from 'ethers';
 import { Contract } from 'ethers';
 
 import type { EthereumAddress } from '@/types/data';
 
+type ConfirmationTimeMap = {
+  [chainId: string]: number;
+};
+
+const DEFAULT_CONFIRIMATION_TIME_BLOCKS = 1;
```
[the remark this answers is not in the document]
GabrielBuragev: From the official ethersjs documentation on what number of confirmations blocks means:
>The number of blocks that have been mined (including the initial block) since this transaction was mined.

So, 1 would mean waiting until the block where the transaction is included in has been mined. 

I would keep it to 1 here so we make sure we wait until the transaction has been included in a mined block.

frontend/src/services/transactions/utils.ts
```diff
@@ -1,10 +1,46 @@
...
+      // See why here: https://ethereum.stackexchange.com/a/10739
+      return;
+    }
+
+    const receipt = await provider.waitForTransaction(
+      event.transactionHash,
+      confirmationTimeBlocks,
+    );
+    if (receipt.status) {
+      handler();
```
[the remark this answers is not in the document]
GabrielBuragev: Thats a very good finding. The token balance real-time update was never properly working due to that.
Just tested it and it fails with: `token is undefined`, as expected.

Will try to fix this.

frontend/tests/unit/services/transactions/utils.spec.ts
GabrielBuragev: Actually, the handler isn't the one we are awaiting for but instead the wrapped function around the handler which is async. 
If i remove the await tests will fail.

frontend/tests/unit/services/transactions/utils.spec.ts
GabrielBuragev: True ✅

GabrielBuragev:
About this
> I'm not sure if it is really necessary to have another abstraction around ether's provider.

I am perfectly fine with using our current abstraction but i would like to see it being used inside these service modules instead of passing around RPC urls and creating new providers each time. That was my whole point. 
That way we could optimize many things like: grouping event listeners https://github.com/beamer-bridge/beamer/issues/1371, configuring safe confirmation block ranges based on connected chain and keeping these values in the chain config files, eliminating the inefficient, unnecessary multi-polling due to the many JsonRpcProvider instances etc.

manuelwedler:
> I am perfectly fine with using our current abstraction but i would like to see it being used inside these service modules instead of passing around RPC urls and creating new providers each time. That was my whole point. That way we could optimize many things like: grouping event listeners #1371, configuring safe confirmation block ranges based on connected chain and keeping these values in the chain config files, eliminating the inefficient, unnecessary multi-polling due to the many JsonRpcProvider instances etc.

Totally agree! As I said, it was rather intended to pass instances of the `IEthereumProvider` interface around instead of the rpc urls.

frontend/src/services/transactions/utils.ts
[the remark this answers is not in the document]
manuelwedler: Ok, then it's perfect how it is!

frontend/src/services/transactions/utils.ts
```diff
@@ -1,10 +1,46 @@
...
 
 import type { EthereumAddress } from '@/types/data';
 
+type ConfirmationTimeMap = {
+  [chainId: string]: number;
+};
+
+const DEFAULT_CONFIRIMATION_TIME_BLOCKS = 1;
+export const CONFIRMATION_TIME_BLOCKS: ConfirmationTimeMap = {
+  1: 2,
```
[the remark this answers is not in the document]
manuelwedler: Alright perfect then!

frontend/tests/unit/services/transactions/utils.spec.ts
manuelwedler: woops, sorry was wrong about this. My IDE marked it as not awaitable. But the reason is just that the return type of the `getSafeEventHandler` is not accurate. The `Listener` type does not return a Promise. The return type should instead be something like: `(...args: Array<unknown>) => Promise<void>`

frontend/tests/unit/services/transactions/utils.spec.ts
manuelwedler: see above

frontend/tests/unit/services/transactions/utils.spec.ts
```diff
@@ -2,18 +2,88 @@ import * as providers from '@ethersproject/providers';
...
+          expect(handler).toHaveBeenCalled();
+
+          handler.mockReset();
+
+          await safeHandler(reorgedEvent);
```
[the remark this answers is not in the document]
manuelwedler: see above

manuelwedler (approved):
Looks perfect now. Thanks! Only the return type of the getSafeEventHandler function should still be updated. Since @GabrielBuragev is on vacation, I will do it myself and squash the fixup commits.

PR #1614, opened 2023-03-29
Add fees to RequestCreated event

bilbeyt:


compojoom (approved):
Looks good to me. That will help with figuring out what agents have earned with our new fees :)

PR #1618, opened 2023-03-29
frontend: Handle MetaMask Mobile error in switchChain of MetaMaskProvider

### LINKED ISSUE #1617: frontend: MetaMask Mobile not possible to add chains
#### System information

MetaMask Mobile Android

#### Expected behavior
When trying to select a source chain that is not added in the user's wallet, the rpc should triggered to add the chain.

#### Actual behavior
Nothing happens. So the source chain cannot be changed.

Select a source chain that is not in the user's wallet.

manuelwedler:
Closes #1617 

MetaMask Mobile wraps the error code differently. Actually it is a bug: https://github.com/MetaMask/metamask-mobile/issues/3312 . We can still handle the code because it is also wrapped inside the error object. This is necessary because without it, the `wallet_addEthereumChain` rpc was not triggered in case the wallet does not know the chain yet.

Additionally, I fixed some of our tests which didn't await .rejects assertions.

### DIFF
--- a/frontend/src/services/web3-provider/metamask-provider.ts
+++ b/frontend/src/services/web3-provider/metamask-provider.ts
@@ -67,7 +67,7 @@ export class MetaMaskProvider extends InjectedProvider implements ISigner {
       await this.web3Provider.send('wallet_switchEthereumChain', [{ chainId: newChainIdHex }]);
       return true;
     } catch (error) {
-      if ((error as { code: number })?.code === unrecognizedChainErrorCode) {
+      if (hasErrorCode(error as MetaMaskError, unrecognizedChainErrorCode)) {
         return false;
       }
       throw error;
@@ -77,4 +77,10 @@ export class MetaMaskProvider extends InjectedProvider implements ISigner {
   listenToEvents(): void {
     this.listenToChangeEvents();
   }
+}
+
+type MetaMaskError = { code?: number; data?: { originalError?: { code: number } } };
+
+function hasErrorCode(error: MetaMaskError, code: number) {
+  return error.code === code || error.data?.originalError?.code === code;
 }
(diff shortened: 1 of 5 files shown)


frontend/src/services/web3-provider/metamask-provider.ts
```diff
@@ -67,7 +67,12 @@ export class MetaMaskProvider extends InjectedProvider implements ISigner {
       await this.web3Provider.send('wallet_switchEthereumChain', [{ chainId: newChainIdHex }]);
       return true;
     } catch (error) {
-      if ((error as { code: number })?.code === unrecognizedChainErrorCode) {
+      type MetaMaskError = { code: number };
+      type MetaMaskMobileError = { data?: { originalError?: { code: number } } };
+      if (
+        (error as MetaMaskError)?.code === unrecognizedChainErrorCode ||
+        (error as MetaMaskMobileError)?.data?.originalError?.code === unrecognizedChainErrorCode
+      ) {
```
GabrielBuragev, line 75: Can we separate this checks into two functions? Currently its really hard to read and understand whats going on here.

The function can be like:
```
type MetaMaskError = { code: number };
function isMetaMaskError(error: any, code: number): error is MetaMaskError {
  const mmError = error as MetaMaskError;
  return !!mmError.code && mmError.code === code;
}
```

GabrielBuragev (approved):
One suggestion to improve readability. Rest looks fine, thanks!

frontend/src/services/web3-provider/metamask-provider.ts
manuelwedler: I think an `isMetaMaskError` function should not check for a specific code to be present. I added a `hasErrorCode`. Please have look at the fixup commit.

GabrielBuragev: Yeah, that looks even better! Thanks

PR #1630, opened 2023-03-30
beamer: agent: improve camel to snake case conversion

istankovic:
So that we can have target_weight_ppm instead of target_weight_p_p_m.


fredo (approved):
let's go with this for now, it works. We can look for improvements later on

PR #1634, opened 2023-03-30
scripts: Change property name casing in deployment

compojoom:
Property case has been changed to snake case.

courtesy of chatGPT

### DIFF
--- a/scripts/deployment/config.py
+++ b/scripts/deployment/config.py
@@ -2,7 +2,7 @@ import json
 from dataclasses import dataclass, field
 from pathlib import Path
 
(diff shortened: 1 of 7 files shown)


scripts/deployment/config.py
```diff
@@ -67,7 +67,7 @@ def __str__(self) -> str:
 
 @dataclass
 class Config:
-    base_chain: _BaseChain = field(metadata=alias("base-chain"))
+    base_chain: _BaseChain = field(metadata=schema(min_items=1))
```
fredo, line 70:
```suggestion
    base_chain: _BaseChain
```

istankovic (approved):

PR #1642, opened 2023-03-31
Makefile: extend the clean target

istankovic:
Clean up contract build artifacts, the dist dir and various caches.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,3 +43,9 @@ docs:
 
 clean:
 	make -C docs clean
+	rm -rf dist
+	rm -rf contracts/.build
+	rm -rf contracts/contracts/.cache
+	rm -rf .pytest_cache
+	find . -type d -name __pycache__ -exec rm -rf {} +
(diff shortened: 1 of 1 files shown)


fredo (approved):

fredo:
should we include / consult frontend for this target?

istankovic:
> should we include / consult frontend for this target?

Definitely. At the minimum I noticed `node_modules/.cache`, which was empty for me, but there could be more.
@compojoom @GabrielBuragev ?

istankovic:
> There are also more potential files in the repo and I wonder where we draw the line here. I.e. if we clean `.pytest_cache`, should we also clean `.mypy_cache`?

Umm, I'm already cleaning `.mypy_cache`, please see the diff.

IMO, we should at least clean stuff that is product of tools that are our dev-dependencies: `ape`, `mypy`, `solc`....

fredo:
how did I overlook? sorry, my bad

PR #1648, opened 2023-04-03
Update to latest web3, 6.0.0

bilbeyt:
This PR moves us to web3 6.0.0.

There were quite a few changes, but I would point out a few relevant to us:
- the removal of support for `dict`-based simple cache middlewares (this PR derives from the newly-added `SimpleCache` and uses `lru.LRU` to keep the same caching semantics that we have had so far)
- the support for accessing struct fields via names (requires `decode_tuples=True` when constructing a contract, which is what this PR does; this allows us to remove index-based field accesses)
-  snake_casing in many methods/functions
- strict byte checking for function arguments (more details in web3 docs)

Apart from these, there were tonnes of async-related changes, which are not relevant to us since we do not use async components of web3.

Closes: #1650


fredo (approved):
Looks good ❤️

fredo (approved):

beamer/agent/events.py
```diff
@@ -9,7 +9,8 @@
 from hexbytes import HexBytes
 from requests.exceptions import HTTPError, ReadTimeout, RequestException
 from web3 import HTTPProvider, Web3
-from web3.contract import Contract, get_event_data
+from web3._utils.events import get_event_data
```
istankovic, line 12: Isn't there another place where we could import `get_event_data` from?
In general, we should not use private things (modules, items) like that.

beamer/agent/util.py
```diff
@@ -120,9 +120,9 @@ def make_web3(
     w3.middleware_onion.inject(geth_poa_middleware, layer=0)
     w3.middleware_onion.add(construct_sign_and_send_raw_middleware(account))
 
-    # Cache data of 1000 least recently used blocks.
-    cache_class = cast(Type[dict[Any, Any]], functools.partial(lru.LRU, 1000))
-    middleware = construct_simple_cache_middleware(cache_class=cache_class)
+    # Cache data of 1000 entry.
```
istankovic, line 123: Typo: `entry` -> `entries`

istankovic: Can you explain how `SimpleCache` works?

fredo: I looked at the code. It is a simple OrderedDict with a FIFO queue. So unfortunately not LRU.

bilbeyt: corrected the typo here.

beamer/tests/contracts/test_l1_resolution.py
```diff
@@ -70,15 +70,15 @@ def test_invalidation_before_and_after_resolution(contracts, request_manager):
...
+    assert request_manager.requests(request_id).filler == ADDRESS_ZERO
 
     request_manager.resolveRequest(
         request_id, FILL_ID, chain_id, address, sender=contracts.l1_messenger
     )
 
     # Resolution validates fill again
     assert not request_manager.isInvalidFill(request_id, FILL_ID)
-    assert request_manager.requests(request_id)[RM_R_FIELD_FILLER] == address
+    assert request_manager.requests(request_id).filler == address
```
fredo, line 81: I was awaiting this such a long time  ❤️

fredo:
Looks awesome! It wasn't a complete renovation after all but nice improvements have been made.

fredo (approved):

PR #1655, opened 2023-04-03
ci: e2e optimism requires create-relayers job

fredo:
Is there a reason why this was missing?

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -369,6 +369,7 @@ workflows:
       - e2e-test-optimism:
           requires:
             - compile-contracts
+            - create-relayers
       - e2e-test-arbitrum:
           requires:
             - compile-contracts


istankovic:
Hm, interesting, why does that job need the `create-relayers` job?

fredo:
> Hm, interesting, why does that job need the `create-relayers` job?

How are messages relayed in e2e tests? Ah wait, didn't have the Optimism docker setup its own relay? That would explain it.

fredo:
Ok I just verified that optimism is the only e2e test that does not use our relayer for l1 resolution. However, it would be good if we did so... 
This effectively means we are not testing our optimism relayer at all in a system test?

istankovic:
> Ok I just verified that optimism is the only e2e test that does not use our relayer for l1 resolution. However, it would be good if we did so... This effectively means we are not testing our optimism relayer at all in a system test?

Yeah, that's pretty much always been the case. :sweat_smile:

istankovic (approved):
Not really useful until we get #1242, but okay.

PR #1666, opened 2023-04-05
Rename chain_id to event_chain_id

bilbeyt:
As ChainUpdated event in RequestManager has chain_id, this conflicts with Event.chain_id in our agent event representation and chain_id field in ChainUpdatedEvent is rewritten. In order to prevent this, we need to rename chain_id in our agent representation to something else like event_chain_id so it wont get overwritten.

### DIFF
--- a/beamer/agent/events.py
+++ b/beamer/agent/events.py
@@ -26,7 +26,7 @@ from beamer.agent.typing import (
 
 @dataclass(frozen=True)
 class Event:
-    chain_id: ChainId
+    event_chain_id: ChainId
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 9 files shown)


istankovic (approved):
Looks good.

One note regarding the PR and commit messages in general: for anyone outside of our discussions, there is no way to tell why this change was made. Ideally, the information about _why_ a change is being made should be present in the commit message(s) so that anyone can see it and understand the motivation behind the changes.

I am also often guilty of writing just "what the commit does", but not "why do we do this".
For simple changes, I realize it may just be annoying, but if there is a more interesting story to tell, we should tell it in the commit message(s).

Another way to look at this:
- the commit diff shows _what the change is_
- the commit message says _why the change is being made_

fredo (approved):

Issue #1657, opened 2023-04-04
Use type 2 transactions for L1 with own gas price strategy


fredo:
The agent is currently using type 0 transactions. Type 0 transactions are pre EIP-1559 transaction types which define `gasPrice`. This leads to the problem that a validator can use up to the full amount of `gasPrice - baseFee` as a tip. 
So we either overprice transactions very much to be sure to be included in time or we risk of not being included due to rising baseFees. 

We can highly optimize this with type 2 (post EIP-1559) transactions. In type 2 transactions we define a `maxFeePerGas` and a `maxPriorityFeePerGas` where `maxFeePerGas` describes the maximum price we are willing to pay. `maxPriorityFeePerGas` describes the maximum tip we are willing to pay for the validator.

So in practice we don't need to "guess" the perfect gas price where we definitely cover the base fee (otherwise the tx cannot be included in the block) and don't overpay the validator.
With type 2 transactions we only pay the base fee + the max priority fee what we have set, even though we would have been willing to pay more if the base fee had been higher.

Update: 

I have found out that web3.py actually uses dynamic fees (post 1559) per default. Only if there is no `maxFeePerGas` or `maxPriorityFeePerGas` and a gas price strategy set or `gasPrice` set in the tx params then static fees are applied. We use a gas price strategy and that's why we use type 0 transactions.

But Removing that is not enough. We need to manually adapt the maxFeePerGas parameter. This can be done via a middleware, but whatever is returned by the RPC is often not enough. I haven't seen `maxFeeStrategies` analog to what we have for the old gasPrice key.
We have have to write our own middleware for this.

[code in web3.py where dynamic or static fees are selected](https://github.com/ethereum/web3.py/blob/3c00810e9a3563c69901549a6a54d4fe19c9d6a9/web3/_utils/transactions.py#L100)

[default values for dynamic fee parameters](https://github.com/ethereum/web3.py/blob/3c00810e9a3563c69901549a6a54d4fe19c9d6a9/web3/_utils/transactions.py#L68)

istankovic:
Thanks for writing this issue, this is quite important!

PR #1671, opened 2023-04-05
Support type 2 transactions

bilbeyt:
Closes: #1657

### DIFF
--- a/beamer/agent/middleware.py
+++ b/beamer/agent/middleware.py
@@ -10,9 +10,40 @@ import lru
 import requests.exceptions
 import structlog
 from web3 import HTTPProvider, Web3
-from web3.types import RPCEndpoint, RPCResponse
+from web3.types import Middleware, RPCEndpoint, RPCResponse
+from beamer.agent.typing import ChainId
+
 
 log = structlog.get_logger(__name__)
+
+CacheMiddleware = Callable[
+    [Callable[[RPCEndpoint, Any], RPCResponse], Web3, "_BlockCache"],
+    Callable[[RPCEndpoint, Any], RPCResponse],
+]
+
+
+class _BlockCache:
+    def __init__(self) -> None:
+        self._block_cache = lru.LRU(1000)
+        self._latest_block_number = -1
+        self._latest_key: tuple[str, bool] | None = None
+        self._lock = threading.Lock()
+
+    def add_block(self, key: tuple[str, bool], data: RPCResponse) -> None:
+        with self._lock:
+            self._block_cache[key] = data
+            if self._latest_block_number < data["result"].number:
+                self._latest_key = key
+                self._latest_block_number = data["result"].number
+
+    def get_block(self, key: tuple[str, bool]) -> RPCResponse:
+        return self._block_cache.get(key)
+
+    def get_latest_block(self) -> RPCResponse | None:
+        return self._block_cache.get(self._latest_key)
+
+
+_BLOCK_STORAGE: dict[ChainId, _BlockCache] = {}
 
 
 def _result_ok(response: RPCResponse) -> bool:
@@ -21,16 +52,24 @@ def _result_ok(response: RPCResponse) -> bool:
     return response.get("result") is not None
 
 
+def generate_middleware_with_cache(middleware: CacheMiddleware, chain_id: ChainId) -> Middleware:
+    global _BLOCK_STORAGE
+
+    if chain_id not in _BLOCK_STORAGE:
+        _BLOCK_STORAGE[chain_id] = _BlockCache()
+
+    cache = _BLOCK_STORAGE[chain_id]
+    return cast(Middleware, functools.partial(middleware, cache=cache))
+
+
 # This middleware caches responses of eth_getBlockByNumber calls. Requests with
 # block number set to 'latest' will never be returned from the cache. However,
 # successful responses of those calls will be stored into the cache so that
 # further requests specifying a concrete block number can be satisfied from
 # the cache.
 def cache_get_block_by_number(
-    make_request: Callable[[RPCEndpoint, Any], RPCResponse], _w3: Web3
+    make_request: Callable[[RPCEndpoint, Any], RPCResponse], _w3: Web3, cache: _BlockCache
 ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
-    cache = lru.LRU(1000)
-
     def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
         if method != "eth_getBlockByNumber":
             return make_request(method, params)
@@ -39,13 +78,13 @@ def cache_get_block_by_number(
             response = make_request(method, params)
             if _result_ok(response):
                 key = hex(response["result"].number), params[1]
-                cache[key] = response
+                cache.add_block(key, response)
         elif params[0].startswith("0x"):
-            response = cache.get(params)
+            response = cache.get_block(params)
             if response is None:
                 response = make_request(method, params)
                 if _result_ok(response):
-                    cache[params] = response
+                    cache.add_block(params, response)
         else:
             response = make_request(method, params)
         return response
@@ -282,3 +321,31 @@ def rate_limiter(
 ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
     state = _RateLimiterState()
     return functools.partial(_rate_limiter, make_request=make_request, w3=w3, state=state)
+
+
+def max_fee_setter(
+    make_request: _MakeRequest, _w3: Web3, cache: _BlockCache
+) -> Callable[[RPCEndpoint, Any], RPCResponse]:
+    def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
+        if method != "eth_sendTransaction":
+            return make_request(method, params)
+        priority_fee_response = make_request(RPCEndpoint("eth_maxPriorityFeePerGas"), [])
+        if _result_ok(priority_fee_response):
+            priority_fee = priority_fee_response["result"]
+        else:
+            return priority_fee_response
+
+        latest_block = cache.get_latest_block()
+        if latest_block is None:
+            latest_block_response = make_request(RPCEndpoint("eth_getBlockByNumber"), ["latest"])
+            if _result_ok(latest_block_response):
+                latest_block = latest_block_response
+            else:
+                return latest_block_response
+        base_fee = latest_block["result"].baseFeePerGas
+        max_fee = 2 * base_fee + priority_fee
+        params[0]["maxPriorityFeePerGas"] = priority_fee
+        params[0]["maxFeePerGas"] = max_fee
+        return make_request(method, params)
+
+    return middleware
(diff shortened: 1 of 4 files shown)


beamer/agent/middleware.py
```diff
@@ -282,3 +282,27 @@ def rate_limiter(
 ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
     state = _RateLimiterState()
     return functools.partial(_rate_limiter, make_request=make_request, w3=w3, state=state)
+
+
+def base_fee_setter(
```
fredo, line 287: I would call this `max_fee_setter` because this is what we are effectively doing

beamer/agent/middleware.py
```diff
@@ -282,3 +282,27 @@ def rate_limiter(
...
+    make_request: _MakeRequest, _w3: Web3
+) -> Callable[[RPCEndpoint, Any], RPCResponse]:
+
+    current_fees = make_request(RPCEndpoint("eth_feeHistory"), ["0x1", "latest", [10, 100]])[
```
fredo, line 291: question: Wouldn't everything outside of the inner middleware function only be called once? 
If so we have to move the fee fetching part into the inner middleware function

beamer/agent/middleware.py
```diff
@@ -282,3 +282,27 @@ def rate_limiter(
...
+        "result"
+    ]
+    base_fee = max(current_fees["baseFeePerGas"])
+    max_fee = 2 * base_fee
+    priority_fee = make_request(RPCEndpoint("eth_maxPriorityFeePerGas"), [])["result"]
```
istankovic, line 296: Please be aware that `make_request` may fail so we need to check whether it was successful.

fredo:
General question: Have you had a look at https://github.com/ethereum/web3.py/blob/3c00810e9a3563c69901549a6a54d4fe19c9d6a9/web3/_utils/transactions.py#L68 ? 
It seems that web3.py's default is effectively doing what we are currently writing in the middleware. Of course, the middleware would give the power into our hands, but we could also save the effort and use the defaults.

Another option which would bring an actual benefit is that we fetch the latest block headers in the event fetcher anyway. There we can read the latest base fee so we could actually save us an RPC request. However, this is a bit tricky because it's stored in our logic and we would have to actively pass it to web3.

A third option could be to leverage the cache `getBlockByNumber` and always set the latest Base fee whenever we get a new block. If those caches were somehow able to communicate or cache the fee history call somehow then we could map it onto our cache system.

beamer/agent/middleware.py
bilbeyt: i mean when i print out the values max fee is definetely changing and priority fee is just constant in ganache.

beamer/agent/middleware.py
```diff
@@ -282,3 +301,34 @@ def rate_limiter(
 ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
     state = _RateLimiterState()
     return functools.partial(_rate_limiter, make_request=make_request, w3=w3, state=state)
+
+
+def max_fee_setter(
+    make_request: _MakeRequest, _w3: Web3, chain_id: ChainId
+) -> Callable[[RPCEndpoint, Any], RPCResponse]:
+    global _LATEST_BLOCK_NUMBER_MAP
+    global _BLOCK_CACHE
```
istankovic, line 310: Do we actually need these `global` declarations here?

beamer/agent/middleware.py
```diff
@@ -21,15 +25,29 @@ def _result_ok(response: RPCResponse) -> bool:
...
 # further requests specifying a concrete block number can be satisfied from
 # the cache.
-def cache_get_block_by_number(
-    make_request: Callable[[RPCEndpoint, Any], RPCResponse], _w3: Web3
+def _cache_get_block_by_number(
+    make_request: Callable[[RPCEndpoint, Any], RPCResponse], _w3: Web3, chain_id: ChainId
 ) -> Callable[[RPCEndpoint, Any], RPCResponse]:
-    cache = lru.LRU(1000)
+    global _BLOCK_CACHE
+    global _LATEST_BLOCK_NUMBER_MAP
```
istankovic, line 50: We probably don't need these `global` declarations here,. right?

beamer/agent/middleware.py
```diff
@@ -39,13 +57,14 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
             response = make_request(method, params)
             if _result_ok(response):
                 key = hex(response["result"].number), params[1]
-                cache[key] = response
+                _BLOCK_CACHE[chain_id][key] = response
+                _LATEST_BLOCK_NUMBER_MAP[chain_id] = hex(response["result"].number)
```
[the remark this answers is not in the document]
istankovic: You're calling `hex(response["result"].number)` twice, once for `key`, once here. We should avoid such duplication of work.

beamer/agent/middleware.py
```diff
@@ -21,15 +25,29 @@ def _result_ok(response: RPCResponse) -> bool:
...
+    if chain_id not in _LATEST_BLOCK_NUMBER_MAP:
+        _LATEST_BLOCK_NUMBER_MAP[chain_id] = None
+    return _cache_get_block_by_number(make_request, _w3, chain_id)
+
+
 # This middleware caches responses of eth_getBlockByNumber calls. Requests with
 # block number set to 'latest' will never be returned from the cache. However,
 # successful responses of those calls will be stored into the cache so that
 # further requests specifying a concrete block number can be satisfied from
 # the cache.
```
istankovic, line 45: Nitpick: I'd move this comment to above the public entry point, `cache_get_block_by_number`, the function below is an implementation detail.

beamer/agent/middleware.py
```diff
@@ -39,13 +57,14 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
             response = make_request(method, params)
             if _result_ok(response):
                 key = hex(response["result"].number), params[1]
-                cache[key] = response
+                _BLOCK_CACHE[chain_id][key] = response
```
istankovic, line 60: Why do we store the whole response, instead of just `response['result']`?

fredo: don't we have to store the whole response in order to imitate the actual call?

beamer/agent/middleware.py
```diff
@@ -11,10 +11,36 @@
...
 log = structlog.get_logger(__name__)
 
 
+class _BlockCache:
+    def __init__(self) -> None:
+        self._block_cache = lru.LRU(1000)
+        self._latest_block_number = ""
+        self._lock = threading.Lock()
+
+    def add_block(self, key: tuple[str, bool], data: RPCResponse) -> None:
```
fredo, line 26: nice ! we can implicitly setting the latest block when adding a new block

beamer/agent/middleware.py
bilbeyt: yes it is because we need to return the whole response from middleware

beamer/agent/middleware.py
```diff
@@ -27,9 +53,12 @@ def _result_ok(response: RPCResponse) -> bool:
 # further requests specifying a concrete block number can be satisfied from
 # the cache.
 def cache_get_block_by_number(
-    make_request: Callable[[RPCEndpoint, Any], RPCResponse], _w3: Web3
+    make_request: Callable[[RPCEndpoint, Any], RPCResponse], _w3: Web3, chain_id: ChainId
```
fredo, line 56: nitpick: 

with 

```py
def partial_cache(middleware, chain_id: ChainId) -> Callable[[RPCEndpoint, Any], RPCResponse]:
    if chain_id not in _BLOCK_STORAGE:
        _BLOCK_STORAGE[chain_id] = _BlockCache()

    cache = _BLOCK_STORAGE[chain_id]

    return partial(middleware, cache=cache)
```

we could convert the chain id agrument into the cache object

beamer/agent/middleware.py
```diff
@@ -282,3 +312,34 @@ def rate_limiter(
...
+) -> Callable[[RPCEndpoint, Any], RPCResponse]:
+    def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
+        if method != "eth_sendTransaction":
+            return make_request(method, params)
+        priority_fee_response = make_request(RPCEndpoint("eth_maxPriorityFeePerGas"), [])
+        if _result_ok(priority_fee_response):
+            priority_fee = priority_fee_response["result"]
+        else:
+            return priority_fee_response
+        latest_block_number = _BLOCK_STORAGE[chain_id].get_latest_block_number()
```
fredo, line 328: reading this part now, I would actually give the `_BlockCache` class a bit more power to directly get the latest block result 

like 

```py
@property
def latest_block() RPCResponse | None:
       return self.get_block(self._latest_key))
```

fredo: This way we could save some boilerplate in hte max fee setter middleware

beamer/agent/middleware.py
```diff
@@ -11,10 +11,36 @@
...
+        with self._lock:
+            self._block_cache[key] = data
+            if self._latest_block_number < data["result"].number:
+                self._latest_key = key
+                self._latest_block_number = data["result"].number
+
+    def get_block(self, key: tuple[str, bool]) -> RPCResponse:
+        return self._block_cache.get(key)
+
+    def get_latest_block(self) -> RPCResponse:
```
fredo, line 37: the return value should be RPCResponse | None

because self._latest_key can be initialized with None

beamer/agent/middleware.py
```diff
@@ -282,3 +311,31 @@ def rate_limiter(
...
+            return priority_fee_response
+
+        latest_block = _BLOCK_STORAGE[chain_id].get_latest_block()
+        if latest_block is None:
+            latest_block_response = make_request("eth_getBlockByNumber", ["latest"])
+            if _result_ok(latest_block_response):
+                latest_block = latest_block_response
+            else:
+                return latest_block_response
+        base_fee = latest_block["result"]["baseFeePerGas"]
```
fredo, line 335: nitpick: can we also use properties? `latest_block["result"].baseFeePerGas`? Would be nicer if we are consistent in the usage

beamer/agent/middleware.py
```diff
@@ -282,3 +311,31 @@ def rate_limiter(
...
+        max_fee = 2 * base_fee + priority_fee
+        params[0]["maxPriorityFeePerGas"] = priority_fee
```
fredo, line 337: are params[0] the txParams?

beamer/agent/middleware.py
```diff
@@ -39,13 +68,13 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
             response = make_request(method, params)
             if _result_ok(response):
                 key = hex(response["result"].number), params[1]
-                cache[key] = response
+                _BLOCK_STORAGE[chain_id].add_block(key, response)
         elif params[0].startswith("0x"):
-            response = cache.get(params)
+            response = _BLOCK_STORAGE[chain_id].get_block(params)
```
fredo, line 73: do you know if we have to set the alternative (param `instead`)?

bilbeyt: imo params is key in the latest part

fredo: sorry, I meant the function `lru.get(key, instead)`. I know for dicts its returning None per default at key error. Is that also the case for lru ?

bilbeyt: yes.

fredo (approved):
Great work!

Issue #1667, opened 2023-04-05
Handle pending transactions


fredo:
When a transaction is pending `web3.eth.wait_for_transaction_receipt()` will timeout eventually. If this happens, the transaction is dropped and no state change will happen (this is similar to what happens if the transaction failed). As a consequence, the agent will send another transaction as it doesnt see the pending one, and no state change has happened, neither by the blockchain nor by the agent internally.

We must have a way to handle pending transactions, as the agent will blindly send new transactions on top, which might not be bad for different transactions like a different claim but for repeating the same transaction, this is definitely a bug.

fredo:
let's go for it now. I haven't come to a better solution with respect to the time it needs

PR #1672, opened 2023-04-05
Handle pending tx

istankovic:
We now wait indefinitely for tx receipt.

One thing to note here is that the `timeout` parameter to `transact` is not so useful as it was before.

Before: setting `timeout=N` would set an upper bound on the runtime of `transact`
After: setting `timeout=N` would not have an effect on the upper bound on the runtime of `transact`

(I'm ignoring the `attempts` parameter and internal `transact` call of course)

Since we've started to use `transact` all over the place, I think we should keep this change of behaviour in mind.

Closes: #1667

### DIFF
--- a/beamer/agent/util.py
+++ b/beamer/agent/util.py
@@ -17,7 +17,7 @@ from eth_utils import is_checksum_address, to_checksum_address
 from web3 import HTTPProvider, Web3
 from web3.contract import ContractConstructor
 from web3.contract.contract import ContractFunction
-from web3.exceptions import ContractLogicError
+from web3.exceptions import ContractLogicError, TimeExhausted
 from web3.gas_strategies.rpc import rpc_gas_price_strategy
 from web3.middleware import (
     construct_sign_and_send_raw_middleware,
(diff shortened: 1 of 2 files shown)


beamer/agent/util.py
```diff
@@ -66,9 +66,16 @@ def transact(
...
-    receipt = func.w3.eth.wait_for_transaction_receipt(
-        txn_hash, timeout=timeout, poll_latency=poll_latency
-    )
+    while True:
+        try:
+            receipt = func.w3.eth.wait_for_transaction_receipt(
+                txn_hash, timeout=timeout, poll_latency=poll_latency
+            )
+        except TimeExhausted as exc:
+            log.error("timed out waiting for tx receipt, retrying", exc=exc)
```
fredo, line 75: nitpick: If it's not a name like function name we normally start logs with capital letters

istankovic: Fixed, thanks!

fredo (approved):
let's go for it now. I haven't come to a better solution with respect to the time it needs

PR #1673, opened 2023-04-06
Add a .nvmrc file

compojoom:
===
It used to annoy me a little, that I first try "yarn" and it tells me that my node version is not compatible. Then had to do: 'nvm use 18.12"...
With this change and granted one has made the nvm shell integration one can just open the beamer project and the correct 18.x version will be auto applied. 
===

A lot of people use NVM to manage node environments. We are generally enforcing node engine 18.x in the main package.json and in the relay’s package.json. For the frontend we say node > 16.

NVM has the option to specify a node version to use in an .nvmrc file. By having this file one can now just do:
```
nvm use
```

Or even have a deeper shell integration and the moment one enters the directory nvm could load the specified node version. For more info on how to set this up: https://github.com/nvm-sh/nvm#nvmrc


istankovic (approved):
Cool, didn't know this was possible!

PR #1678, opened 2023-04-06
Docs contract addresses

istankovic:
This PR adds a script (with a terribly long name, suggestions welcome!) to generate a doc page with contract addresses, given a deployment file. The script expects JSON on stdin and outputs rst  to be included in the docs.

The Makefile targets are updated and also CI config is changed so that whenever the mainnet deployment file is changed, the page will be regenerated and uploaded.

Example for the current deployment:
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,7 @@ workflows:
             scripts/.* backend-updated true
             relayer/.* relayer-updated true
             docs/.* docs-updated true
+            deployments/mainnet/deployment.json docs-updated true
             frontend/.* frontend-updated true
             subgraph/.* subgraph-updated true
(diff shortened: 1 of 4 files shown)


fredo:
How will this behave now, when we have Ethereum as the base layer and as a rollup? It looks like that we would have the Headline Ethereum twice which would look confusing.

fredo:
could you generate and post a screenshot how it looks now?

bilbeyt:
Screenshot 2023-04-12 at 14 30 58

bilbeyt:
> Screenshot 2023-04-12 at 14 30 58

@fredo

fredo (approved):
looks good

PR #1686, opened 2023-04-11
Deployment/mainnet config

fredo:
Current Mainnet and Goerli deployment for V2


scripts/deployment/goerli-testnets.json
```diff
@@ -1,13 +1,44 @@
...
+          "token_address": "mintable_token",
+          "transfer_limit": 10000,
+          "eth_in_token": 1000
+        }
+      ]
+    },
     {
       "name": "Arbitrum Goerli",
-      "rpc": "https://goerli-rollup.arbitrum.io/rpc",
+      "rpc": "",
```
GabrielBuragev, line 41: Do we really need to remove the rpc urls for the testnets?
All of them are public so i don't see a reason why we would make the job of re-deploying on Goerli harder for us (searching online and finding out which rpc urls we should insert here before deploying).

fredo: The public RPCs are not stable enough for a deployment always (risk of rate limit). I'd rather add our dev rpcs. 
On the other hand, the change in the deployment infrastructure will change the concept anyway and we have to add an rpc file.

PR #1711, opened 2023-04-14
Agent/health/custom prefix

compojoom:


beamer/health/health-check-mainnet.conf
```diff
@@ -12,6 +13,11 @@ token=""
 chat-id=""
 request-throttling-in-sec=0
 
+[chains.mainnet]
```
bilbeyt, line 16: nitpick: chains.ethereum we are also referring it like this.

beamer/health/health-check-mainnet.conf
```diff
@@ -27,11 +33,12 @@ chain-id=10
 # All tokens within the same list are considered equivalent and
 # transfers between them are allowed.
 USDC = [
+    ["1", "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"],
     ["10", "0x7F5c764cBc14f9669B88837ca1490cCa17c31607"],
-    ["42161", "0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8"]
+    ["42161", "0xff970a61a04b1ca14834a43f5de4533ebddb5cc8"]
```
bilbeyt, line 38: i think both works here right? then why to change ?

bilbeyt (approved):
small things but looks good

beamer/health/health-check-mainnet.conf
compojoom: I was copying from my local config where it was not checksummed. Thanks!

Issue #1085, opened 2022-09-28
Don't create unneeded InitiateL1Resolution events


istankovic:
We currently create multiple `InitiateL1Resolution` events instead of just one, upon receiving the first challenger's `ClaimMade` event. These events can live in the event for some time, causing multiple futures to be created needlessly. We should change the agent so that we create create only one `InitiateL1Resolution` event, upon receiving the first challenger's `ClaimMade` event.

- [ ] write a test
- [ ] port the changes to the `main` branch

fredo:
As we discussed, Initiating L1 could be done via a flag in the state instead of inducing a new internal event.
The flag would trigger L1 upon `process_request()` or `process_claims()` respectively.

PR #1715, opened 2023-04-14
feat: handle l1 resolutions safely

bilbeyt:
Closes: #1085


istankovic:
@bilbeyt what is the difference in code coverage with this PR?

bilbeyt:
> @bilbeyt what is the difference in code coverage with this PR?

so the latest test-beamer indicates 86% for main and this has 87%, i am refactoring chain.py and state_machine.py in this one and getting rid of some events maybe that is why it is increasing.

beamer/tests/agent/test_challenge.py
```diff
@@ -408,3 +409,92 @@ def test_challenge_7(request_manager, fill_manager, token, config):
 
     assert agent_earnings() == stake
     assert dave_earnings() == stake + 1
+
+
+@pytest.mark.usefixtures("setup_relayer_executable")
+def test_l1_resolution(request_manager, fill_manager, token, config, direction, chain_params):
```
fredo, line 415: why not reusing the agentfixture?

beamer/tests/agent/test_challenge.py
```diff
@@ -408,3 +409,92 @@ def test_challenge_7(request_manager, fill_manager, token, config):
...
+
+    assert claim is not None
+
+    # challenge with enough to go to l1
+    request_manager.challengeClaim(
+        claim.claimId, sender=exploiter, value=ape.convert("2 ether", int)
+    )
+
+    with Sleeper(10) as sleeper:
+        while not request.l1_resolved.is_active:
```
fredo, line 452: question: do we have a relayer running ? or how does the request get into the l1_resolved state. 
If we have a real relayer executable, do we still need the mockup file creation for the relayer in `setup_relayer_executable`?

bilbeyt: no we dont have. l1_resolved state is assigned when the l1 validation called actually, and i am just creating executable here for test that is echoing something not failing and after this step basically. request is l1_resolved

bilbeyt: i think you are right if we have the correct relayer there it might create a problem as it can fail but i dont see we are changing l1_resolved state back in the code either.

beamer/tests/agent/test_challenge.py
bilbeyt: agent is using ganache account which i cant drain really, in this test i wanted to have an agent with limited funds.

beamer/tests/agent/test_challenge.py
bilbeyt: ok also tried with real relayer now, so l1 resolution fails but request still in l1_resolved state so that is why this test passes

beamer/tests/agent/test_challenge.py
bilbeyt: as agent fixture is working in my current trial i am updating to use agent fixture here.

beamer/agent/chain.py
```diff
@@ -476,6 +482,147 @@ def maybe_invalidate(claim: Claim, context: Context) -> None:
...
+def _invalidation_ready_for_l1_relay(claim: Claim) -> bool:
+    return (
+        not claim.invalidated_l1_resolved.is_active
+        and claim.invalidation_tx is not None
+        and claim.invalidation_timestamp is not None
+    )
+
+
+def _l1_resolution_threshold_reached(claim: Claim, context: Context) -> bool:
+    l1_gas_cost = 1_000_000  # TODO: Adapt to real price
```
istankovic, line 502: How far is the actual gas cost on L1 from this value?

fredo: To extend on this a bit, costs actually depend on the two chains, which are part of the resolution. My analysis shows that `from_chain` (the contract where we get the message from), mainly determines the costs. 

Message flow example: 

Opt -> Arb 950k gas
ETH -> Arb 150k gas 

The reason for it is that getting the message from the Optimism outbox, you have to verify a merkle proof but on the EthereumL2Messenger, you just have to encode and hash some values. 

For now I don't see this as a problem as long as we choose a value higher than the highest costs, which is true.

fredo: In any case we have to rework this calculation as we also should incorporate costs for the challenges being done, and we do this check on two different places in the code base (here and [here](https://github.com/beamer-bridge/beamer/blob/d4bbfaf966e0364df45213a66c865afb1e99407d/beamer/agent/chain.py#L441))

bilbeyt: yes in the other PR i guess i changed to use only one calculation

beamer/agent/chain.py
```diff
@@ -476,6 +482,143 @@ def maybe_invalidate(claim: Claim, context: Context) -> None:
...
+        last_challenger = claim.latest_claim_made.last_challenger
+        if last_challenger == context.address:
+            reward -= (
+                claim.latest_claim_made.challenger_stake_total
+                - claim.latest_claim_made.claimer_stake
+            )
+
+        if reward > limit:
+            return True
+    return False
```
fredo, line 523: I know it#s just copy paste here but you could simplify this function by pulling out the condition check like

```py
if claim.claimer == context.address:
    reward = ...
else:
    reward = ....
return reward > limit
```

bilbeyt: ok done

beamer/agent/chain.py
```diff
@@ -476,6 +482,139 @@ def maybe_invalidate(claim: Claim, context: Context) -> None:
...
+    )
+
+    def on_future_done(f: Future) -> None:
+        try:
+            f.result()
+        except Exception as ex:
+            context.logger.error("L1 resolution failed", ex=ex)
+        finally:
+            assert request is not None, "Request object missing"
+            del context.l1_resolutions[request.id]
```
fredo, line 609: hmm we have a race condition here. The relayer process will most likely be done before the event arrived at the agent. In this scenario we would run the relayer again, mistakenly.
Now it gets a bit tricky. If the relayer waits until he sees the event, we can be fine that the event is there. In this case we could add a flag (or just don't remove the value of the dict) 
Because then we can be sure that the agent will read the event eventually. 

Only if the process fails, we have to delete the value in order to allow another process to be started

beamer/agent/chain.py
```diff
@@ -476,6 +482,139 @@ def maybe_invalidate(claim: Claim, context: Context) -> None:
...
+def _timestamp_is_l1_finalized(
+    timestamp: Timestamp, context: Context, target_chain_id: ChainId
+) -> bool:
+    # The internal memory should not be none because at this point
+    # the ChainUpdated event must have arrived, otherwise there would be no request
+    target_chain_finality = context.finality_periods[target_chain_id]
+    return int(time.time()) > timestamp + target_chain_finality
+
+
+def maybe_resolve_invalidation(claim: Claim, context: Context) -> bool:
```
fredo, line 531: question: between `maybe_resolve_invalidation` and `maybe_resolve_request` there seems to be quite some code redundancy. Can we unify the redundancy?

bilbeyt: ok

beamer/agent/chain.py
```diff
@@ -476,6 +489,117 @@ def maybe_invalidate(claim: Claim, context: Context) -> None:
...
+            f.result()
+        except Exception as ex:
+            context.logger.error(L1_RESOLUTION_MESSAGES[info.context_key]["error"], ex=ex)
+        finally:
+            context_field = getattr(context, info.context_key)
+            del context_field[info.identifier]
+
+    future.add_done_callback(on_future_done)
+    context_field = getattr(context, info.context_key)
+    context_field[info.identifier] = future
```
fredo, line 594: Suggestion:

We can unify `context.invalidations` and `context.l1_resolutions` into just `context.l1_resolutions` (which holds both request resolutions and fill invalidations) and just use the tx_hash as the key. This is unique anyway and every message can only be resolved once

beamer/agent/chain.py
```diff
@@ -476,6 +489,117 @@ def maybe_invalidate(claim: Claim, context: Context) -> None:
...
+    timestamp: Timestamp, context: Context, target_chain_id: ChainId
+) -> bool:
+    # The internal memory should not be none because at this point
+    # the ChainUpdated event must have arrived, otherwise there would be no request
+    target_chain_finality = context.finality_periods[target_chain_id]
+    return int(time.time()) > timestamp + target_chain_finality
+
+
+@dataclass
+class _ResolveInfo:
```
fredo, line 539: If we unify then we don't need this private class or the MESSAGES anymore. You can just inline in 552 to 563 and use local variables instead

beamer/agent/chain.py
```diff
@@ -476,6 +478,109 @@ def maybe_invalidate(claim: Claim, context: Context) -> None:
...
+    def on_future_done(f: Future) -> None:
+        try:
+            f.result()
+        except Exception as ex:
+            context.logger.error("L1 resolution failed", ex=ex)
+        finally:
+            del context.l1_resolutions[identifier]
+
+    future.add_done_callback(on_future_done)
+    context.l1_resolutions[identifier] = future
```
fredo, line 577: can we please use the tx hash instead of the id as the key?

beamer/agent/chain.py
```diff
@@ -476,6 +478,109 @@ def maybe_invalidate(claim: Claim, context: Context) -> None:
...
+
+def maybe_resolve(claim: Claim, context: Context) -> bool:
+    request = context.requests.get(claim.request_id)
+
+    assert request is not None, "Active claim for non-existent request"
+
+    timestamp_finalized = False
+    identifier: ClaimId | RequestId
+
+    if claim.invalidation_timestamp:
```
fredo, line 535: I think we have a bug in here. We can have invalidations and fills for the same claim. And we may want to execute both.

fredo: nevermind, we cannot have both

beamer/agent/chain.py
```diff
@@ -293,6 +292,9 @@ def process_claims(context: Context) -> None:
         if claim.claimer_winning.is_active or claim.challenger_winning.is_active:
             maybe_withdraw(claim, context)
             maybe_challenge(claim, context)
+            if _invalidation_ready_for_l1_relay(claim) or _proof_ready_for_l1_relay(request):
```
fredo, line 295: nitpick quesiton: Why did you decide to pull this condition check out of the funciton? If you look at all the `maybe_` functions there is typically a list of condition checks, it would be more consistent to keep it inside.

fredo: (Exception is the l1_cost check in the other PR, because one excludes the other maybe function.

beamer/agent/state_machine.py
```diff
@@ -375,6 +321,8 @@ def _handle_request_resolved(event: RequestResolved, context: Context) -> Handle
         try:
             request.l1_resolve(event.filler, event.fill_id)
             request.invalid_fill_ids.pop(event.fill_id, None)
+            if request.fill_tx:
```
fredo, line 324: I would prefer the explicit check `request.fill_tx is not None`

beamer/agent/state_machine.py
```diff
@@ -407,136 +355,8 @@ def _handle_fill_invalidated_resolved(
...
-                del context.l1_invalidations[claim.id]
-            except Exception as ex:
-                context.logger.error("L1 invalidation failed", ex=ex)
-
-        future.add_done_callback(on_future_done)
-        context.l1_invalidations[claim.id] = future
-        claim.l1_invalidate()
-
-        context.logger.info("Initiated L1 invalidation", request=request, claim=claim)
+        if claim.invalidation_tx:
```
fredo, line 358: same here

PR #1721, opened 2023-04-17
Makefile: improve the dist-exe target

istankovic:
First, make sure that the dist directory exists (it could be missing as a result of running the clean target, for example).

Second, fail if poetry detects an inconsistency between pyproject.toml and poetry.lock. This is to avoid situations where shiv bundles packages from pyproject.toml, and we test everything with packages from poetry.lock.


fredo (approved):
looks good!

Issue #1076, opened 2022-09-26
Darkmode of docs is not nice for graphs


czepluch:
#### Expected behavior
Our docs support dark mode, but some of the graphs are hard to read when the background is dark. It should be possible to read the graphs in both dark and light mode.
See attached photo:
<img width="745" alt="Screenshot 2022-09-26 at 16 23 53" src="https://user-images.githubusercontent.com/1026799/192301844-a7d38ccc-2ea2-44d5-8cd4-691fc7ff9246.png">

#### Steps to reproduce the behavior
Go to https://docs.beamerbridge.com/whitepaper.html and enable darkmode and scroll to a graph and try to read the dark text

#### Acceptance criteria
- [ ] Make the dark text on the graphs readable in darkmode

istankovic:
Thanks for submitting this!

andfletcher:
"[until stakes high enough for L1 resolution]" - is still black and can not be read.

Issue #333, opened 2022-03-13
Exploits to the agent on testnet


fredo:
# Malicious Fredo exploits the agent
I could not keep my hands still over the weekend. I was simply to hyped by what we have achieved and wanted to try out the protocol on testnet. I don't like to bring the bad news but I found some exploits to the protocol and the agent. The good news are: We have an awesome team and I think we can fix those issues in a reasonable amount of time. And it's better that we found out about that before announcing it to the public.

## Attack scenario
Bob is the honest agent and gets tricked by Malicious Alice (M'Alice) to not follow the protocol anymore. This will ultimately lead to the protocol losing funds or DOSing the protocol. 

Assumption: Events from one chain come in at chronological order.

### Explorers

[Request Manager on Boba](https://blockexplorer.rinkeby.boba.network/address/0xB1D64b5d92d0d5Ca0a23a0987Fe10A96c15868bB/transactions)
[Fill Manager on Metis](https://stardust-explorer.metis.io/address/0xA2e8228A4De5976f584471056B04Da0B9f0261c8/transactions)

### Function selectors 
For a better understanding of transaction see the methods' selectors

#### Request Manager
- createRequest: 0x2f4212b8
- claimRequest: 0xb1743c5b
- challengeClaim: 0x92c824d8
- withdraw: 0x2e1a7d4d
#### Fill Manager
- fillRequest: 0xefd39926

### Fill data not checked #336 
- M'Alice sends wrong fill (i.e. amount = 1)
- Bob does not check the metadata of the fill and will assign M'Alice as the filler to the request
- M'Alice claims
- Bob will not challenge as M'alice is assigned as the filler

#### Successful exploit on testnet
- [Request](https://blockexplorer.rinkeby.boba.network/tx/0x30674ff6fd6a454f72b349ea68382c8a0664e14fb2e121973d28e38571e04a1c/token-transfers)
- [Fill](https://stardust-explorer.metis.io/tx/0xdfb4e834c447af2ea324d7e41dbeceb476007887355e5d17a1dedaf08fcad067/token-transfers)
- [Claim](https://blockexplorer.rinkeby.boba.network/tx/0x73a9066e3eecab0dc347d2ca3b306bf8761d41232b5f38af259eaa40831e16ad/internal-transactions)
- [Withdraw](https://blockexplorer.rinkeby.boba.network/tx/0x23c2fd04ce0103712c93c8b816833937b79bb57c21b40bef586b1a8164245bf0/token-transfers)
[...]

PR #1728, opened 2023-04-19
docs: adapt mermaid to dark mode

bilbeyt:
Closes: #1076

### DIFF
--- a/docs/source/_static/css/styles.css
+++ b/docs/source/_static/css/styles.css
@@ -1,1 +1,40 @@
-@import url('https://fonts.googleapis.com/css2?family=Sora:wght@300;400;500;600;700&display=swap');
+@import url("https://fonts.googleapis.com/css2?family=Sora:wght@300;400;500;600;700&display=swap");
+
+@media (prefers-color-scheme: light) {
+  body[data-theme="dark"] .mermaid line {
+    stroke: white !important;
+  }
+
+  body[data-theme="dark"] .mermaid path {
+    stroke: white !important;
+  }
+
+  body[data-theme="dark"] .mermaid .messageText {
+    stroke: none;
+    fill: white !important;
+  }
+
+  body[data-theme="dark"] .mermaid #arrowhead path {
+    stroke: white !important;
+    fill: white !important;
+  }
+}
+
+@media (prefers-color-scheme: dark) {
+  body:not([data-theme="light"]) .mermaid line {
+    stroke: white !important;
+  }
+
+  body:not([data-theme="light"]) .mermaid path {
+    stroke: white !important;
+  }
+
+  body:not([data-theme="light"]) .mermaid .messageText {
+    stroke: none;
+    fill: white !important;
+  }
+  body:not([data-theme="light"]) .mermaid #arrowhead path {
+    stroke: white !important;
+    fill: white !important;
+  }
+}


docs/source/conf.py
```diff
@@ -96,3 +96,4 @@
 html_permalinks = False
 html_show_sphinx = False
 todo_include_todos = True
+mermaid_output_format = "png"
```
istankovic, line 99: How/why does this help?

bilbeyt: mermaid handles things differently looking at output format changing the code to embed to html etc. with image png it is not changing anything with theme changes and you can easily see things inside basically, with svg it is using different thing and you need to customize everything.

bilbeyt: should be much better right @istankovic ?

bilbeyt: hard to understand i can see it normally. is there a chance it is using css classes or something in different browsers? which browser are you using? i am using chrome and it was ok.

istankovic: I'm using Firefox. With Chromium I get it the way it's supposed to be (by the way, much nicer!):

image

bilbeyt: yeah should be fine now. we had a look with @GabrielBuragev and found a much better css to handle this.

istankovic: > yeah should be fine now. we had a look with @GabrielBuragev and found a much better css to handle this.

Hmmm, I still have the same issue with my Firefox on ad56603f90ff8a816891b0059583fe53f6111714.
I did a `make clean` before `make docs`.

bilbeyt: i think you can check this again

istankovic: This is what I get with af5a564db39aa71a5a9bfde695816c43389da107:
image

istankovic: Compared to the Chromium's rendering, Firefox is more blurry, almost looks like bold.
I have no idea what the problem might be, but let's just blame Firefox or my local setup for it.
(It would be nice if the dark theme worked nicely on all major browsers, though. :disappointed: )

istankovic: @bilbeyt is there something we could do about the ugly text rendering in Firefox?

GabrielBuragev: @bilbeyt I also see the texts blurred as on the screenshot Ivan provided. 
The problem here is that the css holds these values for the texts:
`fill: #333;`

This in fact fills the text with black color. 
You might need to also overwrite this fill styling by expanding the css for the message elements like so:

```
body:not([data-theme="light"]) .mermaid .messageText {
        stroke: white !important;
        fill: white !important;
    }
```

GabrielBuragev: After adding the above, the text should look like this:

Screenshot 2023-04-24 at 10 26 42

bilbeyt: much better thanks @GabrielBuragev will add it also

GabrielBuragev: LGTM ✅

GabrielBuragev (approved):
Looks good.

istankovic (changes requested):
Still needs improvement.
The latest image does not look as crisp as the original one from Chromium.

GabrielBuragev:
@bilbeyt I am working on it, pushing a fix soon

istankovic (approved):
This looks good, thanks!

PR #1731, opened 2023-04-20
docker: scripts: add a container image test

istankovic:
This PR
- adds a script that tests the beamer container image
- updates CI workflows so that the `build-and-publish-container-image` is run as part of `backend` and `relayer` workflows; pushes to the container registry will only be done if we're on `main` branch

The test consists of starting a local ganache instance, deploying Beamer onto it, running an agent using the specified container image and checking whether the agents was able to sync successfully.


bilbeyt (approved):
Looks really good

PR #1732, opened 2023-04-21
frontend: fix source token selection reactivity

### LINKED ISSUE #1710: Frontend: App misbehaves when wallet is not connected
Our app doesn't properly pick up the change in the source chain/source token input selection. 
From my first look into the issue, i could determine the source of it. It happens when the source chain selection has been changed after a previous selection of a source token for a different source chain.

This happens due to the fact that the source token selection is not updated with the new values for the newly selected source chain. They still hold the values from the previously selected source chain (contract addresses) and this causes the app to misbehave: fail to properly calculate fees & transfer limits as it is looking up at the wrong contract addresses.

#### Expected behavior
Token selection are updated when a new chain has been selected. We could also just reset the token selection to a clean state (de-select) 

#### Actual behavior

https://user-images.githubusercontent.com/18466333/231970505-a8ddbaa2-23c4-476a-ac71-f2f21869686b.mov

# Acceptance Criteria
- [ ] App properly reacts to source chain change so that token selections are reset/updated to reflect the new state

GabrielBuragev:
Fixes #1710

### DIFF
--- a/frontend/src/components/RequestSourceInputs.vue
+++ b/frontend/src/components/RequestSourceInputs.vue
@@ -286,6 +286,11 @@ const totalRequestTokenAmount = computed(() => {
 });
 
 watch(totalRequestTokenAmount, () => v$.value.selectedTokenAmount?.$touch());
+watch(selectedSourceChain, () => {
+  selectedToken.value =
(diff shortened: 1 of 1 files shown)


manuelwedler (approved):
Tested it and fixes the issue. Thanks!

PR #1739, opened 2023-04-24
fix: avoid race condition of deleting the l1 future

bilbeyt:


fredo (approved):

beamer/agent/chain.py
```diff
@@ -569,7 +569,7 @@ def on_future_done(f: Future) -> None:
         except Exception as ex:
             context.logger.error("L1 resolution failed", ex=ex, tx_hash=proof_tx)
         finally:
-            del context.l1_resolutions[proof_tx]
+            context.l1_resolutions.pop(proof_tx, None)
```
istankovic, line 572: How does this "avoid race condition of deleting the l1 future"?

bilbeyt: there are 2 ways to delete this, first place is here and second place is handling resolution events. we are already using pop for event but in test @fredo working on we are getting event before relayer process ends so in here directly error thrown.

istankovic: > there are 2 ways to delete this, first place is here and second place is handling resolution events. we are already using pop for event but in test @fredo working on we are getting event before relayer process ends so in here directly error thrown.

Ah, I see, thanks for clarifying. :+1: 
(I was surprised to see the above change described as "avoiding race condition" because ti does not change the atomicity of the operation, as far as I'm aware; so the race condition is still there, it's just that `pop` will silently ignore the case the entry is not in the dict, wherease the `del` would complain.)

Issue #1729, opened 2023-04-19
Connection error to RPC needs to halt the agent


fredo:
The event fetcher handles a connection error and just retries fetching the events. The event processor doesn't know about this and accepts the current state as the synced state.
This can lead to several problems, as the agent reacts on the state he assumes is the real state on the chains. This can lead to a loss of funds. For example, with a broken target chain RPC the agent could challenge valid claims.

We need to make sure that the agent stops with potential transactions which are not safe if the state is not up to date.
However, the agent should continue to run because other event processors may not be affected by the broken RPC.

### How to evaluate if the RPC is broken
We could introduce an easy check upon every iteration what the timestamp of the latest block is. If the latest block is n minutes ago, then stop processing. I think it is enough to do this check before `process_claims` and `process_requests`, because these are the functions generating an output which could lead to loss of funds. 

Another option would be the introduction of another communication channel (or reusing the event queue) to communicate to the EventProcessor that a ConnectionError happened. After another successful request the EventProcessor could resume it's work. This would be more time stricter.

### Open question
We should thoughtfully research whether stopping an EventProcessor can lead to a loss of funds in other occasions. For example, There could be an ongoing challenge right before the termination, and this would be actually unaffected by a broken target chain rpc. If we stop the EventProcessor the agent may lose funds because he is not able to l1 resolve or counter challenge

## Acceptance criteria

- [ ] Mark the event monitor as down and communicate to the event processor
- [ ] holt the event processor until the event fetcher is working again 
- [ ] test for specific test challenge situation we had

PR #1742, opened 2023-04-25
feat: check rpc status notifications and halt event processors

bilbeyt:
Closes: #1729 

This PR looks for notifications from EventFetcher and EventMonitor and pushes the state to EventProcessor and if rpc is not working, EventProcessor halting processing claims and requests.

### DIFF
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -6,8 +6,9 @@ import traceback
 from concurrent.futures import Future
 from typing import Callable
 
+import requests
 import structlog
-from web3 import Web3
+from web3 import HTTPProvider, Web3
 from web3.contract import Contract
 from web3.types import Timestamp, Wei
 
@@ -27,6 +28,7 @@ _STOP_TIMEOUT = 2
 
 
 _SyncDoneCallback = Callable[[], None]
+_RPCStatusCallback = Callable[[bool], None]
 _NewEventsCallback = Callable[[list[Event]], None]
 
 
@@ -51,6 +53,7 @@ class EventMonitor:
         deployment_block: BlockNumber,
         on_new_events: list[_NewEventsCallback],
         on_sync_done: list[_SyncDoneCallback],
+        on_rpc_status_change: list[_RPCStatusCallback],
         poll_period: float,
         confirmation_blocks: int,
     ):
@@ -61,6 +64,8 @@ class EventMonitor:
         self._stop = False
         self._on_new_events = on_new_events
         self._on_sync_done = on_sync_done
+        self._on_rpc_status_change = on_rpc_status_change
+        self._rpc_working = True
         self._poll_period = poll_period
         self._confirmation_blocks = confirmation_blocks
         self._log = structlog.get_logger(type(self).__name__).bind(chain_id=self._chain_id)
@@ -81,6 +86,25 @@ class EventMonitor:
     def subscribe(self, event_processor: "EventProcessor") -> None:
         self._on_new_events.append(event_processor.add_events)
         self._on_sync_done.append(event_processor.mark_sync_done)
+        self._on_rpc_status_change.append(event_processor.set_rpc_working)
+
+    def _inner_fetch(self, fetcher: EventFetcher) -> list[Event]:
+        events = []
+        was_working = self._rpc_working
+        try:
+            events.extend(fetcher.fetch())
+        except requests.exceptions.ConnectionError:
+            self._rpc_working = False
+        else:
+            self._rpc_working = True
+        if was_working != self._rpc_working:
+            self._call_on_rpc_status_change(self._rpc_working)
+            assert isinstance(self._web3.provider, HTTPProvider)
+            self._log.info(
+                "RPC stopped working" if was_working else "RPC started working",
+                rpc_url=self._web3.provider.endpoint_uri,
+            )
+        return events
 
     def _thread_func(self) -> None:
         self._log.info(
@@ -93,13 +117,13 @@ class EventMonitor:
         current_block = self._web3.eth.block_number
         events = []
         while fetcher.synced_block < current_block:
-            events.extend(fetcher.fetch())
+            events.extend(self._inner_fetch(fetcher))
         if events:
             self._call_on_new_events(events)
         self._call_on_sync_done()
         self._log.info("Sync done")
         while not self._stop:
-            events = fetcher.fetch()
+            events = self._inner_fetch(fetcher)
             if events:
                 self._call_on_new_events(events)
             time.sleep(self._poll_period)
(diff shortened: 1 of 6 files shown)


istankovic:
@bilbeyt I thought we would not go with a time-based approach here, but rather mark the relevant processors when the event monitor/fetcher notices a failure.

bilbeyt:
> @bilbeyt I thought we would not go with a time-based approach here, but rather mark the relevant processors when the event monitor/fetcher notices a failure.

Well in issue, it is still written like this with time based approach

istankovic:
> > @bilbeyt I thought we would not go with a time-based approach here, but rather mark the relevant processors when the event monitor/fetcher notices a failure.
> 
> Well in issue, it is still written like this with time based approach

Yeah, but I think we simply didn't bother to update the issue.

bilbeyt:
> > > @bilbeyt I thought we would not go with a time-based approach here, but rather mark the relevant processors when the event monitor/fetcher notices a failure.
> > 
> > 
> > Well in issue, it is still written like this with time based approach
> 
> Yeah, but I think we simply didn't bother to update the issue.

should be ok now

beamer/agent/chain.py
```diff
@@ -99,7 +105,15 @@ def _thread_func(self) -> None:
         self._call_on_sync_done()
         self._log.info("Sync done")
         while not self._stop:
-            events = fetcher.fetch()
+            try:
+                events = fetcher.fetch()
+                if self._latest_rpc_status is False:
+                    self._latest_rpc_status = True
+                    self._call_on_rpc_status_change(True)
+            except (ConnectionError, HTTPError):
```
istankovic, line 113: We should always try to put the minimum necessary amount of code into `try...except` blocks.
Here, the `if` statement, with it's included block, cannot throw these exceptions so we should move them out of the `try...except` block.

beamer/agent/chain.py
```diff
@@ -99,7 +105,15 @@ def _thread_func(self) -> None:
...
+                if self._latest_rpc_status is True:
```
istankovic, line 114: PEP8: please do not compare booleans via `is True` or `is False`.

beamer/agent/chain.py
bilbeyt: I was thinking this would be a suggestion to me to do that actually to optimize otherwise i would be constantly changing the flag, the first thinking was just update all the time and changed to this one to reduce the number of calls actually

istankovic: How about this?

```python
            try:
                events = fetcher.fetch()
            except (ConnectionError, HTTPError):
                was_working = self._rpc_working:
                self._rpc_working = False
                if was_working:           
                    self._call_on_rpc_status_change(False)
            else:
                was_working = self._rpc_working:
                self._rpc_working = True
                if not was_working:           
                    self._call_on_rpc_status_change(True)
```

bilbeyt: well except part didnt change in some view but you moved the first part to else, i am ok with this

istankovic, line 113: Another question: why do we catch `HTTPError`?

bilbeyt: I saw HTTPError on EventFetcher is just there in _fetch_range raising in a scenario but it is not handled so would just crash the agent so i just added this also here

beamer/agent/chain.py
```diff
@@ -99,7 +105,18 @@ def _thread_func(self) -> None:
...
+            except (ConnectionError, HTTPError):
+                was_working = self._rpc_working
+                self._rpc_working = False
+                if was_working:
+                    self._call_on_rpc_status_change(False)
+            else:
+                was_working = self._rpc_working
+                self._rpc_working = True
+                if not was_working:
+                    self._call_on_rpc_status_change(self._rpc_working)
```
istankovic, line 119: It occurred to me that even better might be

```python
was_working = self._rpc_working                       
try:                                                  
    events = fetcher.fetch()                          
except (ConnectionError, HTTPError):                  
    self._rpc_working = False                         
else:                                                 
    self._rpc_working = True                          
if was_working != self._rpc_working:                  
    self._call_on_rpc_status_change(self._rpc_working)
```

beamer/agent/chain.py
istankovic: > I saw HTTPError on EventFetcher is just there in _fetch_range raising in a scenario but it is not handled so would just crash the agent so i just added this also here

I would actually prefer in this case to crash the agent and know what happened, i.e. the RPC might still be working, but it could be that they don't like some of our requests. If we just ignore those, we will never find out about such RPCs and will therefore not be able to help the LPs using those RPCs. For that reason, I would catch only `ConnectionError` here.

bilbeyt: ok

beamer/tests/agent/test_challenge.py
```diff
@@ -497,3 +498,63 @@ def test_invalidation(
...
+    with Sleeper(5) as sleeper:
+        while (request := agent.get_context(direction).requests.get(request_id)) is None:
+            sleeper.sleep(0.1)
+
+    with Sleeper(10) as sleeper:
+        while not request.claimed.is_active:
+            sleeper.sleep(0.1)
+
+    collector = EventCollector(request_manager, "ClaimMade")
+    claim = collector.next_event(20)
```
istankovic, line 533: Why such a long wait time, 20s?

beamer/tests/agent/test_challenge.py
```diff
@@ -497,3 +498,63 @@ def test_invalidation(
...
+
+    request_manager.challengeClaim(
+        claim.claimId, sender=exploiter, value=ape.convert("2 ether", int)
+    )
+
+    exploiter_claim = collector.next_event()
+
+    assert exploiter_claim is not None
+
+    # as agent lost the rpc couldnt challenge back event processor is halt
```
istankovic, line 547:
```suggestion
    # as agent lost the RPC connection, it couldn't counter-challenge because the event processor is halted
```

beamer/tests/agent/test_challenge.py
```diff
@@ -497,3 +498,63 @@ def test_invalidation(
...
+    challenge_claim = collector.next_event()
+
+    assert challenge_claim is None
+
+    proxy_l2a.start()
+
+    # as agent has the rpc again, start processing events and counter challenge
```
istankovic, line 554:
```suggestion
    # as agent has the RPC connection again, start processing events and counter challenge
```

beamer/tests/agent/test_challenge.py
```diff
@@ -497,3 +498,63 @@ def test_invalidation(
...
+    challenge_claim = collector.next_event()
+
+    assert challenge_claim is not None
+
+    agent.stop()
+    proxy_l2a.stop()
```
istankovic, line 560: Great test! :+1:

istankovic (approved):
We should log whenever the RPC status changes and there are some minor comment nitpicks, but I think this looks great now. Thanks!

Issue #1706, opened 2023-04-13
Script to run owner commands on all contracts on all chains of a given deployment


fredo:
We frequently need to adjust configurations on the contracts which have the `onlyOwner` modifier. 

In order to save time, provide a script which consumes a config with the parameters to configure and the deployment directory and execute the configuraition with the owner key on all chains.

fredo:
closing in favor of #972

PR #1749, opened 2023-04-25
Beamer configure command

fredo:
implements: #1706

I tried a generic approach. The general problem is that it is hard to map input to output of the contract. Namely, when we have an input function (i.e. `updateChain`) where and how can I read the output (what are the current values)? Generically this is difficult to achieve.

I started now with the first pattern we have, and the script tries to find a the corresponding mapping from input to storage (setter and getter). This can't really be read from the ABI.

Currently the script does: 

updateChain maps to the getter -> chains
updateToken maps to the getter -> tokens
... 

I removed modifying the allowlist which is a different pattern (No values to be set but bool to be read)

From here we could move forward and decide how to evolve this script.

Additionally: 
The script fills missing params with currently stored values
The script checks whether the storage would be changed by executing. If not, don't send transaction.

Future Improvements: 
* There is a lot of room to improve the apischema for type validation
* check that changes were effective
* add more commands
* Better feedback if names cannot be found in the ABI
* We could build a exclusive mapping for getters and setters (less generics)
* ...

### DIFF
--- a/beamer/configure/config.py
+++ b/beamer/configure/config.py
@@ -1,0 +1,39 @@
+import json
+from dataclasses import dataclass, field
+from pathlib import Path
+
+from apischema import deserialize, schema
+
+from beamer.agent.typing import URL
+
+
+@dataclass
+class ChainConfig:
+    name: str
+    rpc: URL
+    tokens: list[list[str]] = field(default_factory=list)
+
+
+@dataclass
+class Param:
+    key: str | int | None = field(default=None)
+    values: dict[str, str | int] = field(default_factory=dict)
+
+
+@dataclass
+class Command:
+    name: str
+    params: list[Param]
+
+
+@dataclass
+class Config:
+    deployment_dir: str
+    chains: dict[str, ChainConfig] = field(metadata=schema(min_items=1))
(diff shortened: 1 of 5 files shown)


beamer/configure/config.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @staticmethod
+    def from_file(config_file: Path) -> "Config":
+        with open(config_file) as f:
+            config = json.load(f)
+        try:
+            config = deserialize(Config, config)
+        except ValidationError:
+            traceback.print_exc()
+        return config
```
istankovic, line 44: It seems weird that we catch `ValidationError`, print the traceback and then just continue.
I assume this would result in `NameError` being raised, which would be misleading.

Personally, I'd just omit the whole `try...except` block and let the exception propagate, terminating the Python process (and our command) with an exit code. The traceback would be output automatically.

For comparison, in `scripts/deployment/config.py` we do this:
```
try:                                    
    config = deserialize(Config, config)
except ValidationError as exc:          
    raise ConfigValidationError from exc
```

fredo: you are completely right

fredo:
will close this due to the creation of new iissues regarding the configure script

PR #1756, opened 2023-04-27
docs: improve wording in the description of the release steps

istankovic:
Make it also more robust and always trigger the CI run manually after pushing the release tag.

### DIFF
--- a/docs/source/releasing.rst
+++ b/docs/source/releasing.rst
@@ -5,11 +5,17 @@ Making a new release
 
 #. Make sure to populate ``CHANGELOG.md`` and change the version number
    (see `this commit <https://github.com/beamer-bridge/beamer/commit/440b7ddffc01d16482d78ff9f18a8830670795bc>`_ for example).
-#. Commit the changes above, create and merge the PR
-   (if releasing from ``main``; there should be no PR for releases from other branches).
-#. Tag the resulting commit on ``main`` or on the release branch (e.g. ``git tag v0.1.8``) and
-   push it to Github (e.g. ``git push origin tag v0.1.8``).
-#. Make sure a CI run is triggered, either automatically or manually, and wait for the run to finish.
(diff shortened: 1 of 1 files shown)


docs/source/releasing.rst
```diff
@@ -5,11 +5,16 @@ Making a new release
...
+#. Commit the changes above, create and merge the PR.
+#. Take note of the commit ID on the release branch (either ``main`` or one of the ``N.x`` branches,
+   depending on where you are releasing from). Tag that commit (e.g. ``git tag v0.1.8 COMMIT_ID``) and
+   push the tag to Github (e.g. ``git push origin tag v0.1.8``).
+
+    .. note:: The ID of the last commit on the release branch may be different from the ID of
+              the last commit on the PR branch, even in cases where those commits have identical changes.
+              Make sure to tag the commit on the release branch, not on the PR branch.
+
+#. Once the tag is pushed, trigger a new CI run manually, to ensure that the built container image is tagged properly.
```
fredo, line 17:
```suggestion
#. Once the tag is pushed, trigger a new CI run on main manually, to ensure that the built container image is tagged properly.
```

fredo (approved):
made one suggestion but otherwise looks good

docs/source/releasing.rst
istankovic: The tag might point to a commit on a branch other than `main`, e.g. `0.x` where we have tag `v0.1.8`. So you need to actually trigger the CI run on the branch containing the commit you tagged.

So something like the following, to be extra clear?
```markdown
#. Once the tag is pushed, manually trigger a new CI run on the branch containing the tagged commit,
   to ensure that the built container image is tagged properly.
```

fredo: oh good point! yes, go with your suggestion

Issue #1745, opened 2023-04-25
Improve code reuse


istankovic:
In order to be able to share more code among different software components, including various scripts that import the `beamer` package we should move common, useful functionality and modules from the `beamer.agent` subpackage to the toplevel `beamer` package i.e. one level up.

 Tasks:
- [x] make the `account` argument from `beamer.agent.util.make_web3` optional and use the same function in `health`, removing the duplicate there
- [x] move most of `beamer.agent.util` to `beamer.util`; the token checker related things should stay in `beamer.agent.util`
- [x] move  `beamer.agent.middleware` to `beamer.middleware`
- [x] move  `beamer.agent.typing` to `beamer.typing`
- [x] move  `beamer.agent.events` to `beamer.events`
- [x] move  `beamer.agent.contracts` to `beamer.contracts`
- [x] move `beamer/agent/data/abi/StandardToken.json` to `beamer/data/abi/StandardToken.json`
- [x] rename `beamer.agent.l1_resolution` to a more-appropriate `beamer.agent.relayer`
- [x] move `TokenDetails`, `get_token_details`, `get_token_balance`, `get_token_amount_in_decimals`  from `beamer.health.util` to `beamer.util`
- [x] remove `beamer.health.util` (should not be necessary now)
- [x] turn `load_ERC20_abi` into a private function, which is called only once to initialize a singleton `_ERC20_ABI` object, introduce a new function `get_ERC20_abi` that returns the singleton and convert all existing callers of `load_ERC20_abi` to use the new getter; this will avoid repeatedly deserializing the ABI data
- [x] fix the fallout in tests and CI

istankovic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-62442a69c5e6e60010e31cf3/issues/beamer-bridge/beamer/1745?planning-poker) with Zenhub @fredo @bilbeyt @compojoom @GabrielBuragev @manuelwedler

fredo:
great summary!

PR #1760, opened 2023-04-28
improve code reuse

bilbeyt:
Close: #1745


beamer/tests/agent/unit/test_handlers.py
```diff
@@ -25,8 +23,10 @@
 )
 from beamer.tests.agent.utils import make_address
 from beamer.tests.constants import FILL_ID
+from beamer.typing import FillId, Termination
+from beamer.util import get_ERC20_abi
 
-ERC2O_ABI = load_ERC20_abi()
+ERC2O_ABI = get_ERC20_abi()
```
istankovic, line 29: Now that we load the ERC20 abi just once, we can remove this global and instead just call `get_ERC20_abi`.

beamer/agent/chain.py
```diff
@@ -12,15 +12,15 @@
...
+from beamer.agent.relayer import run_relayer_for_tx
 from beamer.agent.state_machine import Context, process_event
-from beamer.agent.typing import BlockNumber, ChainId
-from beamer.agent.util import TransactionFailed, load_ERC20_abi, transact
+from beamer.events import Event, EventFetcher
+from beamer.typing import BlockNumber, ChainId
+from beamer.util import TransactionFailed, get_ERC20_abi, transact
 
-_ERC20_ABI = load_ERC20_abi()
+_ERC20_ABI = get_ERC20_abi()
```
istankovic, line 23: We can get rid of this global now and instead just call `get_ERC20_abi` where needed.

istankovic:
We should also move `contracts_for_web3` from `scripts/_util.py` to `beamer.contracts`.

istankovic:
```
>>> import beamer.agent.relayer as r
>>> r.get_relayer_executable()
PosixPath('XXX/beamer/beamer/agent/data/relayers/relayer-node18-linux-x64')
```

This function returns an incorrect path to the relayer.

istankovic (changes requested):
Overall pretty good, but there are few things to be fixed.

bilbeyt:
> ```
> >>> import beamer.agent.relayer as r
> >>> r.get_relayer_executable()
> PosixPath('XXX/beamer/beamer/agent/data/relayers/relayer-node18-linux-x64')
> ```
> 
> This function returns an incorrect path to the relayer. I wonder how this got past our CI. Do you have an idea?

paid special attention but still missed this good catch :)

istankovic:
> paid special attention but still missed this good catch :)

What is concerning is that the CI did not catch it.
I think we should not fix this until we get the CI to fail first because of this.
(That could require writing a new test, or fixing an existing one, if there is one.)

istankovic:
The Makefile needs updating, we copy relayer binaries to the old location.

istankovic (approved):
Good work!

Issue #1209, opened 2022-11-21
Frontend: Integrate toast notifications on transfer completed


GabrielBuragev:
# Rationale
To improve the UX of our transfer request flow, it would be very beneficial to integrate toast notifications for specific status changes of the transfer requests.
Since we are currently allowing multiple requests to be created, from the user's perspective, it is easy to get lost in the list of transfers in the transfer tab. 
Toast notifications are one way to improve this by updating the user on the transfer status changes.

## Acceptance criteria
- [ ] Needs to be reusable for other toast notifications (quickly adding new notifications across the app codebase)
- [ ] Preferably, it should be implemented/integrated as a floating card that fades away after N seconds by default or is closeable by a user action. --> configurable 
- [ ] The position of the card should be configurable. 
- [ ] Top-right by default? (most intuitive position IMO)
- [ ] Includes information on how long it took

GabrielBuragev:
@andfletcher @DBecane What should we write inside the "transfer completed" toast notification?

Please note that we will also include the "share via twitter" (#1684) button inside this notification.
Currently it looks like this:
<img width="1716" alt="Screenshot 2023-04-27 at 10 47 42" src="https://user-images.githubusercontent.com/18466333/234810021-b3e36aa7-1e2c-49b6-894d-56f11cd18f0b.png">

Me and @olympe-silve are already working on the design.

Issue #938, opened 2022-08-17
Frontend: Add tailwindcss plugin to prettier


manuelwedler:
I just found out that there is a prettier plugin for sorting the tailwind classes in our templates. Would be nice to add it. It's much easier to get what the classes do, if they follow a specific order.

See https://github.com/tailwindlabs/prettier-plugin-tailwindcss

# Acceptance criteria
- [ ] Plugin is intergrated
- [ ] Whole frontend codebase is formatted

GabrielBuragev:
Nice one! We will also be able to learn the ordering pattern of the tailwind classes with the help of this plugin.

PR #1765, opened 2023-05-02
Frontend/toast notifications

### LINKED ISSUE #1684: Share experience via twitter after bridge transfer
## Description
As a User, I want to share my bridging experience via Twitter once I completed a bridge transfer to share the positive user experience and speed of Beamer Bridge.

## Acceptance criteria
The information displayed on the toast is also sharable via twitter. #1209
Users can share this experience with a prewritten text on Twitter (i have seen this on some other pages, so this might be a Twitter feature.) that includes the information mentioned above (don't include fees or amount transferred)

GabrielBuragev:
Closes #1684, #1209, #938


manuelwedler:
I think the positioning on mobile doesn't look perfect. A little space around the toast would be nice.

frontend/src/composables/useTransferNotifications.ts
```diff
@@ -0,0 +1,41 @@
...
+  const activeTransfers = computed(() => {
+    return transfers.value.filter((transfer) => !transfer.done);
+  });
+
+  const setupListeners = () => {
+    for (const transfer of activeTransfers.value) {
+      transfer.removeAllListeners();
+
+      transfer.on('completed', () => {
+        toast.success(
```
manuelwedler, line 20: I think we should also remove the listener here in order to clear the references. 

Also, in the case that a transfer fails, it might happen that a listener is never removed.

frontend/src/components/notifications/TransferComplete.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+  <div>
+    <span data-test="message">
+      <span v-if="transfer?.transferTime !== undefined"
+        >Transfer completed in {{ transfer?.transferTime }}s.</span
```
manuelwedler, line 5: In my opinion it would be nice to have some additional info about the transfer in the notification. Like from which rollup to which and what token.

frontend/src/actions/steps/multi-step-action.ts
```diff
@@ -121,6 +124,8 @@ export class MultiStepAction {
         step.deactivate();
       }
     }
+
+    this.emit('completed');
```
manuelwedler, line 128: Wouldn't it be nice to also show a notification if a transfer fails?

frontend/tests/unit/actions/steps/multi-step-action.spec.ts
```diff
@@ -247,4 +247,25 @@ describe('MultiStepAction', () => {
...
+      const one = vi.fn();
+      const two = vi.fn();
+      const methods = { one, two };
+      const emitMock = vi.fn();
+      action.emit = emitMock;
+
+      try {
+        await action.executeSteps(methods);
+      } catch {
+        /* ignore */
```
manuelwedler, line 265: Why ignore? If you want to make the test fail when an error occurs, you can simply let it get thrown.

manuelwedler (commented):
Thanks for the effort! Nice feature! Just a few things I would like to get addressed / clarified.

frontend/src/actions/steps/multi-step-action.ts
GabrielBuragev: I didn't consider that as a part of the scope of this issue for the following reason:
- It requires some more thought process on what to display and how to fetch the right error messages so its somewhat human readable (as opposed to the current error messages)
- I consider this to also go hand in hand with this other big issue: https://github.com/beamer-bridge/beamer/issues/507

We could do a simple "Transfer failed" notificaiton but i am not sure how much benefit that will provide.

Your point is right though. As a user, if i figured out that the app shows a successful notification on transfer completion, i would also expect to get a notification for a tranfer failure as well.

frontend/src/components/notifications/TransferComplete.vue
GabrielBuragev: I also think it doesn't provide much info but we also don't want to bloat them with too much text since we already have all of this info inside the cards. 

I initially thought of linking the toast notification with the transfer card in the following way:
Once the user clicks on the successful toast notification we programatically open the "Activity" tab, scroll to the specific transfer card in the list & expand it.
I gave this suggestion to Olympe but she didn't like it.

Anyways, since this would require a discussion which will lead to a decision, maybe its better to create an issue for this.

frontend/src/composables/useTransferNotifications.ts
GabrielBuragev: Indeed, somehow overlooked this. Will implement accordingly.

frontend/src/composables/useTransferNotifications.ts
```diff
@@ -0,0 +1,41 @@
...
+          {
+            component: TransferComplete,
+            props: { transfer },
+          },
+          {
+            timeout: false,
```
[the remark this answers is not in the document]
GabrielBuragev: Good catch actually. I forgot to add the 5s timeout that was previously decided upon.

frontend/src/plugins/vue-toastification.ts
```diff
@@ -0,0 +1,17 @@
...
+  container: () => document.querySelector('#app') as HTMLElement,
+  toastClassName: '!rounded-lg',
+  toastDefaults: {
+    success: {},
+    error: {},
+    warning: {},
+  },
+};
+
+export { default as ToastPlugin } from 'vue-toastification';
```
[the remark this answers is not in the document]
GabrielBuragev: I think i wanted to contain all the code related to the setup of `vue-toastification` library inside one file so it doesn't bloat the `main.ts` file.

frontend/tests/unit/actions/steps/multi-step-action.spec.ts
GabrielBuragev: Probably a copy/paste mistake here. Fixed it.

GabrielBuragev:
> I think the positioning on mobile doesn't look perfect. A little space around the toast would be nice.
Fixed this.

frontend/src/composables/useTransferNotifications.ts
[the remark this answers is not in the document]
GabrielBuragev: Used 10 seconds instead of 5. We changed the decision today.

frontend/src/plugins/vue-toastification.ts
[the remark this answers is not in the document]
manuelwedler: Ok makes sense. Let's do it like that :)

frontend/src/actions/steps/multi-step-action.ts
manuelwedler: I think this can still be done separately from the mentioned issue. It is clear when a transfer is in a failed state (`failed` getter) and there will be an error message set (`errorMessage` getter). But well, it's true that not all error messages are user friendly at the moment and this needs improvement as the issue suggests.

So yeah let's not handle it in the scope of this PR but write an issue for it.

frontend/src/components/notifications/TransferComplete.vue
manuelwedler: Ok I also just thought it would be nice to link to the activities tab but if Olympe doesn't like it we shouldn't do it.

Can't we just add the text from the titles in the activities tab? For example "Transferred 1.0 TST to Optimism Goerli Testnet in 5.4 s". Was the text given by Olympe?

manuelwedler (approved):
Just fix the nitpick and this can be merged if we don't want to change the text.

frontend/src/composables/useTransferNotifications.ts
```diff
@@ -12,21 +12,26 @@ export function useTransferNotifications(transfers: Ref<Array<Transfer>>) {
...
+  const onTransferCompleted = (transfer: Transfer) => {
+    toast.success({
+      component: TransferComplete,
+      props: { transfer },
+    });
+  };
+
   const setupListeners = () => {
     for (const transfer of activeTransfers.value) {
       transfer.removeAllListeners();
```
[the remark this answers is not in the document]
GabrielBuragev: 👍

frontend/src/components/notifications/TransferComplete.vue
olympe-silve: @GabrielBuragev @manuelwedler I'd much rather a link to the activity tab in the toast, than more info in the toast itself. It could be a link through the underlined "Transferred", for example. 

I don't want the toast itself (as in, the shape) to be clickable because : 1/ it won't be logical that this one is clickable and not the others, 2/ the user would have to click with targeted precision on the toast's shape itself and not the "share on twitter" link.

Issue #973, opened 2022-08-26
Add a script to produce contract ABI files for a given deployment


istankovic:
Assuming #972 is done, it should be possible to produce a set of contract ABI files for a given deployment.
The new script could be invoked like this:
```
python scripts/generate-abi.py deployments/output/mainnet deployments/output/mainnet/abi
```
The first argument to the script is the path to an existing deployment directory, and the second argument is the path that the script should use for generated ABI files.

The script should
- read `layer1.deployment.json` to obtain the git commit ID
- checkout the working tree corresponding to that commit ID to a temporary location
- invoke `ape` to compile the contracts
- copy ABI files to the output directory
- clean up the temporary directory

:warning: An important assumption here is that all deployed contracts have been compiled by taking the source defined by the commit ID recorded in `layer1.deployment.json`.

PR #1779, opened 2023-05-04
feat: generate abi based on artificat contracts commit_id

bilbeyt:
Closes: #973


scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
+import json
+import tempfile
+from pathlib import Path
+from typing import Any
+
+import click
+from psutil import Popen
+
+COMPILE_COMMAND = """
```
istankovic, line 9: Should be private.

scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
...
+git clone git@github.com:beamer-bridge/beamer.git {beamer_path} &&
+cd {beamer_path} &&
+git fetch &&
```
istankovic, line 12: Why do you need `fetch` if you just cloned?

scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
+import json
+import tempfile
+from pathlib import Path
+from typing import Any
+
+import click
+from psutil import Popen
+
+COMPILE_COMMAND = """
+git clone git@github.com:beamer-bridge/beamer.git {beamer_path} &&
```
istankovic, line 10: We can use `--no-checkout` option here to avoid git checking out the main branch by default, since we are anyway going to checkout our own commit.

scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
...
+cd {beamer_path} &&
+git fetch &&
+git checkout {commit_id} &&
+$(which ape) compile
```
istankovic, line 14: Why do you need to invoke `ape` like that?
It seems you can invoke `git` normally, so why should `ape` be different?

scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
...
+"""
+
+
+def _get_contract_commit_ids(deployment_path: Path) -> dict[str, set[str]]:
+    file_contract_commits: dict[str, set[str]] = {}
+    for file_path in deployment_path.glob("*.json"):
```
istankovic, line 20: Nitpick: maybe use `"*.deployment.json"` because that's how deployment files are marked?

scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
...
+        with open(file_path, "r") as file:
```
istankovic, line 21: Please use a name other than `file`, which is a builtin Python type. Commonly used names are `f`, `fp` etc.

scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
...
+            data = json.load(file)
+            data.pop("deployer")
+            for contract_name, contract_info in data.items():
+                if contract_name not in file_contract_commits:
+                    file_contract_commits[contract_name] = set()
+                file_contract_commits[contract_name].add(contract_info["beamer_commit"])
```
istankovic, line 27: Once you load the json from file, there is no need to keep the file open so we can just move these out of the context manager.

scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
...
+
+def _get_contract_commit_ids(deployment_path: Path) -> dict[str, set[str]]:
+    file_contract_commits: dict[str, set[str]] = {}
+    for file_path in deployment_path.glob("*.json"):
+        with open(file_path, "r") as file:
+            data = json.load(file)
+            data.pop("deployer")
+            for contract_name, contract_info in data.items():
+                if contract_name not in file_contract_commits:
+                    file_contract_commits[contract_name] = set()
```
istankovic, line 26: I think this condition will always be true, no?

scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
...
+                file_contract_commits[contract_name].add(contract_info["beamer_commit"])
+    return file_contract_commits
+
+
+def _compile_abis(commit_id: str, beamer_path: str) -> None:
+    Popen(
+        COMPILE_COMMAND.format(beamer_path=beamer_path, commit_id=commit_id).replace("\n", ""),
+        shell=True,
+    ).wait()
```
istankovic, line 35: If the command fails, what happens? Do we get an exception?
Also, I suggest to pull the command formatting out of the `Popen` call, for better readability & nicer formatting.

scripts/generate_abi.py
bilbeyt: i was thinking git would be already there in ci for example and ape is somehow connected to poetry run etc normally that made me do this but i guess there wont be any ape in ci without using some poetry command right?

scripts/generate_abi.py
bilbeyt: no this is for shared ones so there should be only one RequestManager key in the dict

scripts/generate_abi.py
bilbeyt: so this already happened actually, we are getting an error like tree is not there or something :D then execution ends

scripts/generate_abi.py
```diff
@@ -0,0 +1,97 @@
...
+def _get_contract_abis(
+    generated_abis: dict[str, dict[str, Any]]
+) -> dict[str, list[dict[str, Any]]]:
+    contract_abis: dict[str, list[dict[str, Any]]] = {}
+    for abis in generated_abis.values():
+        for contract_name, abi in abis.items():
+            if contract_name not in contract_abis:
+                contract_abis[contract_name] = []
+            contract_abis[contract_name].append(abi)
+    return contract_abis
```
istankovic, line 47: I have a feeling this function would not be needed if you modified the generated mapping in `_generate_abis` to look like
```
generated_abis[contract_name][commit_id]
```
instead of
```
generated_abis[commit_id][contract_name]
```

Put another way, this function should not be necessary if you generate the data in the desired shape directly in `_generate_abis`.

scripts/generate_abi.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def _merge_abis(contract_abis: dict[str, list[Any]]) -> dict[str, dict[str, Any]]:
+    merged_abis = {}
+
+    for contract_name, abis in contract_abis.items():
+        if all(abi == abis[0] for abi in abis):
+            merged_abis[contract_name] = abis[0]
```
istankovic, line 54: ABIs are just strings, right? So we could use the same approach as for commits and store them into a set.
And here we just need to check whether the size of the set is greater than 1.

Actually, if we move this check to `_generate_abis`, we don't even need this function.

scripts/generate_abi.py
bilbeyt: i have done this just to use the if condition to pass some commits if i already processed them, so maybe i can just use another set and completely forget about commit_id here

istankovic (changes requested):
I think the script can be simplified some more.

Also, for consistency, I would rename the script to `scripts/generate-abi.py`.

scripts/generate_abi.py
bilbeyt: again i guess i need it, ok :D

scripts/generate_abi.py
bilbeyt: it is a dict and i cant hash it

bilbeyt: i am returning everything ape provides as abi

bilbeyt: now i got what you mean, changed it. i guess now it is better

scripts/generate-abi.py
```diff
@@ -0,0 +1,87 @@
...
+    return merged_abis
+
+
+def _generate_abis(contract_commit_map: dict[str, set[str]]) -> dict[str, dict[str, str]]:
+    generated_abis: dict[str, dict[str, str]] = {}
+    processed_commit_ids = []
+
+    for commit_ids in contract_commit_map.values():
+        for commit_id in commit_ids:
+            if commit_id in processed_commit_ids:
```
istankovic, line 53: This is going to do a linear search so I suggest making `processed_commit_ids` a `set` to turn this into a constant-time operation.

scripts/generate-abi.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+    Popen(command, shell=True).wait()
+
+
+def _merge_abis(generated_abis: dict[str, dict[str, str]]) -> dict[str, str]:
+    merged_abis = {}
+
+    for contract_name, commit_abis in generated_abis.items():
+        if len(set(commit_abis.values())) > 1:
+            raise ValueError(f"Different ABI found for {contract_name}")
```
istankovic, line 41: This check should be moved to `_generate_abis`, which actually should return a single ABI, so there is no need for the `_merge_abis` function -- each contract would be guaranteed to only have one ABI (same for all commits).

scripts/generate-abi.py
```diff
@@ -0,0 +1,77 @@
...
+_COMPILE_COMMAND = """
+git clone --no-checkout git@github.com:beamer-bridge/beamer.git {beamer_path} &&
+cd {beamer_path} &&
+git checkout {commit_id} &&
+ape compile
+"""
+
+
+def _get_contract_commit_ids(deployment_path: Path) -> dict[str, set[str]]:
+    file_contract_commits: dict[str, set[str]] = {}
```
istankovic, line 17: Nitpick: I would just name this `contract_commits`, not sure what "file_" is supposed to mean here.

scripts/generate-abi.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def _compile_abis(commit_id: str, beamer_path: str) -> None:
+    command = _COMPILE_COMMAND.format(beamer_path=beamer_path, commit_id=commit_id).replace(
+        "\n", ""
+    )
+    Popen(command, shell=True).wait()
+
+
+def _generate_abis(contract_commit_map: dict[str, set[str]]) -> dict[str, str]:
```
istankovic, line 36: Nitpick: I think `contract_commit_map` should just be `contract_commits`, in line with how it's named elsewhere.

scripts/generate-abi.py
```diff
@@ -0,0 +1,77 @@
...
+                _compile_abis(commit_id, tmpdirname)
+                for contract_name in contract_commit_map:
+                    abi_path = Path(tmpdirname) / f"contracts/.build/{contract_name}.json"
+                    with open(abi_path, "r") as fp:
+                        contract_abi = fp.read()
+                    if contract_name not in generated_abis:
+                        generated_abis[contract_name] = contract_abi
+                    else:
+                        if len(set((contract_abi, generated_abis[contract_name]))) > 1:
+                            raise ValueError(f"Different ABI found for {contract_name}")
```
istankovic, line 54: Thinking more about this, it feels somewhat wasteful to create a set, only to query its length, so maybe
we could instead do something like

```python
abi = generated_abis.get(contract_name)                         
if abi is not None and contract_abi != abi:                     
    raise ValueError(f"Different ABI found for {contract_name}")
else:                                                           
    generated_abis[contract_name] = abi                         
```
?

istankovic:
BTW, this should be rebased on top of latest `main` so that the CI can succeed (latest main has fixes regarding the multiple event processors test).

scripts/generate-abi.py
```diff
@@ -0,0 +1,77 @@
...
+            processed_commit_ids.add(commit_id)
+    return generated_abis
+
+
+@click.command()
+@click.argument("deployment_path", type=click.Path(file_okay=False, dir_okay=True, path_type=Path))
+@click.argument("output_path", type=click.Path(file_okay=False, dir_okay=True, path_type=Path))
+@click.option("--only-abi", is_flag=True, show_default=True, default=False)
```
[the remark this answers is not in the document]
bilbeyt: you are right. it is not changing the output with is_flag=True, i will just remove them

istankovic (approved):

Issue #972, opened 2022-08-26
Implement beamer deployment and configuration commands


istankovic:
Make sure to use `apischema` to model deployment configuration and artifacts. Validation should occur immediately after reading deployment configuration and artifacts.

Tasks:  
- [x] Implement the `beamer deploy-base` command
- [x] Implement the `beamer deploy` command
- [ ] Implement the `beamer configure` command
- [x] Store transaction hashes for each contract deployment in the deployment file
- [ ] Update documentation
- [ ] Update instructions in [beamer-bridge/run-your-own-agent#](https://github.com/beamer-bridge/run-your-own-agent)

istankovic:
Tasks:
- [ ] Implement the `deploy` command
- [ ] Implement the `configure-call-chain` command
- [ ] Implement the `configure-finality-periods` command
- [ ] Update Optimism e2e test script
- [ ] Update Arbitrum e2e test script
- [ ] Update documentation
- [ ] Update instructions in [beamer-bridge/run-your-own-agent#](https://github.com/beamer-bridge/run-your-own-agent)

fredo:
for consistency, I would rename `layer1.json` to `1-ethereum.json`

GabrielBuragev:
It would be nice to have the following functionality of the `configure` script:
`Have a function/command to display the currently set fee values in the contract`

PR #1787, opened 2023-05-09
New deployment commands

istankovic:
This PR implements parts of #972, namely the `deploy-base` and `deploy` commands.
Configuration etc. is left to forthcoming PRs.

### DIFF
--- a/deployments/artifacts/goerli/420-optimism.deployment.json
+++ b/deployments/artifacts/goerli/420-optimism.deployment.json
@@ -1,40 +1,55 @@
 {
     "deployer": "0x42405d66fdA09dbDaC90Ff25fC5a4C2353f43E70",
-    "MintableToken": {
-        "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
-        "tx_hash": "0xd212e551e5dfbfade3453d390aa7c94f26c6d8ec082be091687077e3d62a6d19",
-        "address": "0x56BdfB46fEA60884E82243C67D5556aD1beF9232",
-        "deployment_block": 7426258,
-        "deployment_args": [
-            1000000000000000000
-        ]
+    "base": {
+        "chain_id": 5,
+        "OptimismL1Messenger": {
+            "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
+            "tx_hash": "0x48d65d277b0c5157fd6c2bdd2886fb60c18a574b53d17f0170d5265bdc72c3e7",
+            "address": "0xd6ED2cB5405b3869579338E1c425d6DE48906510",
+            "deployment_block": 8750482,
+            "deployment_args": [
+                "0x8e5693140eA606bcEB98761d9beB1BC87383706D"
+            ]
+        }
     },
-    "RequestManager": {
-        "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
-        "tx_hash": "0xe64cde41faa08e1bd2d865258d30451cde4a20949f01424950a32a8c138da7fd",
-        "address": "0x0A19c44943a6BF97d302a84c24E10767A4BE9CA0",
-        "deployment_block": 7426276,
-        "deployment_args": [
-            470000000000000,
-            900,
-            900,
-            900
-        ]
-    },
-    "FillManager": {
-        "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
-        "tx_hash": "0xe51553935aeabb009ecc8e5d17aa846d7d8b48820c2576c6cd577924ac9d7d59",
-        "address": "0x1D5207C35F82EF87332c05c3816e4aA34aB8052e",
-        "deployment_block": 7426288,
-        "deployment_args": [
-            "0x115373D2C9293d5ca21fB437Dee8Fc6adAFd6209"
-        ]
-    },
-    "OptimismL2Messenger": {
-        "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
-        "tx_hash": "0x2e9c0cd1358952bd99c9d4e2cd27474fa48633771629cf76b50e06b7e9693fe7",
-        "address": "0x115373D2C9293d5ca21fB437Dee8Fc6adAFd6209",
-        "deployment_block": 7426273,
-        "deployment_args": []
+    "chain": {
+        "chain_id": 420,
+        "MintableToken": {
+            "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
+            "tx_hash": "0xd212e551e5dfbfade3453d390aa7c94f26c6d8ec082be091687077e3d62a6d19",
+            "address": "0x56BdfB46fEA60884E82243C67D5556aD1beF9232",
+            "deployment_block": 7426258,
+            "deployment_args": [
+                1000000000000000000
+            ]
+        },
+        "RequestManager": {
+            "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
+            "tx_hash": "0xe64cde41faa08e1bd2d865258d30451cde4a20949f01424950a32a8c138da7fd",
+            "address": "0x0A19c44943a6BF97d302a84c24E10767A4BE9CA0",
+            "deployment_block": 7426276,
+            "deployment_args": [
+                470000000000000,
+                900,
+                900,
+                900
+            ]
+        },
+        "FillManager": {
+            "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
+            "tx_hash": "0xe51553935aeabb009ecc8e5d17aa846d7d8b48820c2576c6cd577924ac9d7d59",
+            "address": "0x1D5207C35F82EF87332c05c3816e4aA34aB8052e",
+            "deployment_block": 7426288,
+            "deployment_args": [
+                "0x115373D2C9293d5ca21fB437Dee8Fc6adAFd6209"
+            ]
+        },
+        "OptimismL2Messenger": {
+            "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
+            "tx_hash": "0x2e9c0cd1358952bd99c9d4e2cd27474fa48633771629cf76b50e06b7e9693fe7",
+            "address": "0x115373D2C9293d5ca21fB437Dee8Fc6adAFd6209",
+            "deployment_block": 7426273,
+            "deployment_args": []
+        }
     }
 }
--- a/deployments/artifacts/goerli/421613-arbitrum.deployment.json
+++ b/deployments/artifacts/goerli/421613-arbitrum.deployment.json
@@ -1,40 +1,56 @@
 {
     "deployer": "0x42405d66fdA09dbDaC90Ff25fC5a4C2353f43E70",
-    "MintableToken": {
-        "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
-        "tx_hash": "0x13d1a4aa33af3b768239cd765c0c84e839f9a193efbc2e3a6819e3b370aa5439",
-        "address": "0xf273c166586EDBe0Fad945d24080BAc5624d74d7",
-        "deployment_block": 14509871,
-        "deployment_args": [
-            1000000000000000000
-        ]
+    "base": {
+        "chain_id": 5,
+        "ArbitrumL1Messenger": {
+            "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
+            "tx_hash": "0xfb6494c570afb22527535ac3e6b1ea16893ac92c4f06a17961959c5c41dfceaa",
+            "address": "0xe991Cc19cfAaEcDb399d9adb87812505d8244224",
+            "deployment_block": 8750468,
+            "deployment_args": [
+                "0xaf4159A80B6Cc41ED517DB1c453d1Ef5C2e4dB72",
+                "0x6BEbC4925716945D46F0Ec336D5C2564F419682C"
+            ]
+        }
     },
-    "RequestManager": {
-        "beamer_commit": "0d09825c844d20ddc6d2709c5c832ea987b6fb9a",
-        "tx_hash": "0xbf9c612ac0369bffdf817d4595c1b7b53a5ed76eab2f2193b5abf774dc06578f",
-        "address": "0x30709Dbf6adB4B603D51f6b4582395114b132353",
-        "deployment_block": 14509890,
-        "deployment_args": [
-            470000000000000,
-            900,
-            900,
-            900
-        ]
-    },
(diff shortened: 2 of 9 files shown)


beamer/deploy/util.py
```diff
@@ -0,0 +1,143 @@
...
+    deployment_txhash: str
+    deployment_args: list[Any]
+    name: str
+
+
+def _transact(func: Union[ContractConstructor, ContractFunction]) -> Any:
+    return transact(func, timeout=600, poll_latency=1)
+
+
+def _load_contracts_info(contracts_path: Path) -> dict[str, tuple]:
```
bilbeyt, line 30: maybe, it is better to use built_path or something here because we would be using contracts/.build which has ape abi and __local__.json

beamer/deploy/util.py
```diff
@@ -0,0 +1,143 @@
...
+        with path.open() as fp:
+            info = json.load(fp)
+        contracts[info["contractName"]] = (
+            info["abi"],
+            info["runtimeBytecode"].get("bytecode", ""),
+        )
+    return contracts
+
+
+_CONTRACTS_PATH = Path("contracts/.build")
```
bilbeyt, line 44: same here

beamer/deploy/commands.py
```diff
@@ -0,0 +1,98 @@
...
+
+log = structlog.get_logger(__name__)
+
+
+def _margin_gas_price_strategy(web3: Web3, transaction_params: TxParams) -> Wei:
+    return Wei(int(rpc_gas_price_strategy(web3, transaction_params) * 1.3))
+
+
+def _ensure_commit_is_on_remote() -> None:
+    return
```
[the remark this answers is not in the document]
istankovic: Oops, indeed! Thanks for spotting!

beamer/deploy/commands.py
```diff
@@ -96,3 +97,101 @@ def deploy_base(
...
+    account = beamer.util.account_from_keyfile(keystore_file, password)
+    log.info("Loaded keyfile", address=account.address)
+
+    path = artifacts_dir / "base.deployment.json"
+    base_deployment = beamer.deploy.artifacts.Deployment.from_file(path)
+    log.info("Loaded base chain deployment", path=str(path), chain_id=base_deployment.chain_id)
+
+    url = rpc_info[base_deployment.chain_id]
+    base_w3 = make_web3(url, account, _margin_gas_price_strategy)
+    assert base_w3.eth.chain_id == base_deployment.chain_id
```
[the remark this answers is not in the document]
istankovic: > can we somehow add this assert whenever we access the rpc_info?

I'm not sure what you mean... could you please give more details?

beamer/deploy/commands.py
```diff
@@ -95,3 +96,101 @@ def deploy_base(
...
+        if deploy_mintable_token:
+            l2_contracts += (deploy_contract(w3, ("MintableToken", int(1e18))),)
+
+        path = artifacts_dir / f"{chain.chain_id}-{chain.name}.deployment.json"
+        beamer.deploy.artifacts.generate(path, account.address, l2_contracts)
+        log.info("Generated artifact", path=str(path))
+
+        # Update the base deployment file.
+        for contract in l1_contracts:
+            assert contract.name not in base_deployment.contracts
```
fredo, line 190: This will throw an assertion error as soon as a rollup has the same L1 messenger, like forks (i.e. Optimism, Boba, Base)

beamer/deploy/commands.py
```diff
@@ -95,3 +96,101 @@ def deploy_base(
...
+) -> None:
+    """Deploy L2 Beamer contracts on the specified chains and set up the trusted call chain."""
+    beamer.util.setup_logging(log_level="DEBUG", log_json=False)
+    _ensure_commit_is_on_remote()
+
+    rpc_info = beamer.deploy.config.load_rpc_info(rpc_file)
+    log.info("Loaded RPC file")
+
+    account = beamer.util.account_from_keyfile(keystore_file, password)
+    log.info("Loaded keyfile", address=account.address)
```
fredo, line 158: nitpick: We wanted to get rid of the notion `keyfile` and instead use `keystore (file)`. I just saw that we still have this notion in the tests, but would be nice to remove it

fredo: I just saw that we also still have named the account_from_keyfile function like that

beamer/deploy/commands.py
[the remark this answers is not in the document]
fredo: In this PR we have repeatedly the same pattern

```
chain_id = get_chain_id_from_config_file_or_whereever_from
url = rpc_info[base_deployment.chain_id]
web3 = make_web3(url, account, _margin_gas_price_strategy)
```
In the above code part, we add an assert if the RPC actually corresponds to the same chain_id as defined in the rpc file. But in most other cases we don't have this kind of check. 
When should we assert that this is correct and when not? Is there a reasoning or should we always do the check (I think it would be good).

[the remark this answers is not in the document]
fredo: I'm simply not able to see what led you to make this assert here but not in other places or whether it doesn't have a reason why

beamer/deploy/commands.py
```diff
@@ -95,3 +96,101 @@ def deploy_base(
...
+
+    resolver = base_deployment.obtain_contract(base_w3, "Resolver")
+
+    for path in chains["chain.json"]:
+        chain = beamer.deploy.config.Chain.from_file(path)
+        log.info("Loaded chain config", name=chain.name, chain_id=chain.chain_id, path=str(path))
+
+        url = rpc_info[chain.chain_id]
+        w3 = make_web3(url, account, _margin_gas_price_strategy)
+        assert w3.eth.chain_id == chain.chain_id
```
fredo, line 177: Ok I just saw, that you also have the check here.

beamer/deploy/commands.py
```diff
@@ -0,0 +1,97 @@
...
+    """Deploy resolver on the base chain."""
+    beamer.util.setup_logging(log_level="DEBUG", log_json=False)
+    _ensure_commit_is_on_remote()
+
+    rpc_info = beamer.deploy.config.load_rpc_info(rpc_file)
+    account = beamer.util.account_from_keyfile(keystore_file, password)
+    log.info("Loaded keyfile", address=account.address)
+
+    url = rpc_info[chain_id]
+    w3 = make_web3(url, account, _margin_gas_price_strategy)
```
fredo, line 92: i.e. here is no assert

beamer/deploy/commands.py
```diff
@@ -0,0 +1,97 @@
...
+import beamer.deploy.config
+import beamer.util
+from beamer.deploy.util import deploy_contract, get_commit_id
+from beamer.typing import ChainId
+from beamer.util import make_web3
+
+log = structlog.get_logger(__name__)
+
+
+def _margin_gas_price_strategy(web3: Web3, transaction_params: TxParams) -> Wei:
```
fredo, line 23: nitpick, but I'm pretty sure that we can remove this strategy as it is not applied in rollups, and for our base chains we use the type 2 transactions. We can also remove the argument in `make_web3()`

beamer/deploy/commands.py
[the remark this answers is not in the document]
istankovic: > In this PR we have repeatedly the same pattern
> 
> ```
> chain_id = get_chain_id_from_config_file_or_whereever_from
> url = rpc_info[chain_id]
> web3 = make_web3(url, account, _margin_gas_price_strategy)
> ```
> 
> In the above code part, we add an assert if the RPC actually corresponds to the same chain_id as defined in the rpc file. But in most other cases we don't have this kind of check. When should we assert that this is correct and when not? Is there a reasoning or should we always do the check (I think it would be good).

I agree that it would be good to have these checks and I tried to add them wherever I did `make_web3`. But in this case I missed it. In all other places there is such a check. Thanks for spotting!

beamer/deploy/commands.py
istankovic: Right, this is going away since we're not going to update the base deployment file, but put the L1 messenger info into the L2 deployment file.

beamer/deploy/util.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+    _transact(
+        resolver.functions.addRequestManager(
+            chain.chain_id, request_manager.address, l1_messenger.address
+        )
+    )
+    _transact(l1_messenger.functions.addCaller(resolver.address))
+    _transact(
+        request_manager.functions.addCaller(
+            resolver.w3.eth.chain_id,
```
fredo, line 134: nitpick: It would be more correct if we'd use `l1_messenger.w3.eth.chain_id` as we effectively set the L1 messenger as the caller in thus the origin chain is where the l1_messenger is (even though the resolver is on the same chain)

beamer/deploy/commands.py
```diff
@@ -95,3 +96,101 @@ def deploy_base(
     path = artifacts_dir / "base.deployment.json"
     beamer.deploy.artifacts.generate(path, account.address, (resolver,))
     log.info("Generated artifact", path=str(path))
+
+
+@click.command("deploy")
+@click.option(
+    "--rpc-file",
+    "rpc_file",
```
fredo, line 104: why do we need this here?

beamer/deploy/commands.py
istankovic: Fixed, thanks!

beamer/deploy/commands.py
istankovic: Good point. I remove this specific strategy, but left the `make_web3` function as-is because that arg is still used by the existing deployment script. We'll remove it once we remove the old deployment script.

beamer/deploy/util.py
istankovic: > nitpick: It would be more correct if we'd use `l1_messenger.w3.eth.chain_id` as we effectively set the L1 messenger as the caller in thus the origin chain is where the l1_messenger is (even though the resolver is on the same chain)

You are absolutely right. Done. :+1:

beamer/deploy/commands.py
istankovic: We don't, removed both here and for deploy-base. Thanks.

beamer/deploy/commands.py
fredo: ideally, yes, we should get completely get rid of the naming `keyfile`. 
At least for the interface to the user (options and logs) we were already consistent, but ideally yes, we completely get rid of `keyfile`

istankovic: I have update naming in the log output, but I'd rather leave the whole-repo keyfile -> keystore file conversion for a separate PR, if that's fine with you.

fredo: Yes, I would do the same.

beamer/deploy/config.py
```diff
@@ -0,0 +1,81 @@
...
+class _Token:
+    token_address: str | ChecksumAddress = field(metadata=validators(_validate_token_address))
+    transfer_limit: int = field(metadata=schema(min=0))
+    eth_in_token: float = field(metadata=schema(min=0))
+
+
+@dataclass
+class Chain:
+    name: str
+    chain_id: ChainId = field(metadata=schema(min=1))
```
fredo, line 51: we can reuse your ChainIdParam check from the next commit, to make sure, wherever we write a chain id, it's gonna be a positive integer

beamer/deploy/config.py
```diff
@@ -0,0 +1,81 @@
...
+def _validate_token_address(address: str) -> None:
+    if address != "mintable_token" and not is_checksum_address(address):
+        raise ValidationError(f"expected either 'mintable_token' or a checksum address: {address}")
+
+
+@dataclass
+class _BaseChain:
+    name: str
+    rpc: str
+    chain_id: ChainId = field(metadata=schema(min=1))
```
fredo, line 23: also here

beamer/deploy/commands.py
```diff
@@ -0,0 +1,90 @@
...
+)
+@click.password_option(
+    "--password",
+    type=str,
+    default="",
+    prompt=False,
+    help="The password needed to unlock the keystore file.",
+)
+@click.option(
+    "--artifacts-dir",
```
fredo, line 64: general question, did we ensure the change in wording happened in the whole codebase? 

`output-dir` -> `artifacts-dir`
`deployment files` -> `deployment artifacts`

beamer/deploy/commands.py
```diff
@@ -88,3 +88,96 @@ def deploy_base(
...
+        if deploy_mintable_token:
+            l2_contracts += (deploy_contract(w3, ("MintableToken", int(1e18))),)
+
+        path = artifacts_dir / f"{chain.chain_id}-{chain.name}.deployment.json"
+        beamer.deploy.artifacts.generate(path, account.address, l1_contracts, l2_contracts)
+        log.info("Generated artifact", path=str(path))
+
+        path = artifacts_dir / "base.deployment.json"
+        base_deployment.to_file(path)
+        log.info("Updated base chain deployment", path=str(path))
```
fredo, line 183: This is not needed anymore.

beamer/deploy/config.py
istankovic: You mean, when serializing? But we don't do serialization of config data. This is merely input to us.
On deserialization, the check happens automatically, via apischema.

beamer/deploy/commands.py
istankovic: We will need to update at least the agent, but I decided to leave all of that for later, we're still using the old deployment script. But yes, we need to update the relevant places to use new names.

beamer/deploy/commands.py
```diff
@@ -88,3 +88,96 @@ def deploy_base(
...
+)
+@click.option(
+    "--artifacts-dir",
+    type=click.Path(exists=True, file_okay=False, dir_okay=True, path_type=Path),
+    required=True,
+    metavar="DIR",
+    help="The directory to store contract deployment artifacts in.",
+)
+@click.option(
+    "--deploy-mintable-token",
```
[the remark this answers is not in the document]
istankovic: > Also nothing for this PR, but we should remove this option IMO, and move this functionality into its own script. Also I think the MintableToken should not be in the deployment artifact, to align with mainnet (other tokens are also not in the artifacts as DAI or USDC)

RIght, good point; not sure what the test frontend does, though, we may also need to consider that.

beamer/deploy/commands.py
istankovic: Gah, don't know how I overlooked that. Removed, thanks!

beamer/deploy/config.py
fredo: does it also check for positive numbers?

fredo: ah it does with `field(metadata=schema(min=1))`

istankovic: > ah it does with `field(metadata=schema(min=1))`

Yes, precisely.

fredo (approved):
Awesome job! 🍷

beamer/deploy/commands.py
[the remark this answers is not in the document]
fredo: If the frontend can work with DAI and USDC, then it can work with our deployed token which is not stored in the aritfacts :)

PR #1788, opened 2023-05-09
Add arm64 atomics fix script

panpilkarz:
Add `install` command to Makefile, which consolidates following:
```
yarn install
poetry install
ape compile
```

and conditionally fixes `atomics` problem on arm64 chips.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ CONTRACTS = "contracts/**/*.sol"
 DIFFLINT_PY_RE = "( M|A ) (beamer/|scripts/)"
 DIFFLINT_SOL_RE = "( M|A ) contracts/.*sol"
 IMAGE_NAME := beamer
+ARCH := $(shell uname -m)
 
 all: lint
 
@@ -82,3 +83,13 @@ clean:
 	rm -rf deployments/dist
 	find . -type d -name __pycache__ -exec rm -rf {} +
 	find . -type d -name .mypy_cache -exec rm -rf {} +
+
+install:
+	yarn install
+	poetry install
+ifeq ($(ARCH),arm64)
+	$(eval ATOMICS := $(shell poetry export | grep -E 'atomics==' | awk '{print $$1}'))
+	$(eval VENV := $(shell poetry run poetry env info -p))
+	poetry run pip uninstall atomics -y
+	poetry run pip install $(ATOMICS) --platform=universal2 --no-deps --target $(VENV)/lib/python*/site-packages
+endif
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -5,7 +5,7 @@ Requirements
 ------------
 
 - `Python <https://www.python.org>`_ 3.10
-- `poetry <https://python-poetry.org>`_
+- `poetry <https://python-poetry.org>`_ 1.4
 - `ganache <https://trufflesuite.com/ganache>`_ 7
 - `jq <https://stedolan.github.io/jq/>`_
 - `coreutils <https://formulae.brew.sh/formula/coreutils>`_
@@ -27,13 +27,9 @@ If you already have a repository, make sure the submodules are up to date::
 Testing
 -------
 
-First make sure that the node dependencies are installed::
+First install node and python dependencies::
 
-    yarn install
-
-Then, install python dependencies::
-
-    poetry install
+    make install
 
 Then enter the virtual environment::
 
@@ -49,7 +45,7 @@ Build the relayers::
 
 Start ganache::
 
-    ganache --wallet.totalAccounts 10 --chain.hardfork london --miner.blockGasLimit 12000000 \
+    npx ganache --wallet.totalAccounts 10 --chain.hardfork london --miner.blockGasLimit 12000000 \
             --wallet.mnemonic brownie --server.port 8545 --chain.chainId 1337
 
 Create a JSON keyfile corresponding to one of the accounts pre-funded by ganache::


scripts/ops/fix_arm64.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+TARGET="$(poetry env info -p)/lib/python*/site-packages/"
+
+pip install --upgrade atomics --platform=universal2 --no-deps --target $TARGET
```
istankovic, line 5: Hmmm, I feel like this should be handled by poetry.
In other words, you shouldn't need to upgrade atomics this way, but you would perhaps just need to specify `--platform` to poetry when doing the `install`. Doesn't something like that work?

istankovic: `poetry add` at least has this:
```
--platform=PLATFORM    Platforms for which the dependency must be installed.
```
so at least it looks like it's possibe to have multiple platforms supported.

panpilkarz: I couldn't find a way how to achieve that in poetry, but while doing more research, figured out a simpler command:

```
arch -arm64 pip install atomics --no-binary :all:
```

Would it be OK to just add it as a note to https://docs.beamerbridge.com/contributing.html?

Or perhaps add an install script, which would execute `poetry install` inside and conditionally execute above pip command?

istankovic: > I guess having install script wrapping up `yarn install` and `poetry install` and `ape compile` and fixing arm64 issue would be a nice shortcut.

Do `yarn` and `ape` need that kind of wrapper, though?
From what I understand, you only had issues with the `atomics` package -- everything else worked?

Personally, I'm leaning more toward a Mac-specific note in the docs, but it would also be nice to have feedback from people actually using Arm Macs. @fredo

panpilkarz: I meant following script (pseudo code):

```bash
yarn install
poetry install
if arm64 {
   arch -arm64 pip install atomics --no-binary :all:
}
poetry run ape compile
```

Then to shorten the docs with running `make install` instead of 3 separate commands.

fredo: I think here are two different features: 

One is to improve the handling of the missing arm64 build wheel in atomics, and second creating a make install command which also handles yarn and compilation of contracts.

In general I'm in favor of shorter commands to get started. However, I see that the main benefit here would be the fix for atomics.
@panpilkarz Is there a particular reason why you also added the compile command into the install target?

panpilkarz: No real reason. Happy to get rid of it.

Makefile
```diff
@@ -56,3 +57,13 @@ clean:
...
 	find . -type d -name .mypy_cache -exec rm -rf {} +
+
+install:
+	yarn install
+	poetry install
+ifeq ($(ARCH),arm64)
+	$(eval ATOMICS := $(shell poetry export | egrep 'atomics==' | awk '{print $$1}'))
+	$(eval VENV := $(shell poetry run poetry env info -p))
+	poetry run pip uninstall atomics -y
+	poetry run pip install $(ATOMICS) --platform=universal2 --no-deps --target $(VENV)/lib/python*/site-packages
```
istankovic, line 68: Is `--target` really needed? I wonder since you run pip via poetry, which is going to operate on the virtualenv.

istankovic: >  yes it is required. this `--target $(VENV)/lib/python*/site-packages` hassle is there only to meet the requirement.

I'm curious, where does pip install the package if you omit the `--target` option?

panpilkarz: it won't install it at all.

```
ERROR: Can not use any platform or abi specific options unless installing via '--target'
```

istankovic: :woozy_face:

Makefile
```diff
@@ -56,3 +57,13 @@ clean:
 	rm -rf .pytest_cache
 	find . -type d -name __pycache__ -exec rm -rf {} +
 	find . -type d -name .mypy_cache -exec rm -rf {} +
+
+install:
+	yarn install
+	poetry install
+ifeq ($(ARCH),arm64)
+	$(eval ATOMICS := $(shell poetry export | egrep 'atomics==' | awk '{print $$1}'))
+	$(eval VENV := $(shell poetry run poetry env info -p))
```
istankovic, line 66: I think you can omit one `poetry` invocation:

```
shell poetry env info -p
```
should be sufficient.

Makefile
```diff
@@ -1,8 +1,9 @@
-.PHONY: dist-exe container-image relayers all lint black format docs clean
+.PHONY: dist-exe container-image relayers all lint black format docs clean install
 
 CODE_DIRS = beamer/ scripts/
 CONTRACTS = "contracts/**/*.sol"
 IMAGE_NAME := beamer
+ARCH := $(shell arch)
```
istankovic, line 6: This results in
```
make: arch: No such file or directory
```
on my machine.

Makefile
panpilkarz: wasn't sufficient on my system.
was really struggling on that part and doubling poetry did the trick.

panpilkarz:
@istankovic to follow up your concern raised on chat:

> could you please take a look at https://github.com/python-poetry/poetry/issues/2138#issuecomment-858479917 ?

I think it sucks to use a fixed url with fixed version of the package.

In my solution, "the correct" version of package will be reinstalled, the one which poetry figured out after resolving dependencies puzzle. To achieve that, after `poetry install` is done, the code checks for atomics version installed, and reinstalls the same version from binary.

panpilkarz:
@istankovic Also stuff like this:
```
torch = [
   {markers = "sys_platform == 'linux'", url = "https://download.pytorch.org/whl/cpu/torch-1.8.1%2Bcpu-cp38-cp38-linux_x86_64.whl"},
   {markers = "sys_platform == 'darwin'", url = "https://download.pytorch.org/whl/cpu/torch-1.8.1-cp38-none-macosx_10_9_x86_64.whl"}
]
```

makes it possible to pin a version for a platform not for architecture. No need to patch for darwin users with Intel.

istankovic:
> I think it sucks to use a fixed url with fixed version of the package.

Yeah, I completely agree. Poetry still doesn't have a nice answer for this....

Makefile
istankovic: Let's just use `uname -m` since that is available everywhere.

istankovic:
Could you please squash the commits so that we end up with one commit changing only the Makefile, and another for the docs?

Makefile
```diff
@@ -60,3 +61,13 @@ clean:
 	rm -rf deployments/dist
 	find . -type d -name __pycache__ -exec rm -rf {} +
 	find . -type d -name .mypy_cache -exec rm -rf {} +
+
+install:
+	yarn install
+	poetry install
+ifeq ($(ARCH),arm64)
+	$(eval ATOMICS := $(shell poetry export | egrep 'atomics==' | awk '{print $$1}'))
```
istankovic, line 69: Nitpick: `egrep` is deprecated (not sure about the version you have on Mac, though)

```
egrep: warning: egrep is obsolescent; using grep -E
```

istankovic (approved):
Looks good!

Makefile
panpilkarz: you're right. used `grep -E` in difflint work for that reason.

PR #1789, opened 2023-05-09
Npm package

compojoom:
With this PR we can patch the artifacts and ABIs in an npm package. 

We only include the .json files and don't have an js API to access those. Users who consume the package have to include the files directly. 

The release process looks like this. 
- "main release"
we have to update the version number and create a PR. After the PR is merged circleCI will create a new release with that version number
- "dev release"
whenever someone creates a PR that has modifications to the artifacts folder and provided one is not on the main branch we create a package and set the version number to a --prerelease and we also prepend the current time to that version number.

Example. The package version is 2.0.0
if artifacts have been changed the new version pushed to npm from circleci will be 2.0.1-timestamp-0

if we push new change to the artifacts circleCI will again release a dev version with 2.0.1-newtimestamp-0


deployments/yarn.lock
```diff
@@ -0,0 +1,8 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+typescript@^5.0.4:
+  version "5.0.4"
+  resolved "https://registry.npmjs.org/typescript/-/typescript-5.0.4.tgz"
+  integrity sha512-cW9T5W9xY37cc+jfEnaUvX91foxtHkza3Nw3wkoF4sSlKn0MONdkdEndig/qPBWXNkmplh3NzayQzCiHM4/hqw==
```
istankovic, line 8: Do we really need this file in the repo?

deployments/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "publishConfig": {
+    "access": "public",
+    "registry": "https://registry.npmjs.org/"
+  },
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "abi:mainnet": "python ../scripts/generate-abi.py --only-abi ./artifacts/mainnet ./abis/mainnet",
+    "abi:goerli": "mkdir ./abis/goerli && cp ./goerli/*.* ./abis/goerli",
```
istankovic, line 14: I'm not sure how `cp ./goerli/*.* ./abis/goerli` is going to generate the ABI files.

Also, we probably want 'mkdir -p` instead of plain `mkdir` because the dir might already exist.

compojoom: Haha! 😀 It's not! But since the generate abi for goerli is broken, I'm distributing the abis that we currently have. Once we redeploy in goerli we can change this command.

deployments/yarn.lock
compojoom: Weeeeell, we started off with - let's have an API and turned 180° degrees to - let's just include .json files. I'm including the json files through the tsc binary - which transforms ts to js. We don't have any ts files right, but I use it to copy 
the "./artifacts/**/*.json" and "./abis/**/*.json" files to the dist folder. 

If we don't want to use tsc for this, then I could have a bash script that creates the dist folder and copies stuff to it.

istankovic: I mean, I was just wondering whether we could remove the auto-generated stuff from the repo -- because we don't really need it.

I think the "bash script" you mentioned could easily be just a new Makefile target where you could say e.g.

```
make npm-deployments-package
```
(or whatever name you prefer)
and it would take care to do all of that stuff.

.circleci/workflows.yml
```diff
@@ -331,16 +382,37 @@ jobs:
           command: bash docker/scripts/test-container-image.sh ${IMAGE_NAME}:sha-$(git rev-parse HEAD)
       - when:
           condition:
-             equal: [ main, <<pipeline.git.branch>> ]
+            not: 
+              equal: [ main, <<pipeline.git.branch>> ]
```
istankovic, line 386: Why this change?

.circleci/workflows.yml
```diff
@@ -331,16 +382,37 @@ jobs:
...
+        - equal: [true, <<pipeline.parameters.deployment-artifacts-updated>>]
+        - equal: [true, <<pipeline.parameters.deployment-package-json-updated>>]
+    jobs:
+      - checkout
+      - install-npm-packages:
+          requires:
+            - checkout
+      - install-python:
+          requires:
+            - checkout
```
istankovic, line 410: Do we need `install-python` job here?

GabrielBuragev:
What happens if we remove the `goerli` & `mainnet` folders (which is the plan) from the `deployments` folder at one point? We have to make sure to update the `create-package.sh` script to use the `python ../scripts/generate-abi.py` command for goerli as well. Since the issue with the commit hashes (for goerli) is still not resolved the whole release pipeline for this package will be faulty.

@istankovic can we please prioritize and address the goerli commit hash issue? Maybe do a redeployment on Goerli soon.

Besides this, maybe take a look at Manuel's comment. I have no other requests here.

Thanks for the big effort around this issue!

.circleci/workflows.yml
compojoom: Somehow fucked it up without noticing.

istankovic:
> @istankovic can we please prioritize and address the goerli commit hash issue? Maybe do a redeployment on Goerli soon.

Yeah, that would be a good idea. Please make an issue for that and make sure it is prioritized.
Depending on the timing, we might even use the new deployment commands directly for that.

compojoom:
> > > Why is in the published package a deployment file inside the abis (`abis/goerli/deployment.json`)? There is none in the mainnet folder
> 
> > 
> 
> > Because the create abi script is broken for goerli and we are basically copying the old deployments. Once we have a new correct goerli deployment it won't be there
> 
> 
> 
> This is causing me issues with typechain at the moment. But should be fixable

Ok, let's remove goerli for now?

.circleci/workflows.yml
compojoom: Ah, wait. I thought it was somewhere. This must be correct. This is the dev release. So it only happens on branches which are not main, and only if anything in the artifacts has changed.

.circleci/workflows.yml
```diff
@@ -331,16 +382,37 @@ jobs:
...
                 command: docker push -a ${IMAGE_NAME}
+                                             
 
 workflows:
+  npm-release:
+    when: 
+     condition: 
+      or: 
+        - equal: [true, <<pipeline.parameters.deployment-artifacts-updated>>]
+        - equal: [true, <<pipeline.parameters.deployment-package-json-updated>>]
```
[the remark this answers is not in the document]
compojoom: ok, seems to be valid syntax.

deployments/.gitignore
```diff
@@ -0,0 +1,3 @@
+abis/
+dist/
+*.tgz
```
istankovic, line 3: We can remove `abis` and `*.tgz` here because everything is generated inside `dist`, right?

.circleci/workflows.yml
```diff
@@ -298,6 +305,50 @@ jobs:
...
+            - run:
+                name: Publish to npm
+                command: yarn publish dist/beamer-bridge-deployments.tgz
+                working_directory: ~/beamer/deployments
+      - when:
+          condition:
+            and: 
+              - not:
+                  equal: [ main, <<pipeline.git.branch>> ]
+              - equal: [ true, <<pipeline.parameters.deployment-artifacts-updated>> ]
```
istankovic, line 337: Nitpick: you can just say
```
- <<pipeline.parameters.deployment-artifacts-updated>>
```

istankovic (approved):

istankovic:
Looks good, thanks!

.circleci/workflows.yml
compojoom: I just started changing it, but I don't think this makes it better:
image

At least for my brain this is confusing at first. I have to now convert `-` to mean `equal` and then think about what is 
`pipeline.parameters.deployment-artifacts-updated` supposed to be equal to?

istankovic: Feel free to leave as-is, I just wanted us to be consistent. It's also fine as-is. :+1:

PR #1791, opened 2023-05-10
Frontend/update dependencies

compojoom:
When working on the npm package PR we found out that the vite version we were using was not supporting dynamic imports. After updating vite, we had to update also some more dependencies and fix the failing tests.


manuelwedler (approved):
Looks good to me!

PR #1792, opened 2023-05-10
Add difflint Makefile target

panpilkarz:
Introduce a `difflint` command which runs all python linters against modified/added files only.

___
The purpose of dummy `.difflint` file added to the repo is a need to provide at least a single file to all linters. Tried hard to have a condition in the Makefile to run linters only if any file is subjected for linting (check if `$(FILES)` variable is empty) but unsuccessful. Feel free to share a suggestion on a workaround. According to my research, this one should work but it doesn't:
```
difflint:
    $(eval FILES := $(shell git status --porcelain | grep --color=never -E $(DIFFLINT_RE) | awk '{print $$NF}'))
ifneq ($(FILES),)
    mypy $(FILES) .difflint
    black --check --diff $(FILES) .difflint
    flake8 $(FILES) .difflint
    isort $(FILES) .difflint --diff --check-only
    pylint $(FILES) .difflint
endif
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
-.PHONY: dist-exe container-image relayers all lint black format docs clean
+.PHONY: dist-exe container-image relayers all lint black format docs clean difflint pydifflint soldifflint
 
 CODE_DIRS = beamer/ scripts/
 CONTRACTS = "contracts/**/*.sol"
+DIFFLINT_PY_RE = "( M|A ) (beamer/|scripts/)"
+DIFFLINT_SOL_RE = "( M|A ) contracts/.*sol"
 IMAGE_NAME := beamer
 
 all: lint
@@ -21,6 +23,25 @@ black:
 format: black
 	isort $(CODE_DIRS)
 	npx prettier --write $(CONTRACTS)
+
+difflint: pydifflint soldifflint
+
+pydifflint:
+	$(eval FILES := $(shell git status --porcelain | grep --color=never -E $(DIFFLINT_PY_RE) | awk '{print $$NF}'))
+	@if [ "$(FILES)" != "" ]; then \
+		mypy $(FILES) && \
+		black --check --diff $(FILES) && \
+		flake8 $(FILES) && \
+		isort $(FILES) --diff --check-only && \
+		pylint $(FILES); \
+	fi
+
+soldifflint:
+	$(eval FILES := $(shell git status --porcelain | grep --color=never -E $(DIFFLINT_SOL_RE) | awk '{print $$NF}'))
+	@if [ "$(FILES)" != "" ]; then \
+		npx solhint $(FILES) && \
+		npx prettier --list-different $(FILES); \
+	fi
 
 dist-exe:
 	mkdir -p dist


.difflint
```diff
@@ -0,0 +1 @@
+pass
```
fredo, line 1: could we create and remove the `.difflint` file within the make target?

fredo (approved):
Just some nitpicks, otherwise looks good!

Makefile
```diff
@@ -22,6 +23,14 @@ format: black
 	isort $(CODE_DIRS)
 	npx prettier --write $(CONTRACTS)
 
+difflint:
+	$(eval FILES := $(shell git status --porcelain | grep --color=never -E $(DIFFLINT_RE) | awk '{print $$NF}'))
```
istankovic, line 27: Why not use `git ls-files -m`?

istankovic:
> The purpose of dummy `.difflint` file added to the repo is a need to provide at least a single file to all linters. Tried hard to have a condition in the Makefile to run linters only if any file is subjected for linting (check if `$(FILES)` variable is empty) but unsuccessful. Feel free to share a suggestion on a workaround. According to my research, this one should work but it doesn't:
[...]

I have love/hate relationship with `make`. I would have fully expected your example to work.

Have you tried perhaps stripping the `FILES` variable to remove possible extra whitespace?

Makefile
panpilkarz: Took a look on that and would be nice to use it, however it shows only modified files, while added files are also needed in my opinion.

Makefile
```diff
@@ -22,6 +23,14 @@ format: black
...
+	mypy $(FILES) .difflint
+	black --check --diff $(FILES) .difflint
+	flake8 $(FILES) .difflint
+	isort $(FILES) .difflint --diff --check-only
+	pylint $(FILES) .difflint
```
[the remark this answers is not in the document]
panpilkarz: Added.

istankovic:
> > Have you tried perhaps stripping the FILES variable
> 
> @istankovic yes I did. it really suks
> 
> Running all linters on that dummy file is just a second so not really a big deal.
> 
> ```
> $ time make pydifflint
> echo "pass" > ".difflint.py"
> mypy .difflint.py
> Success: no issues found in 1 source file
> black --check --diff .difflint.py
> All done! ✨ � ✨
> 1 file would be left unchanged.
> flake8 .difflint.py
> isort .difflint.py --diff --check-only
> pylint .difflint.py
> rm -rf ".difflint.py"
> 
> real    0m1.256s
> ```

That is all nice, but the dummy files should really not be needed. Here is a solution without them:
```
pydifflint:                                                                                                           
    $(eval PY_FILES := $(shell git status --porcelain | grep --color=never -E $(DIFFLINT_PY_RE) | awk '{print $$NF}'))
    @if [ "$(PY_FILES)" != "" ]; then \                                                                               
        mypy $(PY_FILES) && \                                                                                         
        black --check --diff $(PY_FILES) && \                                                                         
        flake8 $(PY_FILES) && \                                                                                       
        isort $(PY_FILES) --diff --check-only && \                                                                    
        pylint $(PY_FILES); \                                                                                         
    fi                                                                                                                
```

istankovic (changes requested):
We should get rid of the dummy files, they are not really needed.

istankovic:
@panpilkarz could you please squash all three commits into a single one?

istankovic (approved):
Looks great, thanks!

Issue #619, opened 2022-04-29
Contract verification on Optimism, Arbitrum and Mainnet


istankovic:
Create a script to verify contracts on

- [ ] Optimism
- [ ] Mainnet
- [ ] Arbitrum

It can be called by `python scripts/verify-contracts.py <artifacts_dir> <rpc_file>`

Acceptance criteria

- [ ] Implement the script using ape-etherscan
- [ ] Deploy goerli and test it

taleldayekh:
Waiting for https://github.com/eth-brownie/brownie/issues/1591

fredo:
This should be pulled into the backlog once we are working on the new deployment infrastructure.
We need to find a solution though.
An alternative approach which turned out to be successful is verification on the explorer directly. According to @compojoom description we need to fold all imports into one file. For the current version such a document can be found [here](https://gist.github.com/fredo/01d53f36a8eede4f00a1ec23ab4ddd30). In any case, this should also be automated upon deployment.

andfletcher:
I added this issue to the Rework deployment infrastructure Epic

istankovic:
Now that we have ape, we need to consider using it for contract verification instead of rolling our own solution.
At least we should have an idea about the following:
- how hard it is to add support for contract verification on new chains in ape
- the extent of new dependencies we would be using inside the `beamer` Python package (excluding tests)
- possible minimization of dependencies
- potential changes in import times of `beamer`

bilbeyt:
I have found that we need to also implement ape plugins to make it work for new chains like polygon zkevm, so whenever we need to add new chain, we need to create ape plugin and adjust ape-etherscan.

PR #1794, opened 2023-05-10
scripts: implement contract verifier script

bilbeyt:
Closes: #619

### DIFF
--- a/scripts/verify-contracts.py
+++ b/scripts/verify-contracts.py
@@ -1,0 +1,75 @@
+from pathlib import Path
+
+import ape
+import click
+import structlog
+from eth_utils import to_checksum_address
+
+import beamer.deploy.artifacts
+import beamer.deploy.config
+from beamer.typing import URL, ChainId
+
+log = structlog.get_logger(__name__)
+
+
+def _get_contract_info(artifacts_dir: Path) -> dict[ChainId, dict[str, dict[str, str]]]:
+    contract_info: dict[ChainId, dict[str, dict[str, str]]] = {}
+    for file_path in artifacts_dir.glob("*.deployment.json"):
+        deployment = beamer.deploy.artifacts.Deployment.from_file(file_path)
+        if file_path.stem == "base.deployment":
+            ecosystem = "ethereum"
+            chain_id = deployment.base.chain_id
+            contracts = deployment.base.contracts
+        else:
+            assert deployment.chain is not None
+            ecosystem_info = file_path.stem.split(".")[0]
+            _, ecosystem = ecosystem_info.split("-", 1)
+            chain_id = deployment.chain.chain_id
+            contracts = deployment.chain.contracts
+        if chain_id not in contract_info:
+            contract_info[chain_id] = {}
+        for contract_name, contract in contracts.items():
+            contract_info[chain_id][contract_name] = {
+                "address": contract.address,
+                "ecosystem": ecosystem,
+                "network": file_path.parent.stem,
+            }
+
+    return contract_info
+
+
+def _verify_contract(
+    contract_name: str, contract_info: dict[str, str], rpc_url: URL, chain_id: ChainId
+) -> None:
+    try:
+        ape_ecosystem = getattr(ape.networks, contract_info["ecosystem"])
+        ape_network = getattr(ape_ecosystem, contract_info["network"])
+        with ape_network.use_provider(rpc_url):
+            etherscan = ape.networks.provider.network.explorer
+            assert etherscan is not None
+            getattr(ape.project, contract_name).at(contract_info["address"])
+            log.info(
+                "Verifying contract",
+                contract=contract_name,
+                chain_id=chain_id,
+                address=contract_info["address"],
+            )
+            etherscan.publish_contract(to_checksum_address(contract_info["address"]))
+    except Exception as err:
+        log.error("Failed", err=err)
+
+
+@click.command()
+@click.argument("artifacts_dir", type=click.Path(file_okay=False, dir_okay=True, path_type=Path))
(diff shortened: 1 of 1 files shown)


scripts/verify-contracts.py
```diff
@@ -0,0 +1,68 @@
...
+                "network": file_path.parent.stem,
+            }
+
+    return contract_info
+
+
+def _get_rpc_info(rpc_file: Path) -> dict[str, str]:
+    with open(rpc_file, "r") as fp:
+        data = json.load(fp)
+    return data
```
istankovic, line 37: Now that the new deployment commands PR is merged, we should use new utilities to load RPC info and artifacts.

scripts/verify-contracts.py
```diff
@@ -0,0 +1,68 @@
...
+                    getattr(ape.project, name).at(contract["address"])
+                    log.info(
+                        "Verifying contract",
+                        contract=name,
+                        chain_id=chain_id,
+                        address=contract["address"],
+                    )
+                    etherscan.publish_contract(to_checksum_address(contract["address"]))
+            except Exception as err:
+                log.error("Failed", err=err)
```
istankovic, line 64: Eh, I'm not sure why we should continue in case any exception happens....

bilbeyt: there is no marking for already verified ones but i can just manually verify remaining 1-2 contract for example at the end

scripts/verify-contracts.py
bilbeyt: done

scripts/verify-contracts.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+log = structlog.get_logger(__name__)
+
+
+def _get_contract_info(artifacts_dir: Path) -> dict[ChainId, dict[str, dict[str, str]]]:
+    contract_info: dict[ChainId, dict[str, dict[str, str]]] = {}
+    for file_path in artifacts_dir.glob("*.deployment.json"):
+        deployment = beamer.deploy.artifacts.Deployment.from_file(file_path)
+        assert deployment.chain is not None
```
istankovic, line 18: This is probably going to fail for `base.deployment.json`.

scripts/verify-contracts.py
istankovic: Hm, yes, that's a good point.

Could you please then just 
- move the entire `try..except` block to a separate function (.e.g. `_verify_contract` or something like that) to make the whole loop here more readable
- try to narrow the extent of `try..except` scope, in the sense that we probably don't need it around `ape_network.use_provider`

?

scripts/verify-contracts.py
```diff
@@ -0,0 +1,64 @@
...
+    rpc_info = beamer.deploy.config.load_rpc_info(rpc_file)
+    for chain_id, contracts in contract_info.items():
+        rpc_url = rpc_info[chain_id]
+        for name, contract in contracts.items():
+            ape_ecosystem = getattr(ape.networks, contract["ecosystem"])
+            ape_network = getattr(ape_ecosystem, contract["network"])
+            try:
+                with ape_network.use_provider(rpc_url):
+                    etherscan = ape.networks.provider.network.explorer
+                    assert etherscan is not None
```
istankovic, line 50: Nitpick: why do we call it `etherscan` instead of e.g. `explorer`?

scripts/verify-contracts.py
bilbeyt: use_provider should need to be inside try..except as there will be some (zksync) not using etherscan and assert will quit, instead i just want to continue.

scripts/verify-contracts.py
bilbeyt: because it really is etherscan dependent any other explorers like with blockscout or zksync explorer or for startnet this would return None already

scripts/verify-contracts.py
bilbeyt: yes fixed it now

scripts/verify-contracts.py
```diff
@@ -0,0 +1,74 @@
...
+    contract_info: dict[ChainId, dict[str, dict[str, str]]] = {}
+    for file_path in artifacts_dir.glob("*.deployment.json"):
+        deployment = beamer.deploy.artifacts.Deployment.from_file(file_path)
+        if file_path.stem == "base.deployment":
+            ecosystem = "ethereum"
+            chain_id = deployment.base.chain_id
+            contracts = deployment.base.contracts
+        else:
+            assert deployment.chain is not None
+            _, ecosystem = file_path.stem.split(".")[0].split("-")
```
istankovic, line 25: Please note there may be several `-` characters in the name. Which part do you care about?

bilbeyt: 1-ethereum.json -> ethereum, xxx-polygon-zkevm -> polygon-zkevm

bilbeyt: i think i should use it a little different in this manner

scripts/verify-contracts.py
```diff
@@ -0,0 +1,76 @@
...
+        deployment = beamer.deploy.artifacts.Deployment.from_file(file_path)
+        if file_path.stem == "base.deployment":
+            ecosystem = "ethereum"
+            chain_id = deployment.base.chain_id
+            contracts = deployment.base.contracts
+        else:
+            assert deployment.chain is not None
+            ecosystem_info = file_path.stem.split(".")[0]
+            ecosystem_index = ecosystem_info.index("-") + 1
+            ecosystem = ecosystem_info[ecosystem_index:]
```
istankovic, line 27: You can just use
```python
_, ecosystem = file_path.stem.split('-', 1)
```

istankovic (approved):
Looking good!

Issue #1339, opened 2023-01-11
Replace flake8, pylint and isort with ruff


istankovic:
https://github.com/charliermarsh/ruff is "An extremely fast Python linter, written in Rust."
We are already experiencing slow pylint and flake8 runs and we should try out ruff and replace what we can with it.
At the very least, it looks like we can replace flake8, pylint and isort.

panpilkarz:
Did some research and timing measurements for each linter:

```
$ make lint
time mypy beamer/ scripts/
       18.82 real        11.13 user         0.97 sys

time black --check --diff beamer/ scripts/
        5.53 real         2.43 user         0.42 sys

time flake8 beamer/ scripts/
        1.12 real         2.99 user         0.28 sys

time isort beamer/ scripts/ --diff --check-only
        0.40 real         0.16 user         0.04 sys

time pylint beamer/ scripts/
        7.06 real        20.16 user         0.71 sys
```

`mypy` is the most time consuming check. Unfortunately `ruff` is not a good replacement for `mypy`. I did a small test by removing function return type hint and `ruff` didn't report it as expected according to the [FAQ](https://beta.ruff.rs/docs/faq/#how-does-ruff-compare-to-mypy-or-pyright-or-pyre):

> It's recommended that you use Ruff in conjunction with a type checker, like Mypy

Regarding second most time consuming check `pylint`, `ruff` [is a good replacement for pylint](https://beta.ruff.rs/docs/faq/#how-does-ruff-compare-to-pylint), so we could save 7 seconds (~20% of total time) on `make lint`. 

```
$ ruff beamer/ scripts/
real    0m0.471s
user    0m0.060s
sys     0m0.052s
```

istankovic:
Thanks for doing the measurements!

Yeah, ruff is not a mypy replacement, but we can replace flake8, isort and pylint with it, which would save us ~8.5s, according to your timings, which is not a small saving.

panpilkarz:
IMO there is no strong need for removing isort & flake8 after we have`pydifflint` - what do you think @istankovic?

___

If we go for it, replacement of `flake8` is pretty straightforward, because there is conversion tool for that: https://pypi.org/project/flake8-to-ruff/

Regarding replacing `isort`, ruff output is less helpful comparing to `isort --diff  --check-only`. 
```
$ ruff beamer/
beamer/agent/chain.py:1:1: I001 [*] Import block is un-sorted or un-formatted
[...]

PR #1808, opened 2023-05-12
Add ruff linter

panpilkarz:
https://github.com/beamer-bridge/beamer/issues/1339

Add `ruff` linter.


pyproject.toml
```diff
@@ -105,3 +106,19 @@ filterwarnings = "ignore:Development network has a block height of"
...
 omit = ["beamer/tests/*"]
+
+[tool.ruff]
+exclude = [
+    "build",
+    "dist",
+    ".git",
+    ".mypy_cache",
+    "__pycache__",
+]
```
istankovic, line 117: Nitpick: I think we should sync this list with the one in `[tool.black]` section, and also remove `build` from both because it shouldn't be relevant (anymore).

pyproject.toml
```diff
@@ -22,6 +22,7 @@ atomics = "^1.0.2"
 apischema = "^0.18.0"
 xdg-base-dirs = "^6.0.0"
 typing-extensions = "^4.5.0"
+ruff = "^0.0.265"
```
istankovic, line 25: This should be added as a dev-dependency (`poetry add -D`), not as a regular dependency.

istankovic (changes requested):

istankovic (approved):
Nice, thanks!

PR #1816, opened 2023-05-15
deployments: add some basic tests for npm package

compojoom:
It happened that we had an empty abis/mainnet folder - which caused a broken npm release. The test here would prevent something like this in the future.

@istankovic - running the test out of the make `npm-package` would stop the release of pytest errors out, right?


compojoom:
I also think we should maybe change the circleCI conf:

https://github.com/beamer-bridge/beamer/blob/deployments/add_tests_for_deployment/.circleci/workflows.yml#L324

Instead of creating the package only when we have modified artifacts or abi, we maybe should run the `make npm-package` all the time - this way it would detect if something in the pipeline(generate-abi.py for example) is broken?

deployments/tests/test_dist.py
```diff
@@ -0,0 +1,34 @@
...
+
+@pytest.mark.parametrize("directory", directories)
+def test_json_files(directory):
+    assert os.path.exists(directory), f'{directory} does not exist'
+
+    json_files = [f for f in os.listdir(directory) if is_json_file(f)]
+    assert len(json_files) > 0, f'No JSON files in {directory}'
+
+    for file in json_files:
+        assert is_valid_json(os.path.join(directory, file)), f'File {file} in {directory} is not valid JSON'
```
istankovic, line 34: I'm thinking... is there a way to test ("import") the newly generated npm package before publishing?
Because then we could actually have a proper test (JS test) that imports the package and does some basic sanity checks, e.g. for existence of an address of a certain contract etc.

compojoom: hm, good idea. Do I test for example just the RequestManager? Or do we test all currently available contracts?

istankovic:
> I also think we should maybe change the circleCI conf:
> 
> https://github.com/beamer-bridge/beamer/blob/deployments/add_tests_for_deployment/.circleci/workflows.yml#L324
> 
> Instead of creating the package only when we have modified artifacts or abi, we maybe should run the `make npm-package` all the time - this way it would detect if something in the pipeline(generate-abi.py for example) is broken?

That is a very good point, yes. :+1: 
Whenever the backend is changed, we should run the "build and test" job and if the artifacts are changed, we additionally should also run the "publish" part.

deployments/tests/test_dist.py
istankovic: > hm, good idea. Do I test for example just the RequestManager? Or do we test all currently available contracts?

It depends on how deeply you want to test. Personally, I would at least make sure that all contracts have addresses and that there are at least `RequestManager` and `FillManager` there -- that would be sufficient as a smoke test, IMO.

compojoom:
I've moved the code for the npm package to a separate directory and added a step in the workflow to execute the test. 

The test generates a package, installs it and we compare whether the expected abis are there (we determine that based on the contract names in the artifacts) and we also check whether all artifacts have been copied to the artifacts folder of the package.

istankovic:
Could you please remove commit 5ab9ccf86914fdd2d6c6c7572f58049b385ec0ba from this PR? It's not necessary anymore.

istankovic:
Commit f108275cbf83dce89aec0635cde183e19bae1eef has message
```
deployments: add some basic tests for npm package
```
which is wrong, because that commit is about updating the CircleCI workflow.

istankovic:
@manuelwedler @GabrielBuragev could you also please take a look at this PR?
I am afraid I don't know enough about JS things.

istankovic:
Nitpick: the commit message of the first commit:
```
deployments: move npm packaage to separate folder

We’ve moved the code that generates the package to a separate folder and added some basic tests for whether the package generated contains the expected files.

deployments: wrong files copied
```
mentions "wrong files copied", not sure what that is about, probably a remnant of a rebase, but would be nice to remove it.

Also note the typo in `packaage` in the first line.

Aside from that, looks pretty good! (Note that I left the JS part to @GabrielBuragev and/or @manuelwedler).

compojoom:
Isn't that the way to write it? :D
image

deployments-npm-package/__tests__/test_package.ts
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * We read the artifacts directory and get read all the json artifacts, from there we extract the contracts
+ * inside of the base or chain key and return them as an array.
+ *
+ * @returns {Promise<{ [key: string]: string[] }>}
+ */
+const getContracts = async () => {
+  const contractsPerNetwork: { [key: string]: string[] } = {};
+  try {
+    const artifactItems = await fs.readdir(pathToArtifacts);
```
GabrielBuragev, line 15: can we rename `artifactItems` to `networks`?

deployments-npm-package/__tests__/test_package.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    for (const item of artifactItems) {
+      const artifactPath = path.join(pathToArtifacts, item);
```
GabrielBuragev, line 18: If you apply the above change, you would have to refactor the other places where `artifactsItems` occurs. 
We could have something like:
`for (const network of networks) { ... }`

Once you do these changes, you will see things will start making more sense in this code block :D

deployments-npm-package/__tests__/test_package.ts
```diff
@@ -0,0 +1,113 @@
...
+
+            const contracts = [...baseContracts, ...chainContracts].filter(
+              (key) => key !== 'chain_id',
+            );
+
+            if (!contractsPerNetwork[item]) {
+              contractsPerNetwork[item] = contracts;
+            } else {
+              contractsPerNetwork[item] = [...contractsPerNetwork[item], ...contracts];
+            }
```
GabrielBuragev, line 40: We could move the Set initialization here. I believe this will work better and be more understandable:
```js
const contracts = new Set([...baseContracts, ...chainContracts]);
contracts.delete('chain_id');

if (!contractsPerNetwork[item]) {
  contractsPerNetwork[item] = [...contracts];
} else {
  contractsPerNetwork[item] = [...contractsPerNetwork[item], ...contracts];
}
```

By having this, you can remove lines 51-53 below.

deployments-npm-package/__tests__/test_package.ts
```diff
@@ -0,0 +1,113 @@
...
+import fs from 'fs/promises';
+
+const pathToArtifacts = `${__dirname}/../../deployments/artifacts`;
+/**
+ * We read the artifacts directory and get read all the json artifacts, from there we extract the contracts
+ * inside of the base or chain key and return them as an array.
+ *
+ * @returns {Promise<{ [key: string]: string[] }>}
+ */
+const getContracts = async () => {
```
GabrielBuragev, line 12: `getContractsPerNetwork` as a method name here is more applicable i think.

deployments-npm-package/__tests__/test_package.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+  });
+
+  /**
+   * Test that all artifacts in the artifacts directory are copied to the package dist folder
+   */
+  test('test that artifacts are copied to the package', async () => {
+    const artifactItems = await fs.readdir(pathToArtifacts);
+
+    for (const item of artifactItems) {
```
GabrielBuragev, line 100: `const network of networks` here.

GabrielBuragev (changes requested):
I would like to see the naming inside the tests being addressed/improved. 
It was hard to understand the tests due to having not so easy to understand variable names.
I believe we can make it better by renaming/refactoring some parts.

The rest looks fine!

Issue #1796, opened 2023-05-11
Frontend: transfer completion timestamps are wrong


GabrielBuragev:
#### System information
After implementing the time measuring feature for each transfer we realized that we made a mistake when assuming that the timestamp of the `RequestCreated` event on source chain will always be lower then the timestamp of the `RequestFilled` on the destination chain. Currently we rely on the block timestamps to measure this but that is not a reliable clock source.
Every network/block proposer has a different clock delay (compared to local clock) and therefore we cannot use the block timestamp to calculate the time it took to finalize the transfer.

We need to find a way to calculate this with a local clock.

Possible simple solution is to use local clock instead to measure the time it takes from between the `RequestCreated` event (source chain) and  the `RequestFilled` event (destination chain).
In cases where the user submits a transfer and closes the website (without it being finalized) we interrupt/stop the measuring and do not proceed when he comes back as it might result in a large (false) time gap due to the usage of a local clock.

PR #1819, opened 2023-05-15
Frontend/fix transfer time

### LINKED ISSUE #1684: Share experience via twitter after bridge transfer

GabrielBuragev:
Changed the time strategy: from using `block.timestamp` to using local clock (with `Date.now()`).
You can read more on why this was necessary here: https://github.com/beamer-bridge/beamer/issues/1796

Closes #1684 & #1796

### DIFF
--- a/frontend/src/actions/transfers/transfer.ts
+++ b/frontend/src/actions/transfers/transfer.ts
@@ -167,11 +167,15 @@ export class Transfer extends MultiStepAction implements Encodable<TransferData>
     return this.listenerCleanupCallback !== undefined;
   }
 
-  get transferTime(): number | undefined {
+  get transferTimeSeconds(): number | undefined {
     if (!this._requestInformation?.timestamp || !this._requestFulfillment?.timestamp) {
       return undefined;
     }
-    return this._requestFulfillment.timestamp - this._requestInformation.timestamp;
+
+    const seconds =
+      (this._requestFulfillment.timestamp - this._requestInformation.timestamp) / 1000;
+
+    return Number(seconds.toFixed(1));
   }
 
   protected getStepMethods(provider?: IEthereumProvider): Record<string, CallableFunction> {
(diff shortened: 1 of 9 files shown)


frontend/src/actions/transfers/transfer.ts
```diff
@@ -343,10 +347,8 @@ export class Transfer extends MultiStepAction implements Encodable<TransferData>
     );
 
     try {
-      const timestamp = await Promise.race([fulfillmentPromise, expirationPromise]);
-      if (timestamp) {
-        this._requestFulfillment = new RequestFulfillment({ timestamp });
-      }
+      await Promise.race([fulfillmentPromise, expirationPromise]);
+      this._requestFulfillment = new RequestFulfillment({ timestamp: Date.now() });
```
manuelwedler, line 351: Could you chain the `new Request Fulfillment` call to the `fulfillmentPromise` with a `then`? So we only set the fulfillment if the request was really fulfilled. I think this is the safer approach for future changes.

manuelwedler (changes requested):
Nice fix, thank you! Just one little change I would like to see getting addressed.

frontend/src/actions/transfers/transfer.ts
GabrielBuragev: Good catch! ✅

compojoom (approved):
lgtm

manuelwedler (approved):
Looks like I have to approve another time since I requested the changes.

Issue #1490, opened 2023-02-15
Polygon ZKEVM e2e tests


andfletcher:
Source: https://zkevm.polygon.technology/setup-local-node/
Detailled README: https://github.com/0xPolygonHermez/zkevm-node/blob/develop/docs/running_local.md

We need to run a local instance of zkNode on top aof an L1. From the readme it looks that the setup runs a local geth client for L1.

Acceptance Criteria:
- [x] run zkNode
- [x] deploy our contracts on it
- [x] read contract addresses
- [x] be able to perform e2e test
- [x] Include in CI workflow

@istankovic please revise if something is missing in the acceptance criteria.

Issue #1747, opened 2023-04-25
Relayer for Polygon zkEVM


fredo:
as defined in #1487 the relayer needs to claim messages from and to polygon zkEVM. It is the same interface. 

```sol
interface IBridgeMessageReceiver {
    function onMessageReceived(
        address originAddress,
        uint32 originNetwork,
        bytes memory data
    ) external payable;
}
```

[source](https://github.com/0xPolygonHermez/code-examples/blob/main/zkevm-nft-bridge-example/contracts/polygonZKEVMContracts/interfaces/IBridgeMessageReceiver.sol#L8-L14)

The payload could be fetched from the events, the native bridge emits.

```sol
    event BridgeEvent(
        uint8 leafType,
        uint32 originNetwork,
        address originAddress,
        uint32 destinationNetwork,
        address destinationAddress,
        uint256 amount,
        bytes metadata,
        uint32 depositCount
    );
```

[source](https://github.com/0xPolygonHermez/code-examples/blob/main/zkevm-nft-bridge-example/contracts/polygonZKEVMContracts/PolygonZkEVMBridge.sol#L101-L110)

@GabrielBuragev please revise if information are missing.

GabrielBuragev:
From the first look of their documentation & SDK we managed to narrow down the steps/actions the relayer needs to take in order to relay messages from and to Polygon zkevm.

In general, their bridge infrastructure consists of only [1 contract](https://github.com/0xPolygonHermez/zkevm-contracts/blob/main/contracts/PolygonZkEVMBridge.sol) deployed on both Polygon ZKEVM & Ethereum Mainnet/testnet. This contract provides an easy-to-understand interface for both bridging a message (on source chain) & claiming a message (on destination chain).

What are the required actions in the relayer?
---
### Message Direction: _Polygon_ -> _Any rollup_

Here, before the relayer steps in, we assume that there is a message already being sent (to the **PolygonBridge**) by the **PolygonL2Messenger** with a specific `tx_hash` on the L2 side.

Relayer is called with the L2 `tx_hash` and needs to do the following:

1. Check the logs of the `tx_hash` and look for an emitted [BridgeEvent](https://github.com/0xPolygonHermez/code-examples/blob/main/zkevm-nft-bridge-example/contracts/polygonZKEVMContracts/PolygonZkEVMBridge.sol#L269). If there is none, obviously, throw an exception.
2. From this event the relayer should be able to extract most of the parameters it needs for claiming the message on the destination chain.
[...]

PR #1835, opened 2023-05-17
feat: add polygon-zkevm e2e test

bilbeyt:
Closes: #1490 #1747


deployments/config/local/1001-polygon-zkevm.json
```diff
@@ -0,0 +1,37 @@
...
+    "l1_messenger": [
+        "PolygonZkEVMMessenger",
+        ${native_bridge},
+        "${resolver}",
+        1
+    ],
+    "l2_messenger": [
+        "PolygonZkEVMMessenger",
+        ${native_bridge},
+        "${request_manager}",
```
istankovic, line 13: The addresses should always be strings, like the request manager example here. So you are missing quotes here and the adjustment of types in the Python part is not necessary.

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -0,0 +1,116 @@
...
+down() {
+    echo "Shutting down the end-to-end environment"
+    if [ ! -d ${POLYGON_ZKEVM} ]; then
+        configure_repo
+    fi
+    make -C ${POLYGON_ZKEVM}/test stop
+}
+
+configure_repo() {
+
```
istankovic, line 28: Nitpick: unnecessary empty line.

deployments/config/local/1001-polygon-zkevm.json
bilbeyt: the problem is coming from chain direction being integer and i got validation error as tuple only expects str

bilbeyt: that is why i have a change in config and for native bridge usage here also written in optimism-local-template.json and arbitrum-local-template.json

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -0,0 +1,116 @@
...
+    cd ${POLYGON_ZKEVM}
+    git checkout ${commit_id}
+}
+
+create_deployment_config_file () {
+    if [ ! -d ${POLYGON_ZKEVM} ]; then
+        configure_repo
+        make -C ${POLYGON_ZKEVM} build-docker
+    fi
+    ADDRESS=$(bridge_address)
```
istankovic, line 41: Would be nice to avoid the same variable name as the global one, e.g `bridge_address` could work...

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -0,0 +1,116 @@
...
+    make -C ${POLYGON_ZKEVM}/test run
+}
+
+e2e_test() {
+    l2_rpc=http://localhost:8123
+    password=""
+    relayer=${ROOT}/relayer/relayer-node18-linux-x64
+    contract_addresses="${CACHE_DIR}/addresses.json"
+    echo Copying contract addresses to $contract_addresses
+    ADDRESS=$(bridge_address)
```
istankovic, line 68: Similar here -- we already have global `ADDRESS`.

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env bash
+. "$(realpath $(dirname $0))/../scripts/common.sh"
+
+ROOT="$(get_root_dir)"
+
+# Deployer's address & private key.
+ADDRESS=0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
+PRIVKEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
```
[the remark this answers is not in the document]
bilbeyt: no it is not written in https://github.com/0xPolygonHermez/zkevm-node/blob/develop/docs/running_local.md

deployments/config/local/1001-polygon-zkevm.json
istankovic: > the problem is coming from chain direction (0, 1) being integer and i got validation error as tuple only expects str

I don't understand. Why can't we have
```
        "PolygonZkEVMMessenger",
        "${native_bridge}",
        "${request_manager}",
```
?

bilbeyt: now works

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -0,0 +1,133 @@
...
+    network="{
+        \"l1\": {
+            \"PolygonZKEvmBridge\": ${address}
+        },
+        \"l2\": {
+            \"PolygonZKEvmBridge\": ${address}
+        },
+        \"bridgeServiceUrl\": \"http://localhost:8080\"
+    }"
+    echo ${network} >  $network_file
```
istankovic, line 86: You can use the `cat <<EOF` pattern:
```
cat <<EOF > $network_file
...
EOF
```
Then you would not need to escape quotes.

scripts/deployment/main.py
```diff
@@ -134,6 +135,24 @@ def deploy_beamer(
...
+    )
+    l2_messenger = deploy_contract(
+        web3, _resolve_constructor_args(contract_args, chain.l2_messenger)
+    )
+
+    # Polygon ZkEVM chain ids for networks
+    # mainnnet: 1101
+    # goerli: 1442
+    # local: 1001
+    if chain.chain_id in [1442, 1101, 1001]:
```
istankovic, line 152: Nitpick: please use a tuple if you don't need mutability.

beamer/util.py
```diff
@@ -141,7 +141,7 @@ def make_web3(
     account: Optional[LocalAccount] = None,
     gas_price_strategy: GasPriceStrategy = rpc_gas_price_strategy,
 ) -> Web3:
-    w3 = Web3(HTTPProvider(url, request_kwargs=dict(timeout=5)))
+    w3 = Web3(HTTPProvider(url, request_kwargs=dict(timeout=20)))
```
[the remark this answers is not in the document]
bilbeyt: yes, Fred even used 60 in his goerli deployment branch

istankovic:
Looks pretty good, I am just concerned a bit regarding the timeout change.
I would like to know more about why it is required, whether we can get around it and what the effects of that change would be on our middleware, including the rate limiter.

beamer/util.py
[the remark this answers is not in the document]
bilbeyt: i have a change now regarding timeout in last commit. it should be ok imo as it is only changing timeout for deployments which has the biggest impact.

istankovic (approved):
Note: would be good if somebody also looked at the JS parts as I've only looked at non-JS parts and can only comment on that.

@bilbeyt @GabrielBuragev I think this is a beatiful PR. Thank you for separating the changes in clear, meaningful commits and making it easy to do the review -- that helped a lot.

Very happy how the changes turned out. Great job, guys.  :purple_heart:

relayer/src/common/events/polygon-zkevm/BridgeEvent.ts
```diff
@@ -0,0 +1,54 @@
+import type { Log } from "@ethersproject/providers";
+import type { BigNumber } from "ethers";
+
+import { PolygonZKEvmBridge__factory } from "../../../../types-gen/contracts";
+import { isUndefined } from "../../util";
+
+interface Result extends ReadonlyArray<unknown> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
```
compojoom, line 9: Why are we using `any` here? What could be the possible values be? boolean, bytes, string?

relayer/src/common/util.ts
```diff
@@ -0,0 +1,7 @@
+export const isUndefined = (property: unknown) => {
```
compojoom, line 1: maybe return `property is undefined` here for ts?

relayer/src/common/util.ts
```diff
@@ -0,0 +1,7 @@
...
+  return typeof property === "undefined";
+};
+
+export async function sleep(ms: number) {
```
compojoom, line 5: Promise<void>

relayer/src/common/events/polygon-zkevm/BridgeEvent.ts
GabrielBuragev: Literally anything. Whatever value ethers parses for the event parameter values.

relayer/src/common/util.ts
GabrielBuragev: it is a simple function so type inferring kicks in here. But i will add it just to be explicit.

relayer/src/common/util.ts
GabrielBuragev: 👍

relayer/src/services/polygon-zkevm.ts
```diff
@@ -0,0 +1,322 @@
...
+    console.log("Claiming PolygonZKEVM message on Ethereum L1.");
+
+    const receipt = await this.l2RpcProvider.waitForTransaction(l2TransactionHash, 1, 300000);
+
+    if (!receipt) {
+      throw new Error(`Transaction "${l2TransactionHash}" cannot be found on Polygon ZKEvm...`);
+    }
+    if (!receipt.status) {
+      throw new Error(`Transaction "${l2TransactionHash}" reverted on Polygon ZKEvm...`);
+    }
```
compojoom, line 111: This block is nearly the same as the block in finalize. How about we extract this into a separate function?

```
async function retrieveTransactionReceipt(provider: any, transactionHash: string) {
  const receipt = await provider.waitForTransaction(transactionHash, 1, 300000);

  if (!receipt) {
    throw new Error(`Transaction "${transactionHash}" cannot be found on L1.`);
  }
  if (!receipt.status) {
    throw new Error(`Transaction "${transactionHash}" reverted on L1.`);
  }

  return receipt;
}

```

relayer/src/services/polygon-zkevm.ts
```diff
@@ -0,0 +1,322 @@
...
+    if (responseJson.code && responseJson.message) {
+      throw new Error(responseJson.message);
+    }
+
+    return responseJson.deposit;
+  }
+
+  async prepare(): Promise<boolean> {
+    return true;
+  }
```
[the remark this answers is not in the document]
GabrielBuragev: This is basically the `prepare` step of the relay service. You will learn more about it when the docs arrive.

In short, a relayer program for a message relay from rollup A -> rollup B, does the following:
- Constructs the relayers RelayerA & RelayerB
- Calls `RelayerB.prepare`
- Calls `RelayerA.relayTxToL1`
- Calls `RelayerB.finalize`

In depth:
1. RelayerB has to take care of preparing the right conditions so the message can be executed on L1 and accepted by the corresponding "Inbox" component that is supposed to move the message from L1 -> L2(rollup B)
2. RelayerA, after the RelayerB prepared the right conditions, executes the message coming from rollup A on L1
3. RelayerB, after the RelayerA executed the message on L1, makes sure that the message arrived on rollup B.

So, sometimes you will see that the `prepare` & `finalize` step are empty for some of the relayer services because there is no action required for these steps.
In the case of Optimism and Boba, their relayer component takes care of claiming/finalizing the message traveling to their rollup.

relayer/src/services/polygon-zkevm.ts
```diff
@@ -0,0 +1,322 @@
...
+
+    await L1ClaimTxHash.wait();
+    console.log("Successfully relayed message to L1.");
+
+    return L1ClaimTxHash.hash;
+  }
+
+  async finalize(l1TransactionHash: string): Promise<void> {
+    console.log("\n");
+    console.log("Claiming message travelling to PolygonZKEVM");
```
compojoom, line 190: We can merge this into one console.log. Why the new line though?

relayer/src/services/polygon-zkevm.ts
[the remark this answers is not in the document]
compojoom: thanks!

relayer/src/services/polygon-zkevm.ts
```diff
@@ -0,0 +1,322 @@
...
+    }
+    console.log("Found BridgeEvent. Proceeding with the next steps...");
+
+    // 2. Fetch message data
+    const message = await this.getMessageInfo(
+      bridgeEventParameters.depositCount,
+      bridgeEventParameters.originNetwork,
+    );
+
+    // 3. Check if message was already claimed on destination network
```
compojoom, line 216: my grammar check says it is more accurate to say: `check if the message was already claimed on the destination network`

relayer/src/services/polygon-zkevm.ts
```diff
@@ -0,0 +1,322 @@
...
+    if (message.claim_tx_hash.length) {
+      console.log(`Message already relayed to L2 with transaction hash: ${message.claim_tx_hash}`);
+      return;
+    }
+
+    // 4. Wait until message is ready to be claimed
```
compojoom, line 222: // 4. wait until the message is ready to be claimed

relayer/src/services/polygon-zkevm.ts
```diff
@@ -0,0 +1,322 @@
...
+  1442: {
+    l1: {
+      PolygonZKEvmBridge: "0xF6BEEeBB578e214CA9E23B0e9683454Ff88Ed2A7",
+    },
+    l2: {
+      PolygonZKEvmBridge: "0xF6BEEeBB578e214CA9E23B0e9683454Ff88Ed2A7",
+    },
+    bridgeServiceUrl: "https://bridge-api.public.zkevm-test.net",
+  },
+};
```
compojoom, line 28: weren't we going to introduce a conf file for this?

compojoom: yes, so we are hardcoding them in a JS file instead of having them in a constants file somewhere?

GabrielBuragev: They will anyways be used only in this file, so i didn't see the point of moving them to a constants file.

compojoom (approved):
JS looks good to me.

GabrielBuragev:
Today i went back to elementary school and re-iterated some grammar lessons :P Thanks @compojoom !

PR #1837, opened 2023-05-17
frontend: Fix continuing of transfers

### LINKED ISSUE #1834: frontend: Transfer not continuable
I found a problem with the Safe integration. Transfers are not continued if the frontend is closed before both transactions (allowance + createRequest) succeeded.

For example, a user approves tokens and then closes the frontend. When it opens it again the transfer is shown as failed and the createRequest transaction cannot be sent.

Another scenario: When a transaction is created but not executed yet (for example with Safe, it needs another signature) and the frontend is closed, the transfer cannot get continued and will be shown as failed in the frontend.

The issue is that the `useContinueInterruptedTransfers` composable does not use the connected provider for continuing the transfers. Should be easily fixable.

manuelwedler:
Closes #1834

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -53,7 +53,7 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
 import type { Ref } from 'vue';
-import { computed, onMounted } from 'vue';
+import { onMounted } from 'vue';
 
 import type { Transfer } from '@/actions/transfers';
 import Footer from '@/components/Footer.vue';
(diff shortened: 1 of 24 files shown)


GabrielBuragev (changes requested):
Great job here Manu. This definitely improves the current UX by making sure the right conditions are met for continuing transfers, however i think there is still one part missing (mainly Safe related).

Consider this scenario:
- User tries to bridge using a Safe that is configured with >2 confirmations per tx
- User submits either an allowance or a transfer-request tx and then waits for his sig-mates to confirm, but closes the tab in the meanwhile
- After the tx has been confirmed (from his sig-mates), user comes back on our page just to be greeted with another transaction prompt (i tested this)

Looking at the code, i concluded that this happens due to the fact that we don't really check if a tx was already created for a specific action. Instead of checking if a tx was already submitted and waiting on that same tx to be confirmed/mined, we prompt the user to submit a new tx for the same action. 

I already started working on this but didn't have more time to finalize the solution. Pushed a "WIP" commit for you to take a look at and maybe continue the work. Let me know if you need more info on this.

manuelwedler (commented):
Looks good! Thanks for addressing this issue. I cannot approve though as this PR was opened by me.

PR #1843, opened 2023-05-19
fix: solve deployment related issues

bilbeyt:
This PR is solving following issues:

1. To support Polygon ZkEVM, messengers can accept integer in constructor. Config changed.
2. Old resolver args function cant handle int as argument because it looks for "$" in arg.

### DIFF
--- a/beamer/deploy/config.py
+++ b/beamer/deploy/config.py
@@ -49,8 +49,8 @@ class _Token:
 class Chain:
     name: str
     chain_id: ChainId = field(metadata=schema(min=1))
-    l1_messenger: str | tuple[str, ...]
-    l2_messenger: str | tuple[str, ...]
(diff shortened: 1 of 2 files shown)


beamer/deploy/config.py
```diff
@@ -49,8 +49,8 @@ class _Token:
 class Chain:
     name: str
     chain_id: ChainId = field(metadata=schema(min=1))
-    l1_messenger: str | tuple[str, ...]
-    l2_messenger: str | tuple[str, ...]
+    l1_messenger: str | list[str | int]
+    l2_messenger: str | list[str | int]
```
istankovic, line 53: Why did you change `tuple` to a `list`?

istankovic: I think this should just be `str | tuple[str | int, ...]`.

istankovic (approved):

PR #1855, opened 2023-05-24
frontend: Use new npm deployments package 

### LINKED ISSUE #963: Frontend: Adjust frontend to use deployments npm package
# Rationale
The ABI files related to the contracts that the frontend interact with are currently maintained manually and require manual updates after contract code update & deployment. We need to automate this process so that the frontend can consume a specific version of the contracts ABI easily.

### Why should this feature exist?
With this optimization we are further reducing the need of manual configuration. As we saw by now, manual configuration very often leads to errors.

# Implementation
Epic https://github.com/beamer-bridge/beamer/issues/769 focuses on improving the deployment infra.
Issue https://github.com/beamer-bridge/beamer/issues/1727 promises an easily consumable npm package for these cases.
Once issue #1727 is done, we need to adjust the frontend config generation script so that it consumes all the ABIs/deployment info from this npm package.

# Acceptance criteria
- [x] Config generation script is adjusted to consume the deployments folder as an npm package instead of via the filesystem
- [x] Replace the ABI asset imports (inside the service files) with the imports from the npm package

manuelwedler:
Closes #963 


GabrielBuragev (approved):
Looks really good!

I like how you utilized the `import` mappings inside `package.json`!
Does that mean that the frontend will use diferent ABIs for the production/development environment?
If so, we could try to integrate this approach for the relayer as well.

P.S. in order to not wait for your response, i fixed the one nitpick i had by myself. It's related to the description for one of the test cases in `readFromDirectory()`. 
I applied the following change:
```diff
it('throws an exception if the folder path is invalid', () => {
- returns an empty DeploymentInfo if folder does not exist
+ throws an exception if the folder path is invalid
      fs.readdirSync = vi.fn().mockImplementation(() => {
        throw new Error('Folder does not exist');
      });
      expect(() => DeploymentInfo.readFromDirectory('testnet/')).toThrow();
    });
```

manuelwedler:
Yes, the goerli and the mainnet deployment in the same npm package version might be using different versions of the contracts. So we should make sure to use the correct abis when using the npm package. 
What I still don't get is if we do a contract change and and publish a development version of the package, which version of the ABIs would get updated? Or how to get a version of the ABIs that was not deployed on a chain yet?

PR #1858, opened 2023-05-24
Fix flaky e2e test due to unfinalized message

### LINKED ISSUE #1805: e2e tests wait until message is finalized
We currently have flaky tests because we don't wait until a message is finalized before calling the relayer. On local tests finality period is near 0. That's why we have this race condition. 

Three options to solve this problem

1) If relayer crashes, retry
2) wait a fixed amount of time
3) wait until message is finalized and ready for execution.

I think 3) is the cleanest solution and we could even think about implementing this into the relayer. If not, as the status quo, we need to externally make sure that the relayer only runs after finalization.
Otherwise we could also implement the check into the e2e test shell script.

manuelwedler:
Closes #1805

There might be the case in the e2e tests that a message was not finalized yet and the relayer is already started. To align with the agent, the e2e tests should restart the relayer. For the implementation, I separated the execution of the relayer in a new python script.

### DIFF
--- a/docker/scripts/common.sh
+++ b/docker/scripts/common.sh
@@ -63,6 +63,26 @@ e2e_test_fill() {
     echo Fill tx hash: $e2e_test_l2_txhash
 }
 
+e2e_test_relayer() {
+    local root=$(get_root_dir)
+    local l1_rpc=$1
+    local l2_rpc=$2
+    local network_config=$3
+    local privkey=$4
+    local txhash=$5
+    local relayer=${root}/relayer/relayer-node18-linux-x64
+    
+    echo Starting relayer...
+    timeout 5m bash -c "until ${relayer} --l1-rpc-url '$l1_rpc' \
+                                         --l2-relay-to-rpc-url '$l2_rpc' \
+                                         --l2-relay-from-rpc-url '$l2_rpc' \
+                                         --network-to '$network_config' \
+                                         --network-from '$network_config' \
+                                         --wallet-private-key '$privkey' \
+                                         --l2-transaction-hash '$txhash'; \
+                        do sleep 1s; done"
+}
+
 e2e_test_verify() {
     local root="$(get_root_dir)"
     local deployment_dir=$1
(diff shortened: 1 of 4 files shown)


istankovic (changes requested):

docker/scripts/common.sh
```diff
@@ -63,6 +63,26 @@ e2e_test_fill() {
...
+    local root=$(get_root_dir)
+    local l1_rpc=$1
+    local l2_rpc=$2
+    local network_config=$3
+    local privkey=$4
+    local txhash=$5
+    local relayer=${root}/relayer/relayer-node18-linux-x64
+    
+    echo Starting relayer...
+    timeout 5m bash -c "until ${relayer} --l1-rpc-url '$l1_rpc' \
```
istankovic, line 76: I think we should omit the timeout and let CircleCI's timeout handle that. In some cases even 5min may not be enough so let's just defer that to CircleCi.

manuelwedler: Sure we can do that.
I did it because you said the opposite in the issue: https://github.com/beamer-bridge/beamer/issues/1805#issuecomment-1547705673

scripts/e2e-test-relayer.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        return completed.returncode
+
+    with Sleeper(300) as sleeper:  # todo
+        while run_relayer() != 0:
+            sleeper.sleep(0.1)
+
+
+if __name__ == "__main__":
+    main()
```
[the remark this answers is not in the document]
manuelwedler: You mean the `scripts/common.sh` I guess. Moved the retries to there.

docker/scripts/common.sh
istankovic: Yeah, I think I was not aware just how long this stuff can take when I wrote that comment. We may even have to increase the CircleCI timeout. :disappointed:

istankovic (approved):

docker/scripts/common.sh
manuelwedler: We just talked in person and decided to merge and see if we run into issues with the timeout.

PR #1870, opened 2023-05-26
Update contributing page

panpilkarz:
Fix two errors encountered while following https://docs.beamerbridge.com/contributing.html for sake of running agent on local.


istankovic (approved):
Thanks! :purple_heart:

PR #1872, opened 2023-05-26
WIP: Improve events fetching by using websockets

panpilkarz:
This PR is POC of an improve in `EventFetcher` that can significantly reduce Beamer transaction latency and Alchemy billing.

### What's new
* After agent is synced, no more `eth_getBlockByNumber` rpc calles are expected to be called by agent, thanks to the `block_cache` dictionary that is built based on blocks received by websocket
* Instead of polling for new blocks and afterwards fetching Beamer contracts events, we fetch events after a new Beamer contract event showed up on websocket.

### Implementation
* After EventMonitor did the sync, subscribe for new blocks and new events triggered by Beamer contracts in `EventMonitor` thread
* Just after subscription was established, call `self._web3.eth.get_block("latest")` to catch the events that could had been created during creating subscription
* Collect blocks received in subscription and save them in [block_cache](https://github.com/beamer-bridge/beamer/pull/1872/files#diff-5e24f82cac7fc5916c1b86818b6848ce1e2a313b46ff5e225f55c4640eda4dd0R123) which further is used whenever `web3.eth.get_block()` is called.
* Once there is a new event received in subscription, execute the current logic that fetches events within block range.
 
### Why events are not collected from websocket

Current event fetching logic relies on fetching events within given block range. That is hard to achieve while listening on websocket, because we don't know the order of coming events and don't know whenever the event is the last event of a block. 

TODO:

- [ ] Catch IO errors during ws communication
- [ ] Test coverage (in the test run a websocket server in a thread and make agent to communicate to it).

### DIFF
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -1,3 +1,5 @@
+import json
+import re
 import os
 import sys
 import threading
@@ -8,6 +10,7 @@ from typing import Callable
 
 import requests
 import structlog
+from websockets.sync.client import connect as ws_connect
 from web3 import HTTPProvider, Web3
 from web3.contract import Contract
 from web3.types import Timestamp, Wei
@@ -59,6 +62,7 @@ class EventMonitor:
         self._web3 = web3
         self._chain_id = ChainId(self._web3.eth.chain_id)
         self._contracts = contracts
+        self._contract_addresses = [c.address for c in self._contracts]
         self._deployment_block = deployment_block
         self._stop = False
         self._on_new_events = on_new_events
@@ -68,6 +72,14 @@ class EventMonitor:
         self._poll_period = poll_period
         self._confirmation_blocks = confirmation_blocks
         self._log = structlog.get_logger(type(self).__name__).bind(chain_id=self._chain_id)
+        self._ws_rpc_url = re.sub("^http", "ws", self._web3.provider.endpoint_uri)
+        self._ws_subscribe_request = json.dumps(
+            {
+                "id": 1,
+                "method": "eth_subscribe",
+                "params": ["newHeads", {"address": self._contract_addresses}],
+            }
+        )
 
         for contract in contracts:
             assert self._chain_id == contract.w3.eth.chain_id, f"Chain id mismatch for {contract}"
@@ -87,11 +99,11 @@ class EventMonitor:
         self._on_sync_done.append(event_processor.mark_sync_done)
         self._on_rpc_status_change.append(event_processor.set_rpc_working)
 
-    def _inner_fetch(self, fetcher: EventFetcher) -> list[Event]:
+    def _inner_fetch(self, fetcher: EventFetcher, block_data=None) -> list[Event]:
         events = []
         was_working = self._rpc_working
         try:
-            events.extend(fetcher.fetch())
+            events.extend(fetcher.fetch(block_data))
         except requests.exceptions.ConnectionError:
             self._rpc_working = False
         else:
@@ -108,7 +120,7 @@ class EventMonitor:
     def _thread_func(self) -> None:
         self._log.info(
             "EventMonitor started",
-            addresses=[c.address for c in self._contracts],
+            addresses=self._contract_addresses,
         )
         fetcher = EventFetcher(
             self._web3, self._contracts, self._deployment_block, self._confirmation_blocks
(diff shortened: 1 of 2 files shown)


beamer/agent/chain.py
```diff
@@ -68,6 +68,9 @@ def __init__(
         self._poll_period = poll_period
         self._confirmation_blocks = confirmation_blocks
         self._log = structlog.get_logger(type(self).__name__).bind(chain_id=self._chain_id)
+        self._fetcher = EventFetcher(
```
panpilkarz, line 71: Move to constructor to allow stopping event fetcher thread.

beamer/events.py
```diff
@@ -268,6 +274,14 @@ def __init__(
         self._event_abis = _make_topics_abi_mapping_for_contracts(contracts)
         self._confirmation_blocks = confirmation_blocks
         self._log = structlog.get_logger(type(self).__name__).bind(chain_id=self._chain_id)
+        self._log = structlog.get_logger(type(self).__name__)
+        self._ws_rpc_url = re.sub("^http", "ws", web3.provider.endpoint_uri)
+        # FIXME - add "address" filter
```
panpilkarz, line 279: At the moment we are subscribed to all logs. Should add filtering by address. 

```
addresses=[
    '0xb51722efe9cD480EC416AE455B68Ed36829ca089',   # Beamer contract
    '0x6b6AfE68b3083aE37E7FE91482aEc7bD7267226B',   # Beamer contract
]
{"id": 1, "method": "eth_subscribe", "params": ["logs", {"address": addresses}]}
```

There is an issue I don't understand. Dont receive any logs after I am subscribed to the contracts addresses.
Recognised I am receiving updates on `0xf9d2553991c938c821c3242a81f87958a947e1b5` but have no idea where does it come from.

This works:
```
addresses=[
    '0xf9d2553991c938c821c3242a81f87958a947e1b5',   # Artifical contract...
    '0xb51722efe9cD480EC416AE455B68Ed36829ca089',   # Beamer contract
    '0x6b6AfE68b3083aE37E7FE91482aEc7bD7267226B',   # Beamer contract
]
{"id": 1, "method": "eth_subscribe", "params": ["logs", {"address": addresses}]}
```

beamer/events.py
```diff
@@ -373,3 +404,37 @@ def fetch(self) -> list[Event]:
...
+                dpath.get(json.loads(message), "params/result/blockNumber"), 16
+            )
+        except KeyError:
+            self._latest_consumed_block_number = -1
+
+    def _thread_func(self):
+        while not self._ws_thread_stop:
+            self._latest_consumed_block_number = -1
+            self._log.info("Subscribe to", url=self._ws_rpc_url)
+            with ws_connect(self._ws_rpc_url) as ws:
```
panpilkarz, line 420: No error handling at the moment.

panpilkarz:
The intention was to minimise the number of changes in existing codebase and to only plugin into `def fetch()` and speed it up.

istankovic:
This PR is missing a test for the added functionality. We should test and make sure that the WebSockets connection is working as expected. I am not yet sure how the test should look like -- any suggestions?

panpilkarz:
True. Added 4th item:

[ ] Test coverage (in the test run a websocket server in a thread and make agent to communicate to it).

beamer/events.py
```diff
@@ -334,13 +334,14 @@ def _fetch_range(
...
             )
 
-    def fetch(self) -> list[Event]:
-        try:
-            block_data = self._web3.eth.get_block("latest")
-        except requests.exceptions.ConnectionError:
-            raise
-        except RequestException:
-            return []
+    def fetch(self, block_data=None) -> list[Event]:
```
istankovic, line 337: Why does `fetch` need to receive `block_data`?

panpilkarz: `block_data=None` indicates it doesn't need but it can receive it. When block data is provided, `fetch()` doesn't need to call `self._web3.eth.get_block("latest")` and go straight to fetching logs. 

`fetch()` might be called from two context:
* during the sync or http fallback
* after a block was received by websocket

In the latter, block data is available and can be provided to `fetch()` and above applies.

istankovic: > `block_data=None` indicates it doesn't need but it can receive it. When block data is provided, `fetch()` doesn't need to call `self._web3.eth.get_block("latest")` and go straight to fetching logs.
> 
> `fetch()` might be called from two context:
> 
>     * during the sync or http fallback
> 
>     * after a block was received by websocket
> 
> 
> In the latter, block data is available and can be provided to `fetch()` and above applies.

What I'm trying to understand is, why does `fetch` need to receive the entire block data?
Do you just need the "target block number"?

panpilkarz: It could be done this way but then this line must be changed what makes the change bigger:
https://github.com/beamer-bridge/beamer/blob/main/beamer/events.py#L345

I think the change follows the original implementation of the function which indeed requests a full block and has access to it. If the function was about to always need only block number, we can refactor so it calls a function that returns a block number instead of block data.

istankovic:
Let's archive this for now.

Issue #1881, opened 2023-06-01
Adapt agent to use new deployment artifacts


bilbeyt:
We need to use new deployment artifacts in agent.

Acceptance Criteria:

- [ ] Agent uses new deployment artifacts
- [ ] Any broken tests fixed

Issue #984, opened 2022-08-31
Adjust the e2e test script to new deployment structure


fredo:
with `rpc.json` and the new config and deployment jsons created in #972, the e2e test script needs to be aware of this. 

After this issue is completed the old deployment files and config files can be removed.

- [x] update `docker/scripts/common.sh` and other e2e test related scripts
- [x] remove the old deployment files
- [x] remove old config files from the repo

PR #1884, opened 2023-06-01
feat: adapt agent to use new deployment artifacts

bilbeyt:
Closes: #1881 
Closes: #984


beamer/agent/commands.py
```diff
@@ -41,10 +41,16 @@ def _sigint_handler(agent: Agent) -> None:
     "--account-password", type=str, prompt=False, help="The password needed to unlock the account."
 )
 @click.option(
-    "--deployment-dir",
+    "--artifact-dir",
```
istankovic, line 44: Nitpick: if we are to be consistent with the `beamer deploy` command, this should be `--artifacts-dir`:

beamer/health/check.py
```diff
@@ -43,7 +43,8 @@ class NotificationTypes:
 
 class HealthConfig(TypedDict):
     agent_address: str
-    deployment_dir: Path
+    artifact_dir: Path
```
istankovic, line 46: Nitpick: I think this should be called `artifacts_dir` to be more consistent, similar to above.

istankovic:
Could you please move changes to `deployments/config` out of 415e053a4b459d03b2d43a4484c71be83f22f085 into their commit(s)? 415e053a4b459d03b2d43a4484c71be83f22f085 is doing too much.

docker/arbitrum/arbitrum.sh
```diff
@@ -2,21 +2,23 @@
...
 
 # Deployer's address & private key.
-ADDRESS=0x1CEE82EEd89Bd5Be5bf2507a92a755dcF1D8e8dc
-PRIVKEY=0x3ff6c8dfd3ab60a14f2a2d4650387f71fe736b519d990073e650092faaa621fa
-
-CACHE_DIR=$(obtain_cache_dir $0)
-DEPLOYMENT_DIR="${CACHE_DIR}/deployment"
-DEPLOYMENT_CONFIG_FILE="${CACHE_DIR}/arbitrum-local.json"
+ADDRESS=0x3f1Eae7D46d88F08fc2F8ed27FCb2AB183EB2d0E
+PRIVKEY=0xb6b15c8cb491557369f3c7d2c287b053eb229daa9c22138887752191c9520659
```
istankovic, line 8: Why the change in address/privkey?

docker/arbitrum/arbitrum.sh
```diff
@@ -40,7 +40,7 @@ down() {
 wait_for_sequencer() {
     local RETRIES=180
     local i=0
-    until docker logs nitro-sequencer-1 2>&1 | grep -q "HTTP server started";
+    until docker logs nitro-sequencer-1 2>&1 | grep "HTTP server started";
```
istankovic, line 43: Why remove `-q` here?

docker/arbitrum/arbitrum.sh
```diff
@@ -52,18 +52,25 @@ wait_for_sequencer() {
...
+    git fetch --depth 1 ${repo} ${commit_id} 
+    git checkout ${commit_id}
+    git submodule update --init blockscout
 }
 
+if [ ! -d ${NITRO} ]; then
+    configure_repo
+fi
+patch_test_node_script
+
```
istankovic, line 69: I would move this entire block, together with the `configure_repo` just below `patch_test_node_script` functions, to keep this whole stuff near since it is related.

docker/arbitrum/arbitrum.sh
```diff
@@ -52,18 +52,25 @@ wait_for_sequencer() {
...
 up() {
     echo Starting the end-to-end environment
-    ${TEST_NODE_SCRIPT} --init --detach --no-blockscout
+    ${TEST_NODE_SCRIPT} --init --detach
```
istankovic, line 72: Why remove the `--no-blockscout` option?

docker/arbitrum/arbitrum.sh
```diff
@@ -77,26 +84,30 @@ create_deployment_config_file() {
...
 
 e2e_test() {
+    image_id=$(docker ps --filter "ancestor=nitro-testnode-scripts" --format "{{.ID}}")
+    if [[ ! -z $image_id ]]; then
+        docker stop $image_id
+    fi
     l2_rpc=http://0.0.0.0:8547
     password=""
-    l1_messenger=$(cat ${DEPLOYMENT_DIR}/deployment.json | jq -r '.base_chain.ArbitrumL1Messenger.address')
+    l1_messenger=$(cat ${ARTIFACT_DIR}/412346-arbitrum.deployment.json | jq -r '.base.ArbitrumL1Messenger.address')
```
istankovic, line 98: We don't really need `cat` here, right?
E.g.
```
    l1_messenger=$(jq -r '.base.ArbitrumL1Messenger.address'  ${ARTIFACT_DIR}/412346-arbitrum.deployment.json)
```
should also work, no?

docker/arbitrum/arbitrum.sh
```diff
@@ -77,26 +84,30 @@ create_deployment_config_file() {
...
                    -c "cat localNetwork.json" > $network_config
 
-    e2e_test_fill ${DEPLOYMENT_DIR} ${KEYFILE} "${password}" $l2_rpc
+    e2e_test_fill $ARTIFACT_DIR $KEYFILE "${password}" $l2_rpc $ABI_DIR
 
     export ARBITRUM_L1_MESSENGER=$l1_messenger
 
     e2e_test_relayer http://0.0.0.0:8545 $l2_rpc $network_config $PRIVKEY $e2e_test_l2_txhash
-    e2e_test_verify ${DEPLOYMENT_DIR} $l2_rpc $ADDRESS $e2e_test_request_id
+    e2e_test_verify $ARTIFACT_DIR $l2_rpc $ADDRESS $e2e_test_request_id $ABI_DIR
```
istankovic, line 110: I notice that `e2e_test_fill` and `e2e_test_verify` share some of the arguments, however, they are in different positions. Could we reposition the arguments, so that the shared ones come first, in the same positions.
Something like
```
e2e_test_fill $ARTIFACT_DIR $ABI_DIR $l2_rpc $KEYFILE "${password}" 
e2e_test_verify $ARTIFACT_DIR $ABI_DIR $l2_rpc $ADDRESS $e2e_test_request_id
```
?

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -8,9 +8,12 @@ ADDRESS=0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
 PRIVKEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
 
 CACHE_DIR=$(obtain_cache_dir $0)
+ABI_DIR="${ROOT}/contracts/.build"
+RPC_FILE="${ROOT}/deployments/config/local/rpc.json"
```
istankovic, line 12: Since `RPC_FILE` does not depend on the cache dir and is the same for all e2e test scripts, it would make sense to remove this variable, remove the corresponding argument to `deploy_beamer` function and instead make `deploy_beamer` from `common.sh` deteremine the RPC file location.

docker/optimism/optimism.sh
```diff
@@ -8,8 +8,10 @@ ADDRESS=0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
 PRIVKEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
 
 CACHE_DIR=$(obtain_cache_dir $0)
-DEPLOYMENT_DIR="${CACHE_DIR}/deployment"
-DEPLOYMENT_CONFIG_FILE="${CACHE_DIR}/optimism-local.json"
+ARTIFACT_DIR="${CACHE_DIR}/deployments/artifacts/local"
+ABI_DIR="${ROOT}/contracts/.build"
```
istankovic, line 12: Similarly as for `RPC_FILE`, we could move this to `common.sh`, `deploy_beamer` function.

scripts/call_contracts.py
```diff
@@ -27,11 +27,18 @@ def new_func(ctx: Any, *args: Any, **kwargs: Any) -> Callable:
 
 @click.group("cli")
 @click.option(
-    "--deployment-dir",
+    "--artifact-dir",
```
istankovic, line 30: I think this should be `--artifacts-dir` to be consistent with other commands.

scripts/call_contracts.py
```diff
@@ -27,11 +27,18 @@ def new_func(ctx: Any, *args: Any, **kwargs: Any) -> Callable:
...
     metavar="DIR",
-    help="The directory that stores contract deployment files.",
+    help="The directory that stores deployment artifact files.",
+)
+@click.option(
+    "--abi-dir",
+    type=Path,
+    required=True,
+    metavar="DIR",
+    help="The directory that stores contract abi files.",
```
istankovic, line 41: Nitpick: the `read` command has
```
Path to the directory with contract ABIs.
```
as the help text and we should be consistent.

docs/source/running.rst
```diff
@@ -39,24 +39,22 @@ During the Beamer contracts' deployment process, a directory with deployment inf
...
-
-The above shows contract `deployment on Goerli`_.
-
-The ``deployment.json`` file contains information on the chains that the
+	└── artifacts/
+		├── goerli/
+      ├── 5-ethereum.deployment.json
+      ├── 420-optimism.deployment.json
+      ├── 421613-arbitrum.deployment.json
+      └── 84531-base.deployment.json
```
istankovic, line 47: This does not look aligned well.

docker/arbitrum/arbitrum.sh
bilbeyt: i can change it back

docker/arbitrum/arbitrum.sh
bilbeyt: with q it is not working ok in macos actually it cant find the log with -q option.

docker/arbitrum/arbitrum.sh
compojoom: because there is no --no-blockscout option anymore. By default blockscout is not necessary, you can now opt in to turn it on.

bilbeyt: no option like this now

docker/arbitrum/arbitrum.sh
istankovic: Yeah, that would also be my preffered course of action.

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -81,16 +79,9 @@ e2e_test() {
...
-               --l2-relay-to-rpc-url $l2_rpc \
-               --l2-relay-from-rpc-url $l2_rpc \
-               --network-to $network_file \
-               --network-from $network_file \
-               --wallet-private-key $PRIVKEY \
-               --l2-transaction-hash $e2e_test_l2_txhash
-    e2e_test_verify ${DEPLOYMENT_DIR} $l2_rpc $ADDRESS $e2e_test_request_id
+    e2e_test_fill $ARTIFACT_DIR $l2_rpc $KEYFILE "${password}"
+    e2e_test_relayer http://localhost:8545 $l2_rpc $network_file $PRIVKEY $e2e_test_l2_txhash
+    e2e_test_verify $ARTIFACT_DIR $l2_rpc $ADDRESS $e2e_test_request_id
```
istankovic, line 84: Nice!

docker/scripts/test-container-image.sh
```diff
@@ -1,11 +1,10 @@
 #!/usr/bin/env bash
 . "$(realpath $(dirname $0))/common.sh"
 
-ROOT="$(get_root_dir)"
-
 CACHE_DIR=$(obtain_cache_dir "$0")
-DEPLOYMENT_DIR="${CACHE_DIR}/deployment"
-DEPLOYMENT_CONFIG_FILE="${ROOT}/scripts/deployment/ethereum-local.json"
+
+ARTIFACTS_DIR="${CACHE_DIR}/deployments/artifacts/local"
```
istankovic, line 6: Nitpick: here we're using `ARTIFACTS_DIR`, while in other scripts we use `ARTIFACT_DIR`.
I think we should use `ARTIFACTS_DIR` also in other scripts, and then we would be consistent everywhere (also with commands).

docs/source/configuration.rst
```diff
@@ -325,9 +333,15 @@ Options reference
...
+        artifacts-dir = PATH
+
+     - Path to the deployment artifact directory as it can be seen `here <https://github.com/beamer-bridge/beamer/tree/main/deployments/artifacts>`_.
+
+   * - ::
+
+        abi-dir = PATH
 
-     - Path to the deployment directory as it can be seen `here <https://github.com/beamer-bridge/beamer/tree/07d66e0bb8c76bb1ff219e24e34e1c24ee7890c6/deployments>`_.
+     - Path to the contract abi files directory.
```
istankovic, line 344: This is not consistent within the docs, but also not with the change that was made in `call_contracts` script (the help text). I think we should use the same wording both in the docs and in the script help text i.e. we should make this match the wording from the `call_contracts` script help text.

scripts/generate-contract-addresses-doc-page.py
```diff
@@ -29,22 +30,40 @@ def main() -> None:
...
-            contracts.update(base_contracts)
+    contract_info: dict[ChainId, dict[str, ChecksumAddress]] = {}
+    for artifact_path in artifacts_dir.glob("*.deployment.json"):
+        deployment = Deployment.from_file(artifact_path)
+        base_chain_id = deployment.base.chain_id
+        if base_chain_id not in contract_info:
+            contract_info[base_chain_id] = {}
+        file_chain_id = str(base_chain_id)
+        if not artifact_path.stem == "base.deployment":
+            file_chain_id, _ = artifact_path.stem.split(".")[0].split("-")
```
istankovic, line 41: We should rather use information from the `deployment` instance, instead of parsing this from the filename.

istankovic: Also, note that you can distinguish between a base deployment and regular chain deployment by checking
`deployment.chain is not None` (for base deployment, it will be `None`).

scripts/generate-contract-addresses-doc-page.py
```diff
@@ -29,22 +30,40 @@ def main() -> None:
...
+
+        for contract_name, contract in deployment.base.contracts.items():
+            contract_info[base_chain_id][f"{contract_name}-{file_chain_id}"] = contract.address
+
+        if deployment.chain is None:
+            continue
```
istankovic, line 47: ... which we already have here. :-)

scripts/generate-contract-addresses-doc-page.py
```diff
@@ -29,22 +30,40 @@ def main() -> None:
...
+
+        chain_id = deployment.chain.chain_id
+        if chain_id not in contract_info:
+            contract_info[chain_id] = {}
+        for contract_name, contract in deployment.chain.contracts.items():
+            contract_info[chain_id][f"{contract_name}-{chain_id}"] = contract.address
```
istankovic, line 53: This code is basically duplicated code above -- can we remove this duplication?

docs/source/running.rst
```diff
@@ -38,25 +38,23 @@ The contracts' deployment information
...
-		├── OptimismL2Messenger.json
-		├── RequestManager.json
-		└── Resolver.json
+  deployments/
+  └── artifacts/
+    └── goerli/
+      ├── 5-ethereum.deployment.json
+      ├── 420-optimism.deployment.json
+      ├── 421613-arbitrum.deployment.json
+      └── 84531-base.deployment.json
```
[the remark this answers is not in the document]
istankovic: Ah, but then the chain name part should be `goerli`, not `base`.
Same as we have
```
10-optimism.deployment.json
1101-Polygon ZkEVM.deployment.json'
1-ethereum.deployment.json
42161-arbitrum.deployment.json
base.deployment.json
```
for mainnet. (Note both `base` and `1-ethereum`.)

scripts/generate-contract-addresses-doc-page.py
```diff
@@ -1,25 +1,38 @@
...
+    1101: "https://zkevm.polygonscan.com/address/{address}",
 }
 
+_CONTRACT_INFO: dict[ChainId, dict[tuple[str, ChainId], ChecksumAddress]] = {}
+
+
+def process_contracts_for_chain(
+    chain_id: ChainId, contracts: dict[str, DeployedContractInfo]
+) -> None:
+    global _CONTRACT_INFO
```
istankovic, line 27: You don't need the `global` statement here.

scripts/generate-contract-addresses-doc-page.py
```diff
@@ -29,22 +42,29 @@ def main() -> None:
...
+        base_chain_id = deployment.base.chain_id
+        process_contracts_for_chain(base_chain_id, deployment.base.contracts)
+
+        if deployment.chain is None:
+            continue
+
+        chain_id = deployment.chain.chain_id
+        process_contracts_for_chain(chain_id, deployment.chain.contracts)
+
+    for chain_id, contracts in _CONTRACT_INFO.items():
```
istankovic, line 56: Much nicer! :heart:

Issue #1831, opened 2023-05-16
Add a new command to read contract configuration


istankovic:
## Reading configuration

`beamer config read --rpc-file RPC-FILE --artifact CHAIN-NAME.deployment.json CHAIN-NAME.state`

The `read` command takes an existing artifact file and a path to a (possibly non-existing)
state file. Starting from the deployment block (in case state file is not there) or from
the block specified in the state file, it then fetches all events and updates the
state. Once all contract events up to the latest block are processed, the state,
the block number and the checksum are written to the state file.

Note: updating an existing state file should be done in an atomic manner, that is,
first the command writes the new state to a temporary file and then atomically renames
the temporary file to the existing file.

If the specified state file exists, upon reading it, the command must verify the data
checksum, as described below.

### Checking consistency

Each state file has a checksum that is computed from the JSON data:

```json
{
    "block": 12345,
    "chain_id": 10,
    "checksum": "123abcf231...",
    "token_addresses": {
        "USDC": "0x1234",
        "DAI": "0x4321"
    }
    "RequestManager": {
        "min_fee_ppm": 300000,
        "lp_fee_ppm": 3000,
        "protocol_fee_ppm": 0
        "tokens": {
            "USDC": {
                "transfer_limit": 10000,
                "eth_in_token": 1800
            },
            "DAI": {
                "transfer_limit": 10000,
                "eth_in_token": 1800
            }
        },
        "whitelist": [
            "0x1",
            "0x2",
            "0x3" 
        ]
    },
    "FillManager": {
        "whitelist": [
            "0x1",
            "0x2",
            "0x3" 
        ]
    }
}
```
The checksum is computed by taking the SHA-256 of the serialized JSON content. In the above example, that would be everything without the `checksum` line.

As an illustration, given a dictionary `data` containing contract configuration as well as `block_number`,
one could use the following Python code to generate the checksum:

```python
serialized = json.dumps(data, indent=4)
serialized += '\n'
checksum = hashlib.sha256(serialized.encode('utf-8')).digest().hex()
```

Adding a newline to the end of the JSON data allows us to easily compute the checksum
without having to use a special script:
```bash
$ cat 1-ethereum.state | grep -v checksum | sha256sum
```

## Tasks

- [x] add the new `beamer config read` command
- [x] write a test for reading config from the deployment block
- [x] write a test for reading config from the block specified in the state file
- [x] write documentation

fredo:
It's fascinating with how little code you were able to write this command ❤️

PR #1886, opened 2023-06-01
Add a new command, 'config read'

istankovic:
This PR does a few preparatory steps first (add a way to skip commit checks when deploying, add address field to events) and then adds the implementation of the new `config read` command. Finally, several tests are made that exercise 
- reading state from contract deployment
- reading state starting from an already existing state
- reading state that has a mismatched checksum

There is one `TODO` item in the test code, namely, in the test for checksum mismatch. We can't test this properly now due to a presumed `apischema` bug. That will be investigated independently of this PR and once a fix is found, we will update the code marked by `TODO`.

Closes: #1831

### DIFF
--- a/beamer/config/commands.py
+++ b/beamer/config/commands.py
@@ -1,0 +1,131 @@
+from pathlib import Path
+
+import click
+import structlog
+from web3 import Web3
+
+import beamer.contracts
+import beamer.deploy.config
+import beamer.util
+from beamer.config.state import ChainConfig, Configuration, TokenConfig
+from beamer.deploy.artifacts import Deployment
+from beamer.events import (
+    ChainUpdated,
+    Event,
+    EventFetcher,
+    FeesUpdated,
+    LpAdded,
+    LpRemoved,
+    TokenUpdated,
+)
+from beamer.util import get_ERC20_abi, make_web3
+
+log = structlog.get_logger(__name__)
+
+
+def _replay_event(w3: Web3, deployment: Deployment, config: Configuration, event: Event) -> None:
+    assert deployment.chain is not None
+    match event:
+        case ChainUpdated():
+            config.request_manager.chains[event.chain_id] = ChainConfig(
+                finality_period=event.finality_period,
+                target_weight_ppm=event.target_weight_ppm,
+                transfer_cost=event.transfer_cost,
+            )
+
+        case FeesUpdated():
+            config.request_manager.lp_fee_ppm = event.lp_fee_ppm
+            config.request_manager.min_fee_ppm = event.min_fee_ppm
+            config.request_manager.protocol_fee_ppm = event.protocol_fee_ppm
+
+        case TokenUpdated():
+            token = w3.eth.contract(address=event.token_address, abi=get_ERC20_abi())
+            symbol = token.functions.symbol().call()
+            config.request_manager.tokens[symbol] = TokenConfig(
+                transfer_limit=event.transfer_limit, eth_in_token=event.eth_in_token
+            )
+            address = config.token_addresses.get(symbol)
+            if address is None:
+                config.token_addresses[symbol] = event.token_address
+            else:
+                assert address == event.token_address
+
+        case LpAdded():
+            if event.event_address == deployment.chain.contracts["RequestManager"].address:
+                config.request_manager.whitelist.add(event.lp)
+            elif event.event_address == deployment.chain.contracts["FillManager"].address:
+                config.fill_manager.whitelist.add(event.lp)
+            else:
+                raise ValueError(f"event from an unexpected address: {event}")
+
+        case LpRemoved():
+            if event.event_address == deployment.chain.contracts["RequestManager"].address:
+                config.request_manager.whitelist.remove(event.lp)
+            elif event.event_address == deployment.chain.contracts["FillManager"].address:
+                config.fill_manager.whitelist.remove(event.lp)
+            else:
+                raise ValueError(f"event from an unexpected address: {event}")
+
+
+@click.group()
+def config() -> None:
+    pass
+
+
+@config.command("read")
+@click.option(
+    "--rpc-file",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, path_type=Path),
+    required=True,
+    help="Path to the RPC config file.",
+)
+@click.option(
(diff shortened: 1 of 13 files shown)


panpilkarz (commented):
lgtm

beamer/config/commands.py
```diff
@@ -0,0 +1,133 @@
...
+    fetcher = EventFetcher(
+        w3, (request_manager, fill_manager), start_block=start_block, confirmation_blocks=0
+    )
+
+    if state_path.exists():
+        config = Configuration.from_file(state_path)
+    else:
+        config = Configuration.initial(chain_id, start_block)
+
+    events = fetcher.fetch()
```
fredo, line 128: With this call we cannot be really sure that the `synced_block` is actually latest. I know this is a bit tricky, as while replaying the events the blockchain progresses. 
The fetch() can exit without progressing to the latest block.

The code is good as is, but it can happen that `fetch()` exited early. Then the synced block can be actually far away from the actual latest state. 

One scenario could be that at the very top a `RequestException` is raised upon fetching the latest block data. 

The output would be an empty state file. This is noted with the block number. However the block number doesn't really tell me much when seeing it. Even further, if we were to automize things then it could become quite critical. 

I don't have a "good" solution for now, but maybe adding the timestamp of the block in a readable format could help humans at least.

Regarding automization, it actually has to be handled in the `write` command that the state is not up to date.

fredo:
It's fascinating with how little code you were able to write this command ❤️

beamer/config/commands.py
```diff
@@ -0,0 +1,133 @@
...
+    url = rpc_info[chain_id]
+    w3 = make_web3(url)
+    assert w3.eth.chain_id == chain_id
+    log.info("Connected to RPC", url=url)
+
+    request_manager = deployment.obtain_contract(w3, "chain", "RequestManager")
+    fill_manager = deployment.obtain_contract(w3, "chain", "FillManager")
+    # Here we assume that the configuration has not changed between the two
+    # deployment blocks.
+    start_block = max(
```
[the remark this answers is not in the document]
istankovic: > is it a problem to take `min(..)` and just get rid of the assumption?

Ah, good idea, will change. :+1:

beamer/config/commands.py
istankovic: > Regarding automization, I assume it should actually be handled by the `write` command, noticing that the state is not uptodate and then fetch the latest state. However, there the write command need to be more time sensitive about the synced block then.

That is precisely my thinking. The read side can only do so much, but given that there will always be some time between you read from chain and write to chain, it is reasonable to move the responsibility to the write side. We have to check anyway if there were configuration changes since the last state file was written. If there were, the write command should just exit with an error and advise you to run the read command again. In a way, we don't particularly care if there are 200 blocks between the block from the state file and the current latest block, if there were no state changes in between.

Theoretically, it could happen that the block from the file is too far in the past so we should not waste time fetching events up to the latest block, only to exit with an error. What we could do is check whether the last synced block is older than e.g. 1 day and if yes, just exit with an error and require a new sync via the read command.

beamer/config/state.py
```diff
@@ -0,0 +1,135 @@
...
+
+@dataclass(frozen=True)
+class TokenConfig:
+    transfer_limit: int = field(metadata=schema(min=0))
+    eth_in_token: int = field(metadata=schema(min=0))
+
+
+@dataclass(frozen=True)
+class ChainConfig:
+    finality_period: int = field(metadata=schema(min=0))
```
fredo, line 33: finality period must be greater than 0

beamer/config/state.py
```diff
@@ -0,0 +1,135 @@
...
+    target_weight_ppm: int = field(metadata=schema(min=0))
```
fredo, line 34: is there also a `max` validator? Must be lower than 1000000

beamer/config/state.py
```diff
@@ -0,0 +1,135 @@
...
+
+@dataclass
+class FillManagerConfig:
+    whitelist: set[ChecksumAddress]
+
+
+@dataclass
+class RequestManagerConfig:
+    min_fee_ppm: int = field(metadata=schema(min=0))
+    lp_fee_ppm: int = field(metadata=schema(min=0))
```
fredo, line 46: is there also a max validator? Must be lower than 1000000

beamer/config/state.py
```diff
@@ -0,0 +1,135 @@
...
+    protocol_fee_ppm: int = field(metadata=schema(min=0))
```
fredo, line 47: is there also a max validator? Must be lower than 1000000

beamer/config/state.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+@dataclass
+class FillManagerConfig:
+    whitelist: set[ChecksumAddress]
+
+
+@dataclass
+class RequestManagerConfig:
+    min_fee_ppm: int = field(metadata=schema(min=0, max=999_999))
```
fredo, line 45: can be larger than 999999

fredo: the reason for it is that is could be a more than 100% fee applied if transfer cost were very low.

istankovic: Should we have a "reasonable" upper bound here or shall I just remove the `max` completely?

fredo: would just remove it completely, so that it basically aligns with the contracts

docs/source/commands.rst
```diff
@@ -67,6 +68,34 @@ The ``agent`` command will run a Beamer agent and provide liquidity for the brid
...
+   * - ``--abi-dir DIR``
+     - The directory containing contract ABI files.
+
+   * - ``--artifact CHAIN-NAME.deployment.json``
+     - Path to the deployment artifact.
+
+   * - ``--rpc-file``
+     - Path to the JSON file containing RPC information.
+
+
```
fredo, line 98: It would be nice to see an example output.

fredo (approved):
Looks very good!

docs/source/commands.rst
istankovic: You mean, the example for an RPC file?
If so, I was meaning to have a separate subsection on that and link to it from here and also from the deployment commands, when I get to their docs.

PR #1887, opened 2023-06-01
CI: Add CI configurations to publish frontend build to beamer-bridge/frontend-deployment repo

GabrielBuragev:


.circleci/workflows.yml
```diff
@@ -297,9 +301,37 @@ jobs:
...
+    working_directory: ~/beamer/frontend
+    environment:
+      REPO_URL: git@github.com:beamer-bridge/frontend-deployment.git
+      TMP_DIR: /tmp/frontend-deployment
+    steps:
+      - add_ssh_keys:
+          fingerprints:
+            - fa:3a:6e:26:54:04:d8:58:3f:63:7c:3d:00:a1:d7:c2:be
+      - run: |
+          mkdir -p ~/.ssh
```
panpilkarz, line 315: I think `-p` does nothing here and can be removed. Would make sense if there were any subdirectories in `~/.ssh`

manuelwedler: Had a failing run without the option

manuelwedler:
Should be ready now

GabrielBuragev:
Note to myself:
After merging this, migrate old mainnet & testnet DO Apps to new ones (including domain pointers)

.circleci/workflows.yml
```diff
@@ -468,10 +468,9 @@ jobs:
 workflows:
   npm-release:
     when: 
-     condition: 
       or: 
-        - <<pipeline.parameters.deployment-artifacts-updated>>
-        - <<pipeline.parameters.deployment-package-json-updated>>
+        - equal: [ true, <<pipeline.parameters.deployment-artifacts-updated>> ]
+        - equal: [ true, <<pipeline.parameters.deployment-package-json-updated>> ]
```
istankovic, line 473: Nitpick: why is `equal` necessary here?
Isn't this the same as the expression it replaces? At least that is my impression from reading https://circleci.com/docs/configuration-reference/#logic-statements

istankovic (approved):
Beautiful work, @manuelwedler!

PR #1888, opened 2023-06-02
fix: change event fetching algorithm

bilbeyt:


beamer/tests/agent/test_request.py
```diff
@@ -363,20 +363,6 @@ def test_agent_only_claim_once_after_restart(request_manager, token, agent, dire
                     break
 
 
-def test_event_fetcher_progress_without_events(agent, direction):
```
fredo, line 366: this test should still pass IMO

beamer/tests/agent/test_request.py
```diff
@@ -363,20 +363,6 @@ def test_agent_only_claim_once_after_restart(request_manager, token, agent, dire
...
-    with Sleeper(5) as sleeper:
-        while ape.chain.chain_id not in agent.get_context(direction).latest_blocks:
-            sleeper.sleep(0.1)
-    for _ in range(5):
-        ape.chain.mine(1)
-        with Sleeper(5) as sleeper:
-            while (
-                agent.get_context(direction).latest_blocks[ape.chain.chain_id].number
-                != ape.chain.blocks[-1].number
```
fredo, line 375: What does `[-1]` mean in `ape.chain.blocks[-1].number`

beamer/tests/agent/test_request.py
bilbeyt: actually this test is normally looking for update block number without any new events, and now block number shouldnt increase normally

beamer/tests/agent/test_request.py
bilbeyt: latest block number in this case

bilbeyt: i mean this is `ape.chain.blocks[-1]` latest block

fredo: Ah ok so -1 does not mean `latest - 1`?

beamer/tests/agent/test_request.py
fredo: it should, Even though there aren't new events, new blocks should be updated in fetch()

beamer/tests/agent/test_request.py
bilbeyt: no directly the latest like in python list

fredo: got it, just wanted to be crystal clear. thanks

beamer/events.py
```diff
@@ -355,7 +355,7 @@ def fetch(self) -> list[Event]:
             events = self._fetch_range(from_block, to_block)
             if events is not None:
                 result.extend(events)
-            from_block = BlockNumber(to_block + 1)
+                from_block = BlockNumber(to_block + 1)
```
istankovic, line 358: Can we have a test for this, please?

istankovic (approved):

PR #1902, opened 2023-06-05
Relayer: CLI extension & OP Network update

### LINKED ISSUE #1897: Relayer OP prove() call
The relayer must implement another functionality. That is, to be able to `prove()` a message travelling from OP. 
With the current relayer design (CLI command) this cannot be easily integrated in such a way where we keep a good/clean CLI design (successful return codes only when a message has been fully relayed to its destination). What we need to do is extend our typescript (relayer) program to include separate subcommands for functionalities like this (different than the `relay` functionality).

## Command impl. specs
The intent of this new subcommand is to take care of preparing the message that needs to travel from OP by proving it on L2 so it can then be used on L1 after it reaches the `finality_period`. 
Why we want to split this into a separate subcommand? - Because its a completely separate functionality that either needs to return a successful status code or an exception depending on the status of this specific process and not mix up with different sub-processes.
  
This program (subcommand) must make sure that the `prove()` transaction was included in a block and executed successfully. It should only return with a successful exit code after seeing the successful transaction (and potentially an event).

# Acceptance Criteria
- [ ] Relayer is able to execute a prove transaction for OP Bedrock
- [ ] Relayer exposes a separate CLI command to call the execution of a prove transaction
- [ ] E2E test works

GabrielBuragev:
closes #1897


docker/scripts/common.sh
```diff
@@ -83,6 +84,25 @@ e2e_test_relayer() {
...
+    local root=$(get_root_dir)
+    local l1_rpc=$1
+    local l2_rpc=$2
+    local network_config=$3
+    local privkey=$4
+    local txhash=$5
+    local relayer=${root}/relayer/relayer-node18-linux-x64
+
+    echo Starting OP relayer message prover...
+    timeout 5m bash -c "until ${relayer} prove-op-message \
```
manuelwedler, line 97: Is the retrying necessary here? Shouldn't this command run succesfully in the first place?

relayer/src/cli/programs/prove-op-message.ts
```diff
@@ -0,0 +1,50 @@
...
+    const l2ChainId = await getNetworkId(options.l2RpcUrl);
+    const l1ChainId = await getNetworkId(options.l1RpcUrl);
+
+    const OPRelayer = new OptimismRelayerService(
+      options.l1RpcUrl,
+      options.l2RpcUrl,
+      options.walletPrivateKey,
+      l1ChainId,
+      l2ChainId,
+      undefined,
```
manuelwedler, line 40: I really don't like that it is possible to pass undefined here. The constructor actually requires this parameter. 
We end up with the following: The `destinationChainId` of the `OptimismRelayerService` is declared as number (without the `?`-optional declaration), but in some cases it might be undefined. This can lead to runtime errors, which we actually try to prevent and therefore use TypeScript. Sure it might not be a problem at the moment because `destinationChainId` is only used by the ethereum relayer, but such things will bring problems in the future. You can see this with the l1 tx hash problem that we have with polygon now.

This is only possible because we don't use TypeScript's strict mode in the relayer, yet. I actually enabled it in my architectural changes for the steps. 
As we want to get this merged quickly don't enable the strict mode now, because there will be more typings that need to be solved. But it should be enabled as soon as possible.

This parameter should still be changed. Either provide a value here or make the property optional and handle the case it is undefined in all places we use the `destinationChainId`. Since it is only used in the ethereum relayer, maybe there is also another way to get it without requiring the `BaseRelayerService` to have it as a property.

Sorry for the long text, I know you quickly add such an unwanted `undefined` if your editor doesn't tell you, because strict mode is not enabled. It's just an example of how bad the relayer is in shape, because it evolved from a simple script.

docker/optimism/optimism.sh
```diff
@@ -52,7 +52,11 @@ e2e_test() {
     echo Copying contract addresses to $contract_addresses
     docker exec ops-deployer-1 cat genesis/addresses.json > $contract_addresses
     e2e_test_fill ${DEPLOYMENT_DIR} ${KEYFILE} "${password}" $l2_rpc
-    e2e_test_relayer http://localhost:9545 $l2_rpc $contract_addresses $PRIVKEY $e2e_test_l2_txhash
+    echo Sending Proof
+    e2e_test_op_proof http://localhost:8545 $l2_rpc $CONTRACT_ADDRESSES $PRIVKEY $e2e_test_l2_txhash
+    sleep 20
```
manuelwedler, line 57: Why do you sleep here? Isn't the e2e_test_relayer now retrying anyways if it fails?

manuelwedler (changes requested):

docker/optimism/optimism.sh
GabrielBuragev: Yes, we can remove this. Good points

docker/scripts/common.sh
GabrielBuragev: This command also waits until the message is ready to be proven, so this can be removed.

relayer/src/cli.ts
```diff
@@ -0,0 +1,86 @@
...
+import { OPMessageProverProgram } from "./cli/programs/prove-op-message";
+import type { ProgramOptions as RelayProgramOptions } from "./cli/programs/relay";
+import { RelayerProgram } from "./cli/programs/relay";
+import type { ExtendedProgram } from "./cli/types";
+import { killOnParentProcessChange } from "./common/process";
+
+const program = new Command();
+
+program
+  .name("cross-chain-messaging")
```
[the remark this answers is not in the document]
GabrielBuragev: I'll just use `beamer-relayer` which is the name defined in our `package.json`.
The name is not used as the command.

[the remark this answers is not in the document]
manuelwedler: I guess it is still shown in the help of the command. So `beamer-relayer` sounds much better :+1:

manuelwedler (approved):
Looks good! Thanks for implementing this so quickly. This update is quite stressful :)

relayer/src/cli/programs/prove-op-message.ts
manuelwedler: Well, you missed it because the strict mode is not enabled ;)
Enabling the typescript strict mode is quite important I think, regardless of my step architecture change. That's all I wanted to say.

Issue #1899, opened 2023-06-05
Agent calls relayer for OP prove()


fredo:
Changes required for the upcoming Bedrock update: 

The agent needs to call the relayer to prove a message on L1. This prove() call initiates the 7 day finality period window. For now we will go with the solution that as soon as a claim is challenged and the agent is involved in this challenge, he will need to immediately call the relayer to prove the message.

Only if the relayer returns successfully he doesn't need to call the relayer again.

# Acceptance Criteria

- [ ] Whenever the agent is involved in a challenge (either claimer or challenger), and the target chain is Optimism, the agent calls the relayer to prove the message
- [ ] If relayer fails, call relayer again, until the relayer returns successful
- [ ] If the relayer process is still running, do nothing

PR #1903, opened 2023-06-05
feat: add OP message prove

bilbeyt:
Closes: #1899

### DIFF
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -325,6 +325,7 @@ def process_claims(context: Context) -> None:
 
         if claim.claimer_winning.is_active or claim.challenger_winning.is_active:
             maybe_withdraw(claim, context)
+            maybe_prove(claim, context)
             if _l1_resolution_threshold_reached(claim, context):
                 maybe_resolve(claim, context)
             else:
@@ -572,10 +573,68 @@ def _timestamp_is_l1_finalized(
     return int(time.time()) > timestamp + target_chain_finality
 
 
+def maybe_prove(claim: Claim, context: Context) -> None:
+    # mainnet: 10, goerli: 420, local: 17
+    if context.target_chain_id not in (10, 420, 17):
+        claim.message_proved = True
+        return
+
+    if claim.message_proved:
+        return
+
+    request = context.requests.get(claim.request_id)
+
+    assert request is not None, "Active claim for non-existent request"
+
+    own_challenge_stake = Wei(claim.get_challenger_stake(context.address))
+    can_prove = (
+        claim.claimer == context.address and claim.challenger_exists()
+    ) or own_challenge_stake > 0
+
+    if not can_prove:
+        return
+
+    if request.fill_tx is not None:
+        proof_tx = request.fill_tx
+    elif claim.invalidation_tx is not None:
+        proof_tx = claim.invalidation_tx
+
+    if proof_tx in context.l1_resolutions:
+        return
+
+    future = context.task_pool.submit(
+        run_relayer_for_tx,
+        context.config.base_chain_rpc_url,
+        context.target_rpc_url,
+        context.source_rpc_url,
+        context.config.account.key,
+        proof_tx,
+        True,
+    )
+
+    def on_future_done(f: Future) -> None:
+        try:
+            f.result()
+        except Exception as ex:
+            context.logger.error("Optimism prove failed", ex=ex, tx_hash=proof_tx)
+        else:
+            claim.message_proved = True
+        finally:
+            context.l1_resolutions.pop(proof_tx, None)
+
+    future.add_done_callback(on_future_done)
+    context.l1_resolutions[proof_tx] = future
+
+    context.logger.info("Initiated Optimism prove", request=request, claim=claim, tx_hash=proof_tx)
+
+
 def maybe_resolve(claim: Claim, context: Context) -> bool:
     request = context.requests.get(claim.request_id)
 
     assert request is not None, "Active claim for non-existent request"
+
+    if not claim.message_proved:
+        return False
 
     if not _proof_ready_for_l1_relay(request) and not _invalidation_ready_for_l1_relay(claim):
         return False
--- a/beamer/agent/models/claim.py
+++ b/beamer/agent/models/claim.py
@@ -28,6 +28,7 @@ class Claim(StateMachine):
         self.transaction_pending = False
         self.invalidation_tx: Optional[HexBytes] = None
         self.invalidation_timestamp: Optional[Timestamp] = None
+        self.message_proved: bool = False
         self.unprocessed_claim_made_events: set[ClaimMade] = set()
         super().__init__()
 
@@ -162,6 +163,9 @@ class Claim(StateMachine):
 
         self._latest_claim_made = new_claim_made
         self.transaction_pending = False
+
+    def challenger_exists(self) -> bool:
+        return bool(self._challenger_stakes)
 
     def __repr__(self) -> str:
         state = self.current_state.id
--- a/beamer/agent/relayer.py
+++ b/beamer/agent/relayer.py
@@ -29,6 +29,7 @@ def run_relayer_for_tx(
     l2_relay_to_rpc_url: URL,
     privkey: HexBytes,
     tx_hash: HexBytes,
+    prove_tx: bool = False,
 ) -> None:
     relayer = get_relayer_executable()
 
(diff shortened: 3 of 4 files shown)


beamer/agent/chain.py
```diff
@@ -328,6 +328,7 @@ def process_claims(context: Context) -> None:
             if _l1_resolution_threshold_reached(claim, context):
                 maybe_resolve(claim, context)
             else:
+                maybe_prove(claim, context)
```
fredo, line 331: I think `maybe_prove` must be called unconditionally. It could be the case that a threshold is reached but prove was not called yet.

In this context I would pull the condition to be a Optimism chain out of the maybe_prove and put it here. This way it gets clear to the code reader that it is an Optimism depended step.

fredo: IMO we should pull the `maybe_prove` out of the condition above and call it unconditionally. The check if the message is proved is already in maybe_prove

fredo (commented):
Mostly nitpicks, but I would restructure the conditions and calls a bit to separate the Optimism condition check.

beamer/agent/chain.py
istankovic: > IMO we should pull the `maybe_prove` out of the condition above and call it unconditionally. The check if the message is proved is already in maybe_prove

I agree with this. :+1:
We could also think of this separate "proving" step as a fine-grained step where OP does something and for everything else it's a no-op. (Could also happen that future supported chains need this...)

beamer/agent/chain.py
```diff
@@ -548,6 +549,8 @@ def get_l1_cost(context: Context) -> int:
 
 def _l1_resolution_threshold_reached(claim: Claim, context: Context) -> bool:
     limit = get_l1_cost(context)
+    if context.target_chain_id in [10, 420, 17] and not claim.message_proved:
```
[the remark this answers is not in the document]
istankovic: > It seems misplaced to me to do this check here. This function is purely about a financial limit. I would rather do this check somewhere else.

Definitely. I would move this check to `maybe_prove`, with this being literally the first line in it.
So that `maybe_prove` does not do anything if it's not Optimisim or the "message" is already proved.

beamer/agent/chain.py
```diff
@@ -572,6 +575,61 @@ def _timestamp_is_l1_finalized(
...
+    own_challenge_stake = Wei(claim.get_challenger_stake(context.address))
+    can_prove = (
+        claim.claimer == context.address and claim.challenger_exists()
+    ) or own_challenge_stake > 0
+
+    if not can_prove:
+        return False
+
+    if claim.message_proved:
+        return False
```
istankovic, line 596: This check should be put together with the one regarding the chain ID and moved to top of this function, see my comment about `_l1_resolution_threshold_reached`.

beamer/agent/chain.py
```diff
@@ -572,6 +575,61 @@ def _timestamp_is_l1_finalized(
     return int(time.time()) > timestamp + target_chain_finality
 
 
+def maybe_prove(claim: Claim, context: Context) -> bool:
```
istankovic, line 578: Why does this function return `bool`, or anything? The return value does not seem to be used anywhere.

beamer/agent/chain.py
```diff
@@ -572,6 +575,61 @@ def _timestamp_is_l1_finalized(
...
+    if not can_prove:
+        return False
+
+    if claim.message_proved:
+        return False
+
+    if request.fill_tx:
+        proof_tx = request.fill_tx
+    elif claim.invalidation_tx:
+        proof_tx = claim.invalidation_tx
```
istankovic, line 601: Please check explicitly against `None`, in both `if` and `elif`.

beamer/agent/chain.py
```diff
@@ -572,6 +575,61 @@ def _timestamp_is_l1_finalized(
...
+        context.source_rpc_url,
+        context.config.account.key,
+        proof_tx,
+        True,
+    )
+
+    def on_future_done(f: Future) -> None:
+        try:
+            f.result()
+            claim.message_proved = True
```
istankovic, line 619: Nitpick: I would move the
```python
            claim.message_proved = True
```
to an `else:` block of this `try:` block, to minimise the amount of code under `try`.

istankovic: What if the `relayer prove` process fails?
We will see the failure, log it, but the details will be missing, right?
Perhaps that is best we can do right now, but I'm wondering about situations where one sees `Optimism prove failed` error in agent logs and then... what? What would be the next steps?

Also, in what cases can the `relayer prove` process fail? (Apart from the obvious, e.g. lack of funds.)

beamer/agent/models/claim.py
```diff
@@ -163,6 +164,9 @@ def _on_new_claim_made(self, new_claim_made: ClaimMade) -> None:
         self._latest_claim_made = new_claim_made
         self.transaction_pending = False
 
+    def challenger_exists(self) -> bool:
+        return len(self._challenger_stakes.keys()) > 0
```
istankovic, line 168: Nitpick: it is customary to prefer the boolean truth value of containers in cases like these, e.g.
```python
return bool(self._challenger_stakes)
```

beamer/agent/relayer.py
```diff
@@ -29,27 +29,39 @@ def run_relayer_for_tx(
...
+        "--l1-rpc-url",
+        l1_rpc,
+        "--wallet-private-key",
+        privkey.hex(),
+        "--l2-transaction-hash",
+        tx_hash.hex(),
+    ]
+
+    if prove_tx:
+        command_args.extend(["--l2-rpc-url", l2_relay_from_rpc_url])
```
istankovic, line 52: Is `--l2-rpc-url` a new relayer option, specifically for this case? Or is this a typo?

beamer/agent/chain.py
```diff
@@ -572,6 +575,61 @@ def _timestamp_is_l1_finalized(
...
+        except Exception as ex:
+            context.logger.error("Optimism prove failed", ex=ex, tx_hash=proof_tx)
+        finally:
+            context.l1_resolutions.pop(proof_tx, None)
+
+    future.add_done_callback(on_future_done)
+    context.l1_resolutions[proof_tx] = future
+
+    context.logger.info("Initiated Optimism prove", request=request, claim=claim, tx_hash=proof_tx)
```
istankovic, line 628: Since this is a call on L1 and the fact that it's not easy to debug if things go wrong here, should we perhaps consider adding a check for roughly enough balance on L1, before we invoke the relayer?

beamer/agent/chain.py
istankovic: Hmm... thinking more about this... perhaps it would be best if we made it so that this `maybe_prove` makes all messages "trivially" proved for all chains except Optimism?

E.g. pseudocode
```
def maybe_prove(...):
   if not optimism:
      claim.message_proved = True
      return

  ... all this Optimism related stuff...
```
and then in `process_claims` we could just
```python
           maybe_prove(claim, context)
           if _l1_resolution_threshold_reached(claim, context):
                maybe_resolve(claim, context)
```
To me that would look a little better because all chains except Optimism are proved by making a fill, it's just that for Optimism we have an extra step. But we stuff that inside `maybe_prove` and we don't need to worry about it.
(Future chains might also have different requirements.)

Then `maybe_resolve` should actually always check whether the message is proved.

istankovic:
I looked over this and, similar to Fred, could find only nitpicks and minor stuff. Looks good overall. :+1: 
Will leave the final approve to Fred.

beamer/agent/chain.py
bilbeyt: @GabrielBuragev can you answer this?

beamer/agent/relayer.py
bilbeyt: no, it is planned like this, so new prove command has its own set of arguments.

beamer/agent/chain.py
bilbeyt: I remember we would increase the claimStake and changing the finality period for optimism and whenever there is a challenge agent will send a prove to actually start 7 day period, if we wait for balance here then it wouldnt be possible right @fredo  ?

beamer/agent/chain.py
GabrielBuragev: > Also, in what cases can the `relayer prove` process fail? (Apart from the obvious, e.g. lack of funds.)

Excluding the failures related to incorrect arguments passed to the `relayer prove`, buggy rpcs, etc. (externalities) then this command can only fail if a transaction reverted for any possible reasons.

beamer/agent/chain.py
fredo: claimStake and the actual balance on L1 is something different.

The claimStake just makes sure that the agent will get reimbursed. For the prove transaction we decided to not do any checks regarding financial viability.

Regarding balance on L1, I don't think the way we separated tasks between the relayer and the agent, I would actually prefer to not do the check here in the agent. IMO it's the relayer's job to communicate correctly what happened when trying to execute the task.
I'd rather say that if we want to have better logging of whats going on in the relayer, we should think about communicating more via exit codes or exceptions.

beamer/agent/chain.py
fredo: > What if the relayer prove process fails?

From an agent's perspective, it only cares if the relayer did its job or not. If it failed for whatever reason, it will call the relayer again. See my comment below regarding better logging or information passing.

beamer/agent/chain.py
istankovic: My concern here is wiith this:

> agent will send a prove

The agent can only successfully  do that if there are enough funds on L1. Otherwise, the relayer will fail and we will have no way of knowing why, right? It would be nice if we captured at least `stderr` of the relayer process and dumped it into the agent's log.

fredo: If we go with that we should spin out its own issue for it, and provide a detailed design for communication between relayer and agent via exit codes or exceptions

istankovic: >  I'd rather say that if we want to have better logging of whats going on in the relayer, we should think about communicating more via exit codes or exceptions.

:this:

fredo (approved):
Looks good :+1:

PR #1914, opened 2023-06-08
ARB CI WIP

GabrielBuragev:


relayer/src/services/relayer/arbitrum.ts
```diff
@@ -95,9 +99,13 @@ export class ArbitrumRelayerService extends BaseRelayerService {
...
 
     /**
-     * Note that in principle, a single transaction could trigger any number of outgoing messages; the common case will be there's only one.
-     * For the sake of this script, we assume there's only one / just grab the first one.
+     * Note that in principle, a single transaction could trigger any number of outgoing messages; the common case (for Beamer) will be there's only one.
      */
+    let dataAvailable = false;
+    while (!dataAvailable) {
+      dataAvailable = await l2Receipt.isDataAvailable(this.l2RpcProvider as JsonRpcProvider);
+      await sleep(timeToWaitMs);
```
istankovic, line 107: The way the two statements are ordered within the loop means that if data becomes available, we will unnecessarily wait a minute before we exit the loop. It would be better to exit the loop without that extra wait time.

PR #1916, opened 2023-06-09
relayer: add fallback relay via OP CDMessenger

### LINKED ISSUE #1925: Implement fallback in relayer
For optimism the relayer needs to have a fallback in place where it executes relaying the emssage directly on the CDM instead of the portal.
Since the SDK does not support this, this needs to be implemented by hand. 
Additionally, gas limits are hardcoded to 2M gas.

GabrielBuragev:
Check commit messages for more info.

Closes #1925

### DIFF
--- a/relayer/src/services/relayer/optimism.ts
+++ b/relayer/src/services/relayer/optimism.ts
@@ -1,5 +1,10 @@
 import type { CrossChainMessage, DeepPartial, OEContractsLike } from "@eth-optimism/sdk";
-import { CrossChainMessenger, MessageReceiptStatus, MessageStatus } from "@eth-optimism/sdk";
+import {
+  CrossChainMessenger,
+  hashLowLevelMessage,
+  MessageReceiptStatus,
+  MessageStatus,
+} from "@eth-optimism/sdk";
 
 import type { TransactionHash } from "../types";
 import { BaseRelayerService } from "../types";
@@ -75,27 +80,40 @@ export class OptimismRelayerService extends BaseRelayerService {
     return;
   }
 
-  async safeRelayMessage(message: CrossChainMessage): Promise<boolean> {
+  private async isMessageWithdrawn(messageHash: string) {
+    const OptimismPortal = this.messenger.contracts.l1.OptimismPortal.connect(this.l1Wallet);
+    const isWithdrawn = await OptimismPortal.finalizedWithdrawals(messageHash);
+
+    return isWithdrawn;
+  }
+
+  private async safeRelayMessage(message: CrossChainMessage): Promise<boolean> {
+    const crossChainMessage = await this.messenger.toBedrockCrossChainMessage(message);
+    const lowLevelMessage = await this.messenger.toLowLevelMessage(crossChainMessage);
+
     try {
-      // Finalize message via OptimismPortal (it calls the L1CrossDomainMessenger internally)
-      (await this.messenger.finalizeMessage(message)).wait(1);
-    } catch (err) {
-      if (err.message.includes("OptimismPortal: withdrawal has already been finalized")) {
-        // Here, the case was that OptimismPortal marked the withdrawal as finalized but the relay still failed
+      const hash = hashLowLevelMessage(lowLevelMessage);
+      const isWithdrawn = await this.isMessageWithdrawn(hash);
+
+      if (!isWithdrawn) {
+        // Finalize message via OptimismPortal (it calls the L1CrossDomainMessenger internally)
+        (await this.messenger.finalizeMessage(message)).wait(1);
+      } else {
+        // Here, the case was that OptimismPortal marked the withdrawal as finalized but the relay probably failed
         // We need to call the L1CrossDomainMessenger to relay the message for us instead of going via the OptimismPortal
         console.log("Try relaying via messenger..");
-        const crossChainMessage = await this.messenger.toCrossChainMessage(message);
-        const withdrawal = await this.messenger.toLowLevelMessage(crossChainMessage);
 
         const tx = await this.l1Wallet.sendTransaction({
           to: this.messenger.contracts.l1.L1CrossDomainMessenger.address,
-          // withdrawal.message contains the complete transaction data
-          data: withdrawal.message,
+          // lowLevelMessage.message contains the complete transaction data
+          data: lowLevelMessage.message,
           gasLimit: 2_000_000,
         });
 
         await tx.wait(1);
-      } else if (err.message.includes("CrossDomainMessenger: message has already been relayed")) {
+      }
+    } catch (err) {
+      if (err.message.includes("CrossDomainMessenger: message has already been relayed")) {
         console.log("Message has already been relayed.");
       } else {
         throw err;
--- a/relayer/src/services/relayer/optimism.ts
+++ b/relayer/src/services/relayer/optimism.ts
@@ -95,9 +95,11 @@ export class OptimismRelayerService extends BaseRelayerService {
         });
 
         await tx.wait(1);
-      } else if (!err.message.includes("message has already been received.")) {
+      } else if (err.message.includes("CrossDomainMessenger: message has already been relayed")) {
+        console.log("Message has already been relayed.");
+      } else {
         throw err;
-      } // Otherwise the message was relayed by someone else
+      }
     }
     return true;
   }
--- a/relayer/src/services/relayer/optimism.ts
+++ b/relayer/src/services/relayer/optimism.ts
@@ -91,34 +91,27 @@ export class OptimismRelayerService extends BaseRelayerService {
     const crossChainMessage = await this.messenger.toBedrockCrossChainMessage(message);
     const lowLevelMessage = await this.messenger.toLowLevelMessage(crossChainMessage);
 
-    try {
-      const hash = hashLowLevelMessage(lowLevelMessage);
-      const isWithdrawn = await this.isMessageWithdrawn(hash);
+    const hash = hashLowLevelMessage(lowLevelMessage);
+    const isWithdrawn = await this.isMessageWithdrawn(hash);
 
-      if (!isWithdrawn) {
-        // Finalize message via OptimismPortal (it calls the L1CrossDomainMessenger internally)
-        (await this.messenger.finalizeMessage(message)).wait(1);
-      } else {
-        // Here, the case was that OptimismPortal marked the withdrawal as finalized but the relay probably failed
-        // We need to call the L1CrossDomainMessenger to relay the message for us instead of going via the OptimismPortal
-        console.log("Try relaying via messenger..");
+    if (!isWithdrawn) {
+      // Finalize message via OptimismPortal (it calls the L1CrossDomainMessenger internally)
+      (await this.messenger.finalizeMessage(message)).wait(1);
+    } else {
+      // Here, the case was that OptimismPortal marked the withdrawal as finalized but the relay probably failed
+      // We need to call the L1CrossDomainMessenger to relay the message for us instead of going via the OptimismPortal
+      console.log("Try relaying via messenger..");
 
-        const tx = await this.l1Wallet.sendTransaction({
-          to: this.messenger.contracts.l1.L1CrossDomainMessenger.address,
-          // lowLevelMessage.message contains the complete transaction data
-          data: lowLevelMessage.message,
-          gasLimit: 2_000_000,
-        });
+      const tx = await this.l1Wallet.sendTransaction({
+        to: this.messenger.contracts.l1.L1CrossDomainMessenger.address,
+        // lowLevelMessage.message contains the complete transaction data
+        data: lowLevelMessage.message,
+        gasLimit: 2_000_000,
+      });
 
-        await tx.wait(1);
-      }
-    } catch (err) {
-      if (err.message.includes("CrossDomainMessenger: message has already been relayed")) {
-        console.log("Message has already been relayed.");
-      } else {
-        throw err;
-      }
+      await tx.wait(1);
     }
+
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


relayer/src/services/relayer/optimism.ts
```diff
@@ -76,6 +76,34 @@ export class OptimismRelayerService extends BaseRelayerService {
...
   }
 
+  async safeRelayMessage(message: CrossChainMessage): Promise<boolean> {
+    try {
+      // Finalize message via OptimismPortal (it calls the L1CrossDomainMessenger internally)
+      (await this.messenger.finalizeMessage(message)).wait(1);
+    } catch (err) {
+      if (err.message.includes("OptimismPortal: withdrawal has already been finalized")) {
+        // Here, the case was that OptimismPortal marked the withdrawal as finalized but the relay still failed
+        // We need to call the L1CrossDomainMessenger to relay the message for usinstead (not going via the OptimismPortal)
```
istankovic, line 86: Typo: `usinstead`

relayer/src/services/relayer/optimism.ts
```diff
@@ -76,6 +76,32 @@ export class OptimismRelayerService extends BaseRelayerService {
     return;
   }
 
+  async safeRelayMessage(message: CrossChainMessage): Promise<boolean> {
+    try {
+      // Finalize message via OptimismPortal (it calls the L1CrossDomainMessenger internally)
+      (await this.messenger.finalizeMessage(message)).wait(1);
```
fredo, line 82: I assume you could also check it by reading `portal.finalizedWithdraws[hash]`

relayer/src/services/relayer/optimism.ts
```diff
@@ -76,6 +76,34 @@ export class OptimismRelayerService extends BaseRelayerService {
...
+        const withdrawal = await this.messenger.toLowLevelMessage(crossChainMessage);
+
+        const tx = await this.l1Wallet.sendTransaction({
+          to: this.messenger.contracts.l1.L1CrossDomainMessenger.address,
+          // withdrawal.message contains the complete transaction data
+          data: withdrawal.message,
+        });
+
+        await tx.wait(1);
+      } else if (err.message.includes("CrossDomainMessenger: message has already been relayed")) {
```
istankovic, line 98: Is it `CrossDomainMessenger` or `L1CrossDomainMessenger`?

GabrielBuragev: Its CrossDomainMessenger. This is how the exception is defined in their contracts.

GabrielBuragev: Screenshot 2023-06-09 at 13 05 05

relayer/src/services/relayer/optimism.ts
GabrielBuragev: `// We need to call the L1CrossDomainMessenger to relay the message for us instead of going via the OptimismPortal`
Is this better?

relayer/src/services/relayer/optimism.ts
```diff
@@ -76,6 +76,35 @@ export class OptimismRelayerService extends BaseRelayerService {
     return;
   }
 
+  async safeRelayMessage(message: CrossChainMessage): Promise<boolean> {
+    try {
+      // Finalize message via OptimismPortal (it calls the L1CrossDomainMessenger internally)
+      (await this.messenger.finalizeMessage(message)).wait(1);
```
istankovic, line 82: `wait(1)` means we wait max 1 second here? Is that enough? What happens if the operation takes longer than that?

relayer/src/services/relayer/optimism.ts
GabrielBuragev: The problem with this is that i will also need to do another check to see if it was relayed so we know if we actually need to call the OP CrossDomainMessenger.
Also, how do i get the hash?

GabrielBuragev: Found a utility function for hashing. Lets see about the other parts.

GabrielBuragev: So, for the first part:
>The problem with this is that i will also need to do another check to see if it was relayed so we know if we actually need to call the OP CrossDomainMessenger

I figured i dont need to do this check as i am already checking `if (status === MessageStatus.RELAYED)` before calling this function which means we will never execute this function if the message was already relayed.

---

For the second part:
> I am trying to see if there are any utility functions we can use from the SDK but i cannot find any fitting solution.

I found a way to hash the low level message in a nice and easy to understand way. Checking if there was a withdrawal is done directly via the OptimismPortal contract by checking against `finalizedWithdrawals[hash]`. If this is set true, then i relay via the Messenger.

GabrielBuragev:
I just tested the case:
```
catch (err) {
      if (err.message.includes("CrossDomainMessenger: message has already been relayed")) {
        console.log("Message has already been relayed.");
      } else {
        throw err;
      }
    }
```
and it seems like the transactions are not reverting with an error message anymore, therefore we cannot check against this message anymore. Due to that, and the recent refactor to use chain state, i am removing this try, catch block as it is useless at this point.

relayer/src/services/relayer/optimism.ts
GabrielBuragev: Done. Please take a look again and read my last comments on this PR.

istankovic:
Needs a rebase on top of latest `main`.

relayer/src/services/relayer/optimism.ts
```diff
@@ -91,34 +91,27 @@ export class OptimismRelayerService extends BaseRelayerService {
...
-        console.log("Message has already been relayed.");
-      } else {
-        throw err;
-      }
+    const hash = hashLowLevelMessage(lowLevelMessage);
+    const isWithdrawn = await this.isMessageWithdrawn(hash);
+
+    if (!isWithdrawn) {
+      // Finalize message via OptimismPortal (it calls the L1CrossDomainMessenger internally)
+      (await this.messenger.finalizeMessage(message)).wait(1);
```
fredo, line 99: can we set a gas limit of 2M here as well?

GabrielBuragev: Done.

relayer/src/services/relayer/optimism.ts
```diff
@@ -75,6 +80,41 @@ export class OptimismRelayerService extends BaseRelayerService {
     return;
   }
 
+  private async isMessageWithdrawn(messageHash: string) {
+    const OptimismPortal = this.messenger.contracts.l1.OptimismPortal.connect(this.l1Wallet);
+    const isWithdrawn = await OptimismPortal.finalizedWithdrawals(messageHash);
+
+    return isWithdrawn;
+  }
```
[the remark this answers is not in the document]
fredo: withdrawing a message. The notion of withdrawing speaks about relaying or executing a message. (withdrawing assets from a rollup is also a message)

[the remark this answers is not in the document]
compojoom: isMessageWithdrawn sound like a message status: "The status of the message is withdrawn", but the optimism sdk doesn't define a withdrawn status:
https://github.com/ethereum-optimism/optimism/blob/develop/packages/sdk/src/interfaces/types.ts
```
export enum MessageStatus {
  /**
   * Message is an L1 to L2 message and has not been processed by the L2.
   */
  UNCONFIRMED_L1_TO_L2_MESSAGE,

  /**
   * Message is an L1 to L2 message and the transaction to execute the message failed.
   * When this status is returned, you will need to resend the L1 to L2 message, probably with a
   * higher gas limit.
   */
  FAILED_L1_TO_L2_MESSAGE,

  /**
   * Message is an L2 to L1 message and no state root has been published yet.
   */
  STATE_ROOT_NOT_PUBLISHED,

  /**
   * Message is ready to be proved on L1 to initiate the challenge period.
   */
  READY_TO_PROVE,

  /**
   * Message is a proved L2 to L1 message and is undergoing the challenge period.
   */
  IN_CHALLENGE_PERIOD,

  /**
   * Message is ready to be relayed.
   */
  READY_FOR_RELAY,

  /**
   * Message has been relayed.
   */
  RELAYED,
}
```

Also the OptimismPortal contract doesn't say anything about MessageWithdrawn, but more in the direction of transaction Withdrawn, which implies that a deposit is withdrawn.

https://github.com/ethereum-optimism/optimism/blob/62c7f3b05a70027b30054d4c8974f44000606fb7/packages/contracts-bedrock/contracts/L1/OptimismPortal.sol#L18

That's why I feel somewhat confused by "messageWithdrawn", but if we want to keep it general, then I guess it is ok.

relayer/src/services/relayer/optimism.ts
```diff
@@ -88,6 +88,7 @@ export class OptimismRelayerService extends BaseRelayerService {
   }
 
   private async safeRelayMessage(message: CrossChainMessage): Promise<boolean> {
+    const ESTIMATION_CONTRACT_ADDRESS = "0x0000000000000000000000000000000000000001";
```
compojoom, line 91: Maybe add a doc comment to the optimism contract where they expect this address for estimating gas.

GabrielBuragev: 👍

relayer/src/services/relayer/optimism.ts
```diff
@@ -97,19 +98,43 @@ export class OptimismRelayerService extends BaseRelayerService {
 
     if (!isWithdrawn) {
       // Finalize message via OptimismPortal (it calls the L1CrossDomainMessenger internally)
+      const OptimismPortalContract = await this.messenger.contracts.l1.OptimismPortal.connect(
+        this.l1RpcProvider,
+      );
+      const gasLimit = await OptimismPortalContract.estimateGas.finalizeWithdrawalTransaction(
```
compojoom, line 104: Is this https://docs.ethers.org/v5/api/contract/contract/#contract-estimateGas a common thing in other projects?

I'm more used to the syntax where you do provider.estimateGas(tx).

relayer/src/services/relayer/optimism.ts
```diff
@@ -97,19 +98,43 @@ export class OptimismRelayerService extends BaseRelayerService {
...
+      });
+
       tx = await this.l1Wallet.sendTransaction({
         to: this.messenger.contracts.l1.L1CrossDomainMessenger.address,
         // lowLevelMessage.message contains the complete transaction data
         data: lowLevelMessage.message,
-        gasLimit: 2_000_000,
+        gasLimit,
       });
     }
```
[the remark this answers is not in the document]
GabrielBuragev: I agree, i'll try to make it cleaner.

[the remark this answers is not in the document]
GabrielBuragev: Check again.

relayer/src/services/relayer/optimism.ts
```diff
@@ -8,12 +14,21 @@ export class OptimismRelayerService extends BaseRelayerService {
   customNetworkContracts: DeepPartial<OEContractsLike> | undefined;
   messenger: CrossChainMessenger | undefined;
 
+  /** Used in the gas estimation process
+   *  https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/contracts/L1/OptimismPortal.sol#L417
+   *  https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/contracts/libraries/Constants.sol#L14-L21
+   *
+   *  This is the only way to catch reverts when relaying messages with incorrectly set minGasLimit. We also catch internal reverts from our contracts.
```
compojoom, line 21: I'm not 100% sure this is a correct description. By using the 0x1 address in the estimateGas call we basically force a revert in the contract, which then forces the estimate gas algorithm to also look at different execution paths adding more gas to the calculation. If the call is not being done from the 0x1 address, then upon hitting the check about the gasLimit the function just returns and we end up with a low gasLimit value.

I would go with something of the sort:

This is a "hack" in the optimism contracts. If the user provides a too low gasLimit the solidity call will fail. By faking the transaction as coming from the 0x1 address the optimism contracts are triggering a revert that causes the estimate algorithm to check all other possible execution paths and as such returns a higher/correct gasLimit value.

relayer/src/services/relayer/optimism.ts
```diff
@@ -75,6 +90,90 @@ export class OptimismRelayerService extends BaseRelayerService {
...
+  }
+
+  private async isMessageWithdrawn(messageHash: string) {
+    const OptimismPortal = this.messenger.contracts.l1.OptimismPortal.connect(this.l1Wallet);
+    const isWithdrawn = await OptimismPortal.finalizedWithdrawals(messageHash);
+
+    return isWithdrawn;
+  }
+
+  private async relayMessageViaCrossDomainMessenger(
```
compojoom, line 107: I would add some docblocks explaining the quirks of the optimism contracts. If it wasn't for @fredo TLDR-ing me on the problem I wouldn't understand why we need to use different contracts for basically the same thing. 
3 months from now we wouldn't know it either. 

Other than that - it's great!

relayer/src/services/relayer/optimism.ts
fredo: >  If the user provides a too low gasLimit the solidity call will fail.

The call will actually not fail. The transaction will end successfully (and not even calling our internal message relay)

relayer/src/services/relayer/optimism.ts
```diff
@@ -75,6 +100,102 @@ export class OptimismRelayerService extends BaseRelayerService {
...
+   * 2. A call to `CrossDomainMessenger.relayMessage` that was forwarded by the OptimismPortal in the previous step which in the end executes our message on L1.
+   *
+   * The first call will always result in a successful transaction no matter the status of the underlying call (CrossDomainMessenger.relayMessage).
+   * When the underlying message execution on L1 failed, we have no way to replay the relay process via the OptimismPortal anymore as
+   * the message was marked as withdrawn in the first call.
+   * What we need to do instead is to go via the `CrossDomainMessenger.relayMessage` to replay a failed message relay.
+   *
+   * This function takes care of handling these failures by deciding on which route should be used.
+   */
+
```
compojoom, line 178: empty line before the function

GabrielBuragev: Done

Issue #1921, opened 2023-06-12
extended e2e test for optmism relayer


fredo:
With OP Bedrock, the relay process changed in such a way that the contract that is responsible for relaying messages changed. It is the OptimismPortal's responsibility to take care of this now.
During the research we found out that a failed message relay (via the OptimismPortal) cannot be replayed via the same contract.
For such cases, the relayer needs to be able to detect when a previous relay attempt via the OptimismPortal succeeded but the underlying message was not relayed and use a fallback approach (by calling the OP's L1CrossDomainMessenger directly).

e2e Test scenario
- Send a fill transaction with a _fictitious chain id_
- execute prove with relayer
- Simulate a failed relay scenario: execute a failing relay via the OptimismPortal by using an external script (not relayer)
- Make sure the relay via OptimismPortal (previous step) was executed but message failed to relay
- add the _fictitious chain id_ to the resolver's configuration pointing to the correct Optimism's L1 Messenger and RequestManager
- execute relay with our relayer this time
- assert that the fallback was used (via CrossDomainMessenger)
- assert that message arrived at the RequestManager

PR #1933, opened 2023-06-13
feat: op relayer fallback e2e

bilbeyt:
Closes: #1921

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -232,6 +232,7 @@ jobs:
       - run: bash $OPTIMISM_FOLDER/optimism.sh up
       - run: bash $OPTIMISM_FOLDER/optimism.sh deploy-beamer
       - run: bash $OPTIMISM_FOLDER/optimism.sh e2e-test
+      - run: bash $OPTIMISM_FOLDER/optimism.sh e2e-test-fallback
       - run: bash $OPTIMISM_FOLDER/optimism.sh down
 
   e2e-test-arbitrum:
--- a/docker/optimism/optimism.sh
+++ b/docker/optimism/optimism.sh
@@ -22,7 +22,7 @@ e2e_test_op_proof(){
     local l2_rpc=$2
     local privkey=$3
     local txhash=$4
-    local relayer=${ROOT}/relayer/relayer-node18-linux-x64
+    local relayer=$(get_relayer_binary)
 
     echo Starting OP relayer message prover...
     timeout 5m bash -c "until ${relayer} prove-op-message \
@@ -95,15 +95,55 @@ e2e_test() {
     e2e_test_verify ${DEPLOYMENT_DIR} $l2_rpc $ADDRESS $e2e_test_request_id
 }
 
+
+e2e_test_fallback() {
+    relayer=$(get_relayer_binary)
+    l2_rpc=http://localhost:9545
+    password=""
+
+    export SOURCE_CHAIN_ID=123
+    e2e_test_fill ${DEPLOYMENT_DIR} ${KEYFILE} "${password}" $l2_rpc
+    echo Sending Proof
+    
+    e2e_test_op_proof http://localhost:8545 $l2_rpc $PRIVKEY $e2e_test_l2_txhash
+    echo L1 Resolve
+    sleep 15
+    # we need this relay call to fail
+    if ${relayer} relay \
+        --l1-rpc-url http://localhost:8545 \
+        --l2-relay-to-rpc-url $l2_rpc \
+        --l2-relay-from-rpc-url $l2_rpc \
+        --wallet-private-key $PRIVKEY \
+        --l2-transaction-hash $e2e_test_l2_txhash; then 
+        echo Relayer failed to fail
+        exit 1
+    else
+        echo Relayer failed as expected
+    fi
+    local output=$(poetry run python $ROOT/scripts/e2e-test-op-commands.py ${KEYFILE} "${password}" http://localhost:8545 verify-portal-call)
+    export BLOCK_NUMBER=$(echo "$output" | awk -F: '/Block Number/ { print $2 }')
+    poetry run python $ROOT/scripts/e2e-test-op-commands.py ${KEYFILE} "${password}" http://localhost:8545 set-chain-on-resolver ${DEPLOYMENT_DIR} $l2_rpc
+    ${relayer} relay \
+        --l1-rpc-url http://localhost:8545 \
+        --l2-relay-to-rpc-url $l2_rpc \
+        --l2-relay-from-rpc-url $l2_rpc \
+        --wallet-private-key $PRIVKEY \
+        --l2-transaction-hash $e2e_test_l2_txhash
+    poetry run python $ROOT/scripts/e2e-test-op-commands.py ${KEYFILE} "${password}" http://localhost:8545 verify-messenger-call
+    e2e_test_verify ${DEPLOYMENT_DIR} $l2_rpc $ADDRESS $e2e_test_request_id
+}
+
 usage() {
     cat <<EOF
-$0  [up | down | deploy-beamer | e2e-test ]
+$0  [up | down | deploy-beamer | e2e-test | e2e-test-fallback]
 
 Commands:
-  up             Bring up a private Optimism instance.
-  down           Stop the Optimism instance.
-  deploy-beamer  Deploy Beamer contracts.
-  e2e-test       Run a test that verifies L2 -> L1 -> L2 messaging.
+  up                    Bring up a private Optimism instance.
+  down                  Stop the Optimism instance.
+  deploy-beamer         Deploy Beamer contracts.
+  e2e-test              Run a test that verifies L2 -> L1 -> L2 messaging.
+  e2e-test-fallback     Run a test that verifies relayer's fallback mechanism.
+
 EOF
 }
 
@@ -123,6 +163,10 @@ case $1 in
 
     e2e-test)
         e2e_test
+        ;;
+
+    e2e-test-fallback)
+        e2e_test_fallback
         ;;
 
     *)
--- a/scripts/e2e-test-op-commands.py
+++ b/scripts/e2e-test-op-commands.py
@@ -1,0 +1,115 @@
+import json
+import os
+from pathlib import Path
+from typing import Any
+
+import click
+from eth_account.account import LocalAccount
+from eth_utils import to_checksum_address
+from hexbytes import HexBytes
+from web3 import Web3
+
+from beamer.contracts import contracts_for_web3, make_contracts, prepare_deployment_infos
+from beamer.typing import URL, ChainId
+from beamer.util import account_from_keyfile, make_web3, transact
+from scripts._util import pass_args
+
+
+# Topic signatures are taken from:
+# https://github.com/ethereum-optimism/optimism/blob/develop/op-bindings/bindings/l1crossdomainmessenger.go
+# https://github.com/ethereum-optimism/optimism/blob/develop/op-bindings/bindings/optimismportal.go
+FAILED_RELAYED_MESSAGE_TOPIC = HexBytes(
+    "0x99d0e048484baa1b1540b1367cb128acd7ab2946d1ed91ec10e3c85e4bf51b8f"
+)
+WITHDRAWAL_FINALIZED_TOPIC = HexBytes(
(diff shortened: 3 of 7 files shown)


docker/optimism/optimism.sh
```diff
@@ -95,15 +95,52 @@ e2e_test() {
...
+    timeout 1m bash -c "until ${relayer} relay \
+                                         --l1-rpc-url http://localhost:8545 \
+                                         --l2-relay-to-rpc-url $l2_rpc \
+                                         --l2-relay-from-rpc-url $l2_rpc \
+                                         --wallet-private-key $PRIVKEY \
+                                         --l2-transaction-hash $e2e_test_l2_txhash; \
+                        do sleep 1s; done"
+    PORTAL_ADDRESS="0x6900000000000000000000000000000000000001"
+    MESSENGER_ADDRESS="0x6900000000000000000000000000000000000002"
+    local output = poetry run scripts/e2e-test-op-commands.py ${KEYFILE} "${password}" http://localhost:8545 verify-portal-call $PORTAL_ADDRESS $MESSENGER_ADDRESS
```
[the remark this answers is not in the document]
bilbeyt: some kind of state tracking, i am looking from start in this command, then with this i am aiming to get block number and save it as environment variable which i will use in another call after successful relay.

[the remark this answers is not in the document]
bilbeyt: i dont want to get older events in that command

docker/optimism/optimism.sh
```diff
@@ -95,15 +95,53 @@ e2e_test() {
     e2e_test_verify ${DEPLOYMENT_DIR} $l2_rpc $ADDRESS $e2e_test_request_id
 }
 
+
+e2e_test_fallback() {
+    set +eo pipefail
```
istankovic, line 100: What is the reason for this line?

docker/optimism/optimism.sh
```diff
@@ -95,15 +95,53 @@ e2e_test() {
...
+    e2e_test_op_proof http://localhost:8545 $l2_rpc $PRIVKEY $e2e_test_l2_txhash
+    echo L1 Resolve
+    sleep 15
+    ${relayer} relay \
+        --l1-rpc-url http://localhost:8545 \
+        --l2-relay-to-rpc-url $l2_rpc \
+        --l2-relay-from-rpc-url $l2_rpc \
+        --wallet-private-key $PRIVKEY \
+        --l2-transaction-hash $e2e_test_l2_txhash 
+    set -eo pipefail
```
istankovic, line 118: I am assuming there is a reason for this, but would like to know it.

scripts/e2e-test-op-commands.py
```diff
@@ -0,0 +1,137 @@
...
+) -> None:
+    portal_logs = web3_l1.eth.get_logs(
+        {"address": to_checksum_address(portal_address), "fromBlock": 0}
+    )
+    messenger_logs = web3_l1.eth.get_logs(
+        {"address": to_checksum_address(messenger_address), "fromBlock": 0}
+    )
+    assert any(WITHDRAWAL_FINALIZED_TOPIC in log["topics"] for log in portal_logs)
+    assert any(FAILED_RELAYED_MESSAGE_TOPIC in log["topics"] for log in messenger_logs)
+    print("Processed Block Number:", web3_l1.eth.block_number)
```
istankovic, line 72: What does `Processed` here mean? Why not just `Block number`?

scripts/e2e-test-op-commands.py
```diff
@@ -0,0 +1,137 @@
...
+)
+@click.argument("password", type=str, required=True)
+@click.argument("l1-rpc", type=str, required=True)
+@click.pass_context
+def cli(ctx: Any, keystore_file: Path, password: str, l1_rpc: str) -> None:
+    account = account_from_keyfile(keystore_file, password)
+    web3_l1 = make_web3(URL(l1_rpc), account)
+    ctx.ensure_object(dict)
+    ctx.obj["account"] = account
+    ctx.obj["web3_l1"] = web3_l1
```
istankovic, line 45: This is a cool way to reduce duplication of code, I think I will use this approach in other commands. :+1:

scripts/e2e-test-op-commands.py
```diff
@@ -0,0 +1,137 @@
...
+import structlog
+from eth_account.account import LocalAccount
+from eth_utils import to_checksum_address
+from hexbytes import HexBytes
+from web3 import Web3
+
+from beamer.contracts import contracts_for_web3, make_contracts, prepare_deployment_infos
+from beamer.typing import URL, ChainId
+from beamer.util import account_from_keyfile, make_web3, transact
+from scripts.call_contracts import pass_args
```
istankovic, line 16: This script should not import (or depend on) `call_contracts` script.

scripts/e2e-test-op-commands.py
```diff
@@ -0,0 +1,137 @@
...
+
+logger = structlog.get_logger(__name__)
```
istankovic, line 18: We are importing `structlog`, initializing the logger and then... just use regular `print` statement.
Let's then remove `structlog` completely from this script.

scripts/e2e-test-op-commands.py
```diff
@@ -0,0 +1,137 @@
...
+
+FAILED_RELAYED_MESSAGE_TOPIC = HexBytes(
+    "0x99d0e048484baa1b1540b1367cb128acd7ab2946d1ed91ec10e3c85e4bf51b8f"
+)
+WITHDRAWAL_FINALIZED_TOPIC = HexBytes(
+    "0xdb5c7652857aa163daadd670e116628fb42e869d8ac4251ef8971d9e5727df1b"
+)
+RELAYED_MESSAGE_TOPIC = HexBytes(
+    "0x4641df4a962071e12719d8c8c8e5ac7fc4d97b927346a3d7a335b1f7517e133c"
+)
```
istankovic, line 28: It would be nice to have a comment saying where these come from, e.g. point to a contract etc.

scripts/e2e-test-op-commands.py
```diff
@@ -0,0 +1,137 @@
...
+@click.argument("l2-rpc", type=str)
+@pass_args
+def set_messenger_on_resolver(
+    account: LocalAccount, web3_l1: Web3, deployment_dir: Path, l2_rpc: URL
+) -> None:
+    web3 = make_web3(l2_rpc, account)
+    env_chain_id = os.getenv("SOURCE_CHAIN_ID")
+    assert env_chain_id is not None
+    target_chain_id = ChainId(int(env_chain_id))
```
istankovic, line 89: You can replace these with just one line:
```python
    target_chain_id = ChainId(int(os.environ['SOURCE_CHAIN_ID']))
```
If the env variable is not there, you will get an except, same as on `assert`.

Also, why are we calling it `target_chain_id` when we are using `SOURCE_CHAIN_ID` for the env variable?

scripts/e2e-test-op-commands.py
```diff
@@ -0,0 +1,137 @@
...
+    type=str,
+)
+@pass_args
+def verify_messenger_call(
+    account: LocalAccount,  # pylint:disable=unused-argument
+    web3_l1: Web3,
+    portal_address: str,
+    messenger_address: str,
+) -> None:
+    block_number = int(os.environ["PROCESSED_BLOCK_NUMBER"])
```
istankovic, line 125: I'm not sure why this is called "processed" block number and not just "block number".

istankovic:
Commit b2f536ccfc43a1337d712e575caba9a982393623 should probably be first, as it prepares the terrain for the Python script.

.circleci/workflows.yml
```diff
@@ -234,6 +234,17 @@ jobs:
...
+    executor: vm
+    environment:
+      OPTIMISM_FOLDER: "docker/optimism"
+    steps:
+      - initialize-environment
+      - run: bash $OPTIMISM_FOLDER/optimism.sh up
+      - run: bash $OPTIMISM_FOLDER/optimism.sh deploy-beamer
+      - run: bash $OPTIMISM_FOLDER/optimism.sh e2e-test-fallback
+      - run: bash $OPTIMISM_FOLDER/optimism.sh down
+
```
istankovic, line 247: This is wasteful. Why not just add `e2e-test-fallback` to the existing `e2e-test-optimism`?

istankovic (changes requested):
Needs more work.

docker/optimism/optimism.sh
```diff
@@ -95,15 +95,53 @@ e2e_test() {
...
+    PORTAL_ADDRESS="0x6900000000000000000000000000000000000001"
+    MESSENGER_ADDRESS="0x6900000000000000000000000000000000000002"
```
istankovic, line 120: Are these addresses going to change? Do they need to be here and be given as an argument to the `verify-portal-call`, instead of just being hardcoded in the Python script (like event topics are)?

docker/optimism/optimism.sh
```diff
@@ -95,15 +95,53 @@ e2e_test() {
...
+    l2_rpc=http://localhost:9545
+    password=""
+
+    export SOURCE_CHAIN_ID=123
+    e2e_test_fill ${DEPLOYMENT_DIR} ${KEYFILE} "${password}" $l2_rpc
+    echo Sending Proof
+    
+    e2e_test_op_proof http://localhost:8545 $l2_rpc $PRIVKEY $e2e_test_l2_txhash
+    echo L1 Resolve
+    sleep 15
```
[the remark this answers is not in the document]
bilbeyt: it is a required time to wait for l1 resolve to get ready without using any timeout etc.

docker/optimism/optimism.sh
bilbeyt: we need to let relayer to fail and if we dont do this then whole pipeline is failing.

docker/optimism/optimism.sh
bilbeyt: added to script.

.circleci/workflows.yml
bilbeyt: added

scripts/e2e-test-op-commands.py
bilbeyt: changed

scripts/e2e-test-op-commands.py
bilbeyt: it is not a contract but golang test files in op code i found them :D (signatures)

scripts/e2e-test-op-commands.py
bilbeyt: changed

docker/optimism/optimism.sh
```diff
@@ -95,15 +95,53 @@ e2e_test() {
...
 Commands:
-  up             Bring up a private Optimism instance.
-  down           Stop the Optimism instance.
-  deploy-beamer  Deploy Beamer contracts.
-  e2e-test       Run a test that verifies L2 -> L1 -> L2 messaging.
+  up                    Bring up a private Optimism instance.
+  down                  Stop the Optimism instance.
+  deploy-beamer         Deploy Beamer contracts.
+  e2e-test              Run a test that verifies L2 -> L1 -> L2 messaging.
+  e2e-test-fallback     Run a test that verifies L2 -> L1 -> L2 messaging.
```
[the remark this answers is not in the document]
bilbeyt: changed

docker/optimism/optimism.sh
bilbeyt: pointed the reason in above comment

scripts/e2e-test-op-commands.py
```diff
@@ -0,0 +1,115 @@
...
+
+from beamer.contracts import contracts_for_web3, make_contracts, prepare_deployment_infos
+from beamer.typing import URL, ChainId
+from beamer.util import account_from_keyfile, make_web3, transact
+from scripts._util import pass_args
+
+
+# Topic signatures are taken from:
+# https://github.com/ethereum-optimism/optimism/blob/develop/op-bindings/bindings/l1crossdomainmessenger.go
+# https://github.com/ethereum-optimism/optimism/blob/develop/op-bindings/bindings/optimismportal.go
```
istankovic, line 20: Thanks for this comment!

docker/optimism/optimism.sh
istankovic: But don't we already set the `pipefail` option and `e` in `common.sh`?

istankovic: Ah, I'm dumb.

istankovic (approved):

Issue #1832, opened 2023-05-16
Add a new command to write contract configuration


istankovic:
## Writing configuration

`beamer config write --rpc-file RPC-FILE --artifact CHAIN-NAME.deployment.json current.state desired.state`

The `write` command takes an existing artifact file, a path to a "current" state file
and a path to the desired state file. It then issues appropriate transactions so that
the contracts' chain state reflects the desired state.

More concretely, the command does the following things, in order:

- read the current state file and note the block number, `block`
- compute the checksum and check if it matches the one written in the current state file
- if the checksums don't match, exit with an error
- read the desired state file (this one lacks both the block number and the checksum)
- get the latest block number, `latest_block`
- if `latest_block <= block`, exit with an error
- if there were any configuration changes between `block` and `latest_block`, exit with an error
- compute the set `S` of contract calls that need to be performed in order to make the desired
  state current
- for each call in `S`
  - get the current state from the RPC
  - if the desired state differs from the current state, issue a transaction to the RPC
- compute the checksum of `latest_block` and the desired state
- write the `latest_block`, checksum and the desired state to the desired state file

## Tasks

- [x] add the new `beamer config write` command
- [x] write a test for the case where no transactions need to be issued (i.e. the config on chain is already in the desired state)
- [ ] write a test that makes sure only necessary transactions are issued
- [x] write documentation

istankovic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-62442a69c5e6e60010e31cf3/issues/beamer-bridge/beamer/1832?planning-poker) with Zenhub @bilbeyt @compojoom @fredo @GabrielBuragev @manuelwedler @panpilkarz

PR #1951, opened 2023-06-16
Config write

istankovic:
This PR adds the `config write` command, related tests and docs.

Closes: #1832

### DIFF
--- a/beamer/config/commands.py
+++ b/beamer/config/commands.py
@@ -1,13 +1,16 @@
+import sys
 from pathlib import Path
+from typing import Generator
 
 import click
 import structlog
+from eth_typing import ChecksumAddress
 from web3 import Web3
 
 import beamer.contracts
 import beamer.deploy.config
 import beamer.util
-from beamer.config.state import ChainConfig, Configuration, TokenConfig
+from beamer.config.state import ChainConfig, Configuration, DesiredConfiguration, TokenConfig
 from beamer.deploy.artifacts import Deployment
 from beamer.events import (
     ChainUpdated,
@@ -18,9 +21,13 @@ from beamer.events import (
     LpRemoved,
     TokenUpdated,
 )
+from beamer.typing import BlockNumber
 from beamer.util import get_ERC20_abi, make_web3
 
 log = structlog.get_logger(__name__)
+
+
+_CONFIG_UPDATE_EVENTS = frozenset((ChainUpdated, FeesUpdated, TokenUpdated, LpAdded, LpRemoved))
 
 
 def _replay_event(w3: Web3, deployment: Deployment, config: Configuration, event: Event) -> None:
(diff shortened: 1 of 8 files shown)


beamer/config/commands.py
```diff
@@ -18,11 +21,15 @@
     LpRemoved,
     TokenUpdated,
 )
+from beamer.typing import BlockNumber
 from beamer.util import get_ERC20_abi, make_web3
 
 log = structlog.get_logger(__name__)
 
 
+_CONFIG_UPDATE_EVENTS = frozenset((FeesUpdated, LpAdded, LpRemoved, TokenUpdated))
```
bilbeyt, line 30: ChainUpdated is missing here.

beamer/tests/config/test_write.py
```diff
@@ -0,0 +1,205 @@
...
+        )
+
+
+@pytest.fixture
+def tmp_deployment_path(tmp_path_factory):
+    return tmp_path_factory.mktemp("deployment")
+
+
+@pytest.fixture
+def deployment(tmp_deployment_path, deployer):
```
bilbeyt, line 59: maybe we can change the name to deployment_summary just to differentiate between deployment instance below

docs/source/commands.rst
```diff
@@ -96,6 +96,41 @@ are fetched from the deployment block.
      - Path to the JSON file containing RPC information.
 
 
+.. _command-config-write:
+
+``beamer config write``
+^^^^^^^^^^^^^^^^^^^^^^^
+
+``beamer config write --rpc-file RPC-FILE --abi-dir DIR --artifact CHAIN-NAME.deployment.json
```
bilbeyt, line 104: CHAIN_ID-CHAIN-NAME.deployment.json

docs/source/commands.rst
```diff
@@ -96,6 +96,41 @@ are fetched from the deployment block.
...
+.. list-table::
+   :header-rows: 1
+
+   * - Command-line option
+     - Description
+
+   * - ``--abi-dir DIR``
+     - The directory containing contract ABI files.
+
+   * - ``--artifact CHAIN-NAME.deployment.json``
```
bilbeyt, line 121: also here

beamer/config/commands.py
```diff
@@ -110,19 +110,20 @@ def read(
...
+        start_block = config.block
     else:
+        start_block = min(
+            deployment.chain.contracts["RequestManager"].deployment_block,
+            deployment.chain.contracts["FillManager"].deployment_block,
+        )
         config = Configuration.initial(chain_id, start_block)
 
+    fetcher = EventFetcher(
+        w3, (request_manager, fill_manager), start_block=start_block, confirmation_blocks=0
```
fredo, line 132: Are we potentially replaying the events for `start_block` twice? Meaning, is `config.block` and `EventFetcher.start_block` both inclusive? 
I think it is most likely not a problem but it would be great to get this straight.

beamer/config/commands.py
```diff
@@ -131,3 +138,255 @@ def read(
...
+            address = current_config.token_addresses[symbol]
+            # The token is disabled by setting its transfer limit to zero,
+            # but we set also eth_in_token to zero.
+            yield "RequestManager", "updateToken", address, 0, 0
+
+    # Then check for newly added and modified tokens.
+    for symbol, desired_token in desired_tokens.items():
+        current_token = current_tokens.get(symbol)
+        if current_token is None or current_token != desired_token:
+            address = desired_config.token_addresses[symbol]
```
fredo, line 197: Can't we do this check on validation of the config? I would consider a config where the token mapping from symbol to addresses for a given token is missing, an invalid config

beamer/config/commands.py
```diff
@@ -131,3 +138,255 @@ def read(
...
+    start_block = BlockNumber(current_config.block + 1)
+    _ensure_no_config_updates_since(w3, deployment, start_block)
+
+    receipt = None
+    for contract, function, *args in _generate_updates(current_config, desired_config):
+        log.info("Sending transaction", call=f"{contract}.{function}({', '.join(map(str, args))})")
+        contract = deployment.obtain_contract(w3, "chain", contract)
+        call = getattr(contract.functions, function)(*args)
+        try:
+            receipt = beamer.util.transact(call)
```
fredo, line 380: nitpick and not really related: I was confused what transact can actually return. 
Return type is Any but looking at the function, it can only return `TxReceipt`. Wouldn't it be a good time to change that?

beamer/config/commands.py
```diff
@@ -131,3 +138,255 @@ def read(
...
+        except beamer.util.TransactionFailed as exc:
+            log.error("Transaction failed", exc=exc)
+            sys.exit(1)
+        else:
+            txhash = receipt.transactionHash.hex()
+            log.info("Transaction sent", block=receipt.blockNumber, txhash=txhash)
+
+    if receipt is not None:
+        # Update desired config with block number and checksum.
+        desired_config.block = receipt.blockNumber
```
fredo, line 390: can you please elaborate a bit on the concept of block number and checksum in the desired config file. Why do we need to update it and what are the initial values set to? 
I consider the block parameter to be a value specifically set for a state file. It tells me the block number of the state. If I have a desired state file, I would assume it shows me a desired future state. So I don't understand what the block would mean in this case. 
Same goes for the checksum. I would assume that the checksum secures that a historic state file has not been tampered with. For a future desired file, I actually do not really understand the need for it. So in both cases I would argue that it's not supposed to be in the file in the first place.

beamer/config/commands.py
```diff
@@ -131,3 +138,255 @@ def read(
...
+        desired_config.to_file(desired_state_path)
```
fredo, line 391: From this line I can see that actually the desired state file becomes the new "historic" state file.  However is this the correct way to express that the update was successful? 
I would probably rather create a new state file updated to the current block, if that is wished (For now I would probably fine to say, update was successful and don't write anything to a file, Checks could be done afterwards with a new read).
However, a new state file output after config write is a nice to have. But I would probably not go and modify the desired config file, but rather create a proper new state file to the current block.

beamer/config/commands.py
istankovic: :eyes: 
Definitely.  Fixed, thanks!

beamer/config/commands.py
istankovic: Good point! Changed.

beamer/tests/config/util.py
```diff
@@ -18,6 +24,25 @@ def write_keystore_file(path, private_key, password):
     path.write_text(json.dumps(obj))
 
 
+def read_config_state(rpc_file, artifact):
+    root = pathlib.Path(__file__).parents[3]
```
fredo, line 29: I foresee some time of debugging if we ever change the directory structure here 🤓

fredo (dismissed):
Looks very good ! 🍷 
I would like to discuss the structure of desired state files first, though

beamer/config/commands.py
istankovic: Removed this part. Regarding the block number and checksum we are going to have separate Python types for the regulard configuration and desired configuration so we distinguish between them on type level too.

beamer/config/commands.py
istankovic: Removed.

beamer/tests/config/util.py
istankovic: > I foresee some time of debugging if we ever change the directory structure here nerd_face

Hm, yeah, that's not too ideal and we already have a couple, including one in `test_cli` from agent tests.
So I added `beamer.tests.util.get_repo_root` and updated all the places to make use of that function instead of doing it manually.

beamer/config/commands.py
istankovic: > Can't we do this check on validation of the config? I would consider a config where the token mapping from symbol to addresses for a given token is missing, an invalid config

We already have that check when loading the configuration.

This is about something else: when we generate updates for tokens, we cannot assume that a desired token e.g. `TST` already exists in the current configuration. So we need to check first whether `current_token is None` -- if it is, that means we have a new token in the desired configuration, for which we need to issue a transaction.

Similarly, if `current_token != desired_token` (note: `current_token` and `desired_token` are of type `TokenConfig`) then the token does exist in the current configuration, but its configuration differs from the desired one, which means we need to issue a transaction.

These two cases are what the comment above the `for` loop points at:
```
# Then check for newly added and modified tokens.
```

I hope that clarifies things.

beamer/tests/config/test_write.py
```diff
@@ -0,0 +1,205 @@
...
+    rpc_file, artifact = deploy(deployer, tmp_deployment_path)
+    deployment = Deployment.from_file(artifact)
+    assert deployment.chain is not None
+
+    token = deployer.deploy(ape.project.MintableToken, int(1e18))
```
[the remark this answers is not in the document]
istankovic: Good idea, changed. :+1:

beamer/tests/config/test_write.py
istankovic: Decided on `deployment_objects` in a joint discussion.

beamer/config/state.py
```diff
@@ -83,6 +76,65 @@ def _check_tokens(self) -> Generator:
...
+
+    def to_file(self, path: Path) -> None:
+        data = apischema.serialize(self)
+        with open(path, "wt") as f:
+            json.dump(data, f, indent=4)
+
+    def to_config(self, block: BlockNumber) -> "Configuration":
+        data = apischema.serialize(self, no_copy=False)
+        data["block"] = block
+        return apischema.deserialize(Configuration, data)
```
bilbeyt, line 121: not sure if we need try/except here like other places

beamer/config/state.py
```diff
@@ -97,25 +149,17 @@ def initial(chain_id: ChainId, block: BlockNumber) -> "Configuration":
             fill_manager=fm_config,
         )
 
+    def to_desired_config(self) -> DesiredConfiguration:
+        data = apischema.serialize(self, no_copy=False)
+        del data["block"]
+        return apischema.deserialize(DesiredConfiguration, data)
```
bilbeyt, line 155: also here

beamer/config/commands.py
```diff
@@ -131,3 +138,250 @@ def read(
...
+        for address in desired_whitelist - current_whitelist:
+            yield contract, "addAllowedLp", address
+
+    yield from helper(
+        "RequestManager",
+        current_config.request_manager.whitelist,
+        desired_config.request_manager.whitelist,
+    )
+    yield from helper(
+        "FillManager", current_config.fill_manager.whitelist, desired_config.fill_manager.whitelist
```
bilbeyt, line 228: we may also use same formatting for RequestManager

beamer/tests/config/util.py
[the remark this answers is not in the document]
bilbeyt: 763e9b8f8fbfc17c258b286bc6fc18d96d001874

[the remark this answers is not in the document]
bilbeyt: then you wouldnt need the last commit actually, same for true the commit before changing the read command start block

beamer/config/state.py
istankovic: We don't because this is already a live instance which was validated once upon loading. (It could happend that it is modified in an invalid way, but then we will get an exception, which is what we want anyway.)

beamer/config/state.py
istankovic: Same reasoning as above.

beamer/tests/config/util.py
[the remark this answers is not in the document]
istankovic: I am not sure I understand, could you please explain in more detail what you have in mind here?

fredo:
Addressed all my requested changes, I leave it up to you and Tolga, to resolve the last comments. nice job overall!

beamer/config/commands.py
bilbeyt: :D like this 
`
yield from helper(
        "FillManager",
        current_config.fill_manager.whitelist,
        desired_config.fill_manager.whitelist,
    )
`

It looks just weird like this :D but ok

beamer/tests/config/util.py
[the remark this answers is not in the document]
bilbeyt: I was thinking, you could add get_root_repo in https://github.com/beamer-bridge/beamer/commit/763e9b8f8fbfc17c258b286bc6fc18d96d001874 and do the changes in other commits as well, also the small start_block change can be moved to de7f8032c60bd18e20810649f4feb51fc50df241 or 9bf96301a2a1172e25569e54c3ee426789f1101e.

beamer/tests/config/test_write.py
```diff
@@ -0,0 +1,201 @@
...
+@pytest.fixture
+def deployment_objects(tmp_deployment_path, deployer, token):
+    rpc_file, artifact = deploy(deployer, tmp_deployment_path)
+    deployment = Deployment.from_file(artifact)
+    assert deployment.chain is not None
+
+    # Call updateToken so that the eventual 'config read' command can pick up the token.
+    address = deployment.chain.contracts["RequestManager"].address
+    request_manager: Any = ape.project.RequestManager.at(address)  # type: ignore
+    request_manager.updateToken(token.address, 0, 0, sender=deployer)
```
bilbeyt, line 68: just fyi, you dont need to write sender=deployer as deployer is the default sender for ape operations.

beamer/tests/config/util.py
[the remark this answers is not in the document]
istankovic: > I was thinking, you could add get_root_repo in [763e9b8](https://github.com/beamer-bridge/beamer/commit/763e9b8f8fbfc17c258b286bc6fc18d96d001874) and do the changes in other commits as well

The separation here is intentional -- I didn't want to combine the two unrelated changes (move of code, introduction of a new function) into the same commit. In general, I think that's not a good thing because it makes it harder to review, understand the commit changes and provide good commit messages. I left this part as is.

>  also the small start_block change can be moved to [de7f803](https://github.com/beamer-bridge/beamer/commit/de7f8032c60bd18e20810649f4feb51fc50df241) or [9bf9630](https://github.com/beamer-bridge/beamer/commit/9bf96301a2a1172e25569e54c3ee426789f1101e).

Nicely spotted! I folded the start_block change into 9bf9630 and updated the commit message appropriately.

beamer/tests/config/test_write.py
istankovic: Removed, thanks!

bilbeyt (approved):
Looks good to me!

istankovic:
Thanks @bilbeyt and @fredo for the review! :heart:

Issue #1947, opened 2023-06-15
Agent needs to check that he has enough balance to claim a fill


fredo:
The agent does not check before filling if he has enough funds on the target side to claim (I neglect the fact of challenging on purpose, since claiming is much more time critical then challenging).

solution:
Check whether there are enough funds on source side to be able to claim. We could introduce another threshold value when an agent would stop providing the service (currently this threshold is 0, implicitly). For the case of L1, this threshold and the L1 resolution threshold would simply add up.

andfletcher:
Default is non "0"

PR #1953, opened 2023-06-16
Introduce minimum source balance

fredo:
Closes: #1947

### DIFF
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -9,6 +9,7 @@ from typing import Callable
 import requests
 import structlog
 from web3 import HTTPProvider, Web3
+from web3._utils.empty import Empty
 from web3.contract import Contract
 from web3.types import Timestamp, Wei
 
@@ -347,9 +348,26 @@ def fill_request(request: Request, context: Context) -> None:
         context.logger.info("Request expired, ignoring", request=request)
         request.ignore()
         return
-    w3 = context.fill_manager.w3
-    token = w3.eth.contract(abi=get_ERC20_abi(), address=request.target_token_address)
-    address = w3.eth.default_account
+
+    source_web3 = context.request_manager.w3
+    source_address = source_web3.eth.default_account
+    assert not isinstance(source_address, Empty)
+    source_balance = source_web3.eth.get_balance(source_address)
+    min_source_balance = context.config.min_source_balance_per_chain.get(
+        context.source_chain.name, context.config.min_source_balance
+    )
+    if source_balance < min_source_balance:
+        context.logger.info(
+            "Not enough balance to claim, ignoring",
+            request=request,
+            min_source_balance=min_source_balance,
+            source_balance=source_balance,
+        )
+        return
+
+    target_web3 = context.fill_manager.w3
+    token = target_web3.eth.contract(abi=get_ERC20_abi(), address=request.target_token_address)
+    address = target_web3.eth.default_account
     balance = token.functions.balanceOf(address).call()
     if balance < request.amount:
         context.logger.info(
@@ -369,7 +387,7 @@ def fill_request(request: Request, context: Context) -> None:
             request_amount=request.amount,
         )
         return
-    chain_id = ChainId(w3.eth.chain_id)
+    chain_id = ChainId(target_web3.eth.chain_id)
     token_address = token.address
     mutex = context.fill_mutexes[(chain_id, token_address)]
     with mutex:
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -160,7 +160,7 @@ class EventProcessor:
         self._num_syncs_done = 0
         self._context = context
         self._rpc_working = True
-        self._chain_ids = {self._context.source_chain_id, self._context.target_chain_id}
+        self._chain_ids = {self._context.source_chain.id, self._context.target_chain.id}
 
     @property
     def context(self) -> Context:
@@ -453,11 +453,11 @@ def maybe_challenge(claim: Claim, context: Context) -> bool:
 
     if request.fill_timestamp is not None:
         finalized = _timestamp_is_l1_finalized(
-            request.fill_timestamp, context, context.target_chain_id
+            request.fill_timestamp, context, context.target_chain.id
         )
     elif claim.invalidation_timestamp is not None:
         finalized = _timestamp_is_l1_finalized(
-            claim.invalidation_timestamp, context, context.target_chain_id
+            claim.invalidation_timestamp, context, context.target_chain.id
         )
     if not finalized:
         return False
@@ -575,7 +575,7 @@ def _timestamp_is_l1_finalized(
 
 def maybe_prove(claim: Claim, context: Context) -> None:
     # mainnet: 10, goerli: 420, local: 901
-    if context.target_chain_id not in (10, 420, 901):
+    if context.target_chain.id not in (10, 420, 901):
         claim.message_proved = True
         return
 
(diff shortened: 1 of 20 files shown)


beamer/agent/data/agent.conf.example
```diff
@@ -15,6 +16,7 @@ prometheus-port = 9101
 [base-chain]
 rpc-url = "GOERLI_TESTNET_RPC_URL"
 poll-period = 60.0
```
bilbeyt, line 18: Config seems wrong to me, base-chain shouldnt have any options like poll-period or min-source-balance instead we should use normal chains.

beamer/agent/config.py
```diff
@@ -26,6 +27,8 @@ class Config:
     token_checker: TokenChecker
     fill_wait_time: int
     unsafe_fill_time: int
+    min_source_balance: int
+    min_source_balance_per_chain: dict[str, int]
```
bilbeyt: we can change this to dict[ChainId, int]

beamer/agent/chain.py
```diff
@@ -419,7 +419,7 @@ def fill_request(request: Request, context: Context) -> None:
         context.logger.info(
             "Filled request",
             request=request,
-            txn_hash=receipt.transactionHash.hex(),
+            txn_hash=receipt.transactionHash.hex(),  # type: ignore
```
bilbeyt, line 421: do we really need ignores ?

beamer/agent/chain.py
```diff
@@ -779,5 +781,5 @@ def _invalidate(request: Request, claim: Claim, context: Context) -> None:
         request=request.id,
         fill_id=claim.fill_id,
         claim=claim.id,
-        txn_hash=receipt.transactionHash.hex(),
+        txn_hash=receipt.transactionHash.hex(),  # type: ignore
```
istankovic, line 783: Eh, I would prefer a way to avoid sprinkling these `# type: ignore` all over.
In general, I like the change to move to `TxReceipt` for `transact`, but I think we should search for a better solution, possibly after merging this.

beamer/agent/chain.py
```diff
@@ -353,9 +353,8 @@ def fill_request(request: Request, context: Context) -> None:
     source_address = source_web3.eth.default_account
     assert not isinstance(source_address, Empty)
     source_balance = source_web3.eth.get_balance(source_address)
-    min_source_balance = context.config.min_source_balance_per_chain.get(
-        context.source_chain.name, context.config.min_source_balance
-    )
+    min_source_balance = context.config.chains[context.source_chain.name].min_source_balance
```
istankovic, line 356: Would this work if an agent sees a request for chains that were not configured in its config file?

beamer/tests/conftest.py
```diff
@@ -219,22 +219,25 @@ def config(request_manager, fill_manager, token, token_list):
...
         prometheus_metrics_port=None,
-        base_chain_rpc_url=URL(url),
+        base_chain_rpc_url=url,
         log_level="debug",
-        poll_period=0.5,
-        poll_period_per_chain={},
+        chains=chains,
     )
     beamer.agent.metrics.init(config=config, source_rpc_url=url, target_rpc_url=url)
     return config
```
istankovic, line 243: I think we should move this agent-related function to under `beamer/tests/agent` i.e. in the agent-specific `conftest.py` there.

beamer/tests/agent/test_request.py
```diff
@@ -418,3 +419,32 @@ def test_handling_reorgs(config, direction, token, request_manager):
             sleeper.sleep(0.1)
 
     agent.stop()
+
+
+def test_no_source_balance_for_fill(direction, config, token, request_manager):
+    requester, target = alloc_accounts(2)
+    min_source_balance = to_wei(1, "ether")
+    config.chains["l2a"].min_source_balance = min_source_balance
+    ape.chain.provider._make_request("evm_setAccountBalance", [config.account.address, "0x0"])
```
istankovic, line 428: Could you please avoid using a private method of the `provider`?

docs/source/configuration.rst
```diff
@@ -244,6 +246,22 @@ Options reference
...
+        min-source-balance = ETH
+
+     - Minimum ETH balance on source chain to fill requests on target chain..
+       The value applies to all chains that don't have the chain-specific min-source-balance defined.
+       Default: ``0.1``.
+
+   * - ::
+
+        [chains.NAME]
+        poll-period = ETH
```
istankovic, line 260: Is this a typo? Shouldn't this be `min-source-balance = ETH`?

docs/source/configuration.rst
```diff
@@ -244,6 +246,22 @@ Options reference
...
+
+     - Minimum ETH balance on source chain to fill requests on target chain.
```
istankovic, line 262: Nitpick: I would make this slightly clearer, e.g.
```
     - Minimum ETH balance on chain ``NAME`` to fill requests originating from it.
```

beamer/agent/chain.py
fredo: no, fill request is called in process_requests which only runs over request objects. If a fill is for an invalid chain pair it's going to be dropped before processing this event.

docs/source/configuration.rst
bilbeyt: done

docs/source/configuration.rst
bilbeyt: thanks for noticing! changed.

beamer/tests/agent/test_request.py
bilbeyt: done

beamer/tests/conftest.py
bilbeyt: moved all fixtures related with agent to its own conftest

beamer/agent/util.py
```diff
@@ -1,13 +1,42 @@
...
+        provider = cast(HTTPProvider, self.w3.provider)
+        assert provider.endpoint_uri is not None
+        return URL(provider.endpoint_uri)
+
+
+@dataclass
+class Chain(BaseChain):
+    name: str
+    tokens: list[tuple[ChainId, ChecksumAddress]]
+    _contracts: dict[str, Contract]
```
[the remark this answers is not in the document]
istankovic: > I believe I did this, because it was never accessed outside, and actually shouldn't.

Okay, we still have one external access to this field, namely during instance construction (see my comment above).
We should fix that.

beamer/tests/agent/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@pytest.fixture
+def setup_relayer_executable():
+    relayer = get_relayer_executable()
+    relayer.parent.mkdir(parents=True, exist_ok=True)
+    relayer.write_text("#!/bin/sh\nsleep 5\necho 'hello'")
+    relayer.chmod(relayer.stat().st_mode | stat.S_IEXEC)
+    yield
+    relayer.unlink()
```
istankovic, line 80: Thank you for moving this to the proper place!

docs/source/configuration.rst
```diff
@@ -244,6 +246,22 @@ Options reference
...
+     - Minimum ETH balance on source chain to fill requests on target chain..
+       The value applies to all chains that don't have the chain-specific min-source-balance defined.
+       Default: ``0.1``.
+
+   * - ::
+
+        [chains.NAME]
+        min-source-balance = ETH
+
+     - Minimum ETH balance on chain ``NAME`` to fill requests originating from it.
```
istankovic, line 262: Nitpick: seeing that we refer to `NAME` as `NAME` (without double backticks) elsewhere, I would also remove the backticks here, to be consistent.

beamer/agent/util.py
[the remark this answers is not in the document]
bilbeyt: I guess @istankovic right here, with a normal class and setter/getter approach we could have a private contracts, but with current setup it is not ideal, so will change this now.

istankovic (approved):

PR #1955, opened 2023-06-16
.circleci: update container image workflow deps

GabrielBuragev:
Fixes the incorrect dependency definitions of the `build-and-publish-container-image` job inside the `relayer` workflow. 

The current definition of the `relayer` workflow results in the following flow: https://app.circleci.com/pipelines/github/beamer-bridge/beamer/3132/workflows/2b23d0ef-783f-422e-9447-b0439830029c

It should instead be looking like: https://app.circleci.com/pipelines/github/beamer-bridge/beamer/3127/workflows/c7c203e2-d67d-454d-8db8-b116bd0b4558

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -589,9 +589,11 @@ workflows:
             - create-relayers
       - build-and-publish-container-image:
           requires:
-            - install-python
-            - create-relayers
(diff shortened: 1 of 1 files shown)


istankovic (approved):
Good catch!

PR #1959, opened 2023-06-19
Relayer: Recover l1 transaction hash in Arbitrum relayer service

### LINKED ISSUE #1693: Relayer: Prevent deposits when message was already relayed but not yet finalized
# Rationale

The current relayer design has no prior knowledge if the message has already been relayed. In some cases, this leads to redundant multiple deposits in the Messenger contract. 

### Example:
Direction: Any rollup -> Arbitrum

For a message that has already been relayed, any future attempt to relay the **same message** leads to the relayer depositing additional funds into the ArbitrumL1Messenger contract. 
Instead of this, the relayer should skip the `prepare` (depositing) step and move on to the next steps.

# Acceptance criteria
- [ ] Relayer checks if message has already been relayed and skips the logic for preparing (depositing) and relaying the message

### LINKED ISSUE #1911: Relayer: ARB service needs to be able to recover l1 transaction hash
The problem appears only for relaying messages from ARB to any other network that requires an implemented `finalize` step. It is present only in the following case:

Lets take a message relay from ARB to PolygonZKEvm as an example. The relayer needs to:
- Prepare & make sure the Polygon side is ready for a succcesful relay (step 1 - `prepare`). 
**This is currently a no-op function** as Polygon doesn't require any preparation (unlike ARB that needs a user deposit for paying out the ticket creation)
- Read the message on ARB and execute it on L1 via the ARB Outbox (step 2 - `executeOnL1`)
- Wait until message reached Polygon's side via the Resolver and PolygonZKEvmL1Messenger. 
Once the message is available for claiming on PolygonZKEvm's side, the relayer claims it and marks the whole process as done. (step 3 `finalize`)

This all works fine as long as the program goes uninterrupted by an agent shutdown or an unforeseen crash in the relayer.
If for some reason the relayer gets interrupted after finishing point 2 above (step 2 - `executeOnL1`) and before finishing with point 3 above (step 3 - `finalize`), the relayer will not be able to recover the state when called again. It loses track of the L1 transaction (hash) that was executed as a part of the second step `executeOnL1` which is needed in the third step `finalize` and therefore is not able to proceed with relaying to the destination (finalizing).

This only comes apparent once we add support for PolygonZKEvm which has a `finalize` step in the direction ARB -> PolygonZKEvm.

# Acceptance Criteria
- [ ] Implement a way to recover the transaction hash from a previous failed relay attempt

manuelwedler:
closes #1911 
closes #1693

I reused the logic for recovering the transaction hash of the Ethereum service. Could actually be used by any chain.

### DIFF
--- a/relayer/src/cli/programs/relay.ts
+++ b/relayer/src/cli/programs/relay.ts
@@ -64,18 +64,20 @@ export class RelayerProgram {
     );
     let l1TransactionHash: string;
 
-    if (relayStepCompleted === false) {
+    if (relayStepCompleted) {
+      l1TransactionHash = await this.l2RelayerFrom.relayTxToL1Step.recoverL1TransactionHash(
+        this.l2TransactionHash,
+      );
+    } else {
       if (this.l2RelayerTo.prepareStep) {
         const prepareStepCompleted = await this.l2RelayerTo.prepareStep.isCompleted();
 
-        if (prepareStepCompleted === false) {
+        if (!prepareStepCompleted) {
           await this.l2RelayerTo.prepareStep.execute();
         }
       }
 
       l1TransactionHash = await this.l2RelayerFrom.relayTxToL1Step.execute(this.l2TransactionHash);
-    } else {
-      l1TransactionHash = relayStepCompleted;
     }
 
     if (this.l2RelayerTo.finalizeStep) {
@@ -83,7 +85,7 @@ export class RelayerProgram {
         l1TransactionHash,
       );
 
-      if (finalizeStepCompleted === false) {
+      if (!finalizeStepCompleted) {
         await this.l2RelayerTo.finalizeStep.execute(l1TransactionHash);
       }
     }
--- a/relayer/src/cli/programs/relay.ts
+++ b/relayer/src/cli/programs/relay.ts
@@ -20,7 +20,7 @@ export class RelayerProgram {
   ) {}
 
   static validateArgs(args: ProgramOptions): Array<string> {
-    const validationErrors = [];
+    const validationErrors: string[] = [];
 
     if (!args.l2TransactionHash.startsWith("0x") || args.l2TransactionHash.trim().length != 66) {
       validationErrors.push(
--- a/relayer/src/cli/programs/relay.ts
+++ b/relayer/src/cli/programs/relay.ts
@@ -1,13 +1,13 @@
 import { getNetworkId } from "../../common/network";
 import { createRelayer } from "../../services/relayer/map";
-import type { BaseRelayerService } from "../../services/types";
+import type { BaseRelayerService, TransactionHash } from "../../services/types";
 
 export type ProgramOptions = {
   l1RpcUrl: string;
   l2RelayFromRpcUrl: string;
   l2RelayToRpcUrl: string;
   walletPrivateKey: string;
-  l2TransactionHash: string;
+  l2TransactionHash: TransactionHash;
   networkFrom?: string;
   networkTo?: string;
 };
@@ -16,7 +16,7 @@ export class RelayerProgram {
   constructor(
     readonly l2RelayerFrom: BaseRelayerService,
     readonly l2RelayerTo: BaseRelayerService,
-    readonly l2TransactionHash: string,
+    readonly l2TransactionHash: TransactionHash,
   ) {}
 
   static validateArgs(args: ProgramOptions): Array<string> {
@@ -59,10 +59,33 @@ export class RelayerProgram {
   }
 
   async run(): Promise<void> {
-    await this.l2RelayerTo.prepare();
-    const l1TransactionHash = await this.l2RelayerFrom.relayTxToL1(this.l2TransactionHash);
-    if (l1TransactionHash) {
-      await this.l2RelayerTo.finalize(l1TransactionHash);
+    const relayStepCompleted = await this.l2RelayerFrom.relayTxToL1Step.isCompleted(
+      this.l2TransactionHash,
+    );
+    let l1TransactionHash: string;
+
+    if (relayStepCompleted === false) {
+      if (this.l2RelayerTo.prepareStep) {
+        const prepareStepCompleted = await this.l2RelayerTo.prepareStep.isCompleted();
+
+        if (prepareStepCompleted === false) {
+          await this.l2RelayerTo.prepareStep.execute();
+        }
+      }
+
+      l1TransactionHash = await this.l2RelayerFrom.relayTxToL1Step.execute(this.l2TransactionHash);
+    } else {
+      l1TransactionHash = relayStepCompleted;
+    }
+
+    if (this.l2RelayerTo.finalizeStep) {
+      const finalizeStepCompleted = await this.l2RelayerTo.finalizeStep.isCompleted(
+        l1TransactionHash,
+      );
+
+      if (finalizeStepCompleted === false) {
+        await this.l2RelayerTo.finalizeStep.execute(l1TransactionHash);
+      }
     }
   }
 }
--- a/relayer/src/services/relayer/arbitrum.ts
+++ b/relayer/src/services/relayer/arbitrum.ts
@@ -43,6 +43,7 @@ export class ArbitrumRelayerService extends BaseRelayerService {
   relayTxToL1Step = new RelayStep(
     async (l2TransactionHash) => await this.relayTxToL1(l2TransactionHash),
     async (l2TransactionHash) => await this.isRelayCompleted(l2TransactionHash),
+    async (l2TransactionHash) => await this.recoverL1TransactionHash(l2TransactionHash),
   );
   finalizeStep = new FinalizeStep(
     async (l1TransactionHash) => await this.finalize(l1TransactionHash),
@@ -73,17 +74,14 @@ export class ArbitrumRelayerService extends BaseRelayerService {
     return deltaSubmissionFee;
   }
 
-  private async isPrepareCompleted(): Promise<void | false> {
+  private async isPrepareCompleted(): Promise<boolean> {
     const deltaSubmissionFee = await this.getDeltaSubmissionFee();
 
     const currentDeposit: BigNumber = await this.arbitrumL1Messenger.deposits(
       this.l1Wallet.address,
     );
 
-    if (currentDeposit.lte(deltaSubmissionFee)) {
-      return false;
-    }
-    console.log("ArbitrumL1Messenger has enough funds to proceed!");
+    return currentDeposit.gte(deltaSubmissionFee);
   }
 
   private async prepare(): Promise<void> {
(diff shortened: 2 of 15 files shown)


relayer/src/services/relayer/ethereum.ts
```diff
@@ -128,64 +55,62 @@ export class EthereumRelayerService extends BaseRelayerService {
...
-
-    if (!callParameters) {
-      throw new Error(
-        "Couldn't find a matching event (RequestFilled | FillInvalidated) in the transaction logs.",
-      );
+  private async isRelayCompleted(
+    l1TransactionHash: TransactionHash,
+  ): Promise<TransactionHash | false> {
+    if (!this.callParameters) {
+      this.callParameters = await this.parseFillEventDataFromTxHash(l1TransactionHash);
```
GabrielBuragev, line 62: Same as in `arbitrum.ts`. Lets declare this variable locally in the scope of this function instead of the class scope.

relayer/src/services/relayer/optimism.ts
```diff
@@ -157,29 +185,12 @@ export class OptimismRelayerService extends BaseRelayerService {
...
 
-  async prepare(): Promise<boolean> {
-    return true;
-  }
-
-  async relayTxToL1(l2TransactionHash: TransactionHash): Promise<TransactionHash | undefined> {
+  private async relayTxToL1(l2TransactionHash: TransactionHash): Promise<TransactionHash> {
     console.log("Optimism outbox execution.");
-
-    await this.l2RpcProvider.waitForTransaction(l2TransactionHash, 1);
```
GabrielBuragev, line 167: I think we can keep this line just in case the `relayTxToL1` step is being called without a prior call to `isRelayCompleted`.

relayer/src/services/relayer/polygon-zkevm.ts
```diff
@@ -107,45 +122,73 @@ export class PolygonZKEvmRelayerService extends BaseRelayerService {
...
-    // 2. Fetch message data
-    const message = await this.getMessageInfoSafe(
-      bridgeEventParameters.depositCount,
-      bridgeEventParameters.originNetwork,
-    );
+  private async isRelayCompleted(
+    l2TransactionHash: TransactionHash,
+  ): Promise<TransactionHash | false> {
+    if (!this.relayBridgeEventParameters) {
+      this.relayBridgeEventParameters = await this.getRelayBridgeEventParameters(
```
GabrielBuragev, line 147: Same here as in `arbitrum.ts` service. Declare the variable inside the scope of this function instead.

relayer/src/services/relayer/arbitrum.ts
```diff
@@ -145,24 +226,40 @@ export class ArbitrumRelayerService extends BaseRelayerService {
...
+    }
 
     console.log("Waiting for L2 side. It may take 10-15 minutes ⏰⏰");
-    const messageResult = await message.waitForStatus();
+    const messageResult = await this.l1ToL2Message.waitForStatus();
 
     const status = messageResult.status;
     if (status === L1ToL2MessageStatus.REDEEMED) {
       console.log(`L2 retryable txn executed ${messageResult.l2TxReceipt.transactionHash}`);
       return;
```
GabrielBuragev, line 243: `return false;` instead?

GabrielBuragev (changes requested):
Well done man! The much needed refactoring of the relayer .. you nailed it. 
I have a few improvement suggestions.

In order to speed up the review process, i already addressed a few of the nitpicks by myself. Check the fixup commit and let me know if that is fine with you.

GabrielBuragev:
Btw, i tested this branch by sending fill invalidations in all directions (for ARB, OP, ETH, POLY)  and it all worked flawless!

relayer/src/services/relayer/arbitrum.ts
```diff
@@ -30,61 +36,101 @@ const L1_CONTRACTS: Record<number, { ARBITRUM_L1_MESSENGER: string }> = {
...
-        value: deltaSubmissionFee,
-      });
+    return deltaSubmissionFee;
+  }
 
-      const transaction = await arbitrumL1Messenger.deposit({
-        gasLimit: estimatedGasLimit,
-        value: deltaSubmissionFee,
-      });
+  private async isPrepareCompleted(): Promise<void | false> {
```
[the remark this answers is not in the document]
manuelwedler: The reason is that the `isCompleted` function of the relay step needs to return a value. I agree that a simple boolean would be much nicer. One option could be to separate the recovering of the hash into another function. Since this would be a bigger change, I will  try to fix the rest first and see if we can make this today.

relayer/src/services/relayer/arbitrum.ts
```diff
@@ -30,61 +36,101 @@ const L1_CONTRACTS: Record<number, { ARBITRUM_L1_MESSENGER: string }> = {
...
-          `Successfully deposited funds on ArbitrumL1Messenger: ${receipt.transactionHash}`,
-        );
-        return true;
-      } else {
-        throw new Error("Arbitrum L1 deposit transaction reverted on chain.");
-      }
-    } else {
-      console.log("ArbitrumL1Messenger has enough funds to proceed!");
+  private async prepare(): Promise<void> {
+    if (!this.deltaSubmissionFee) {
```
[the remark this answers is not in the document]
manuelwedler: Didn't think of this. Thanks for noticing! Will change it :)

relayer/src/services/relayer/arbitrum.ts
```diff
@@ -30,61 +36,101 @@ const L1_CONTRACTS: Record<number, { ARBITRUM_L1_MESSENGER: string }> = {
...
+    });
 
+    const transaction = await this.arbitrumL1Messenger.deposit({
+      gasLimit: estimatedGasLimit,
+      value: this.deltaSubmissionFee,
+    });
+
+    await transaction.wait();
+
+    const receipt = await this.arbitrumL1Messenger.provider.waitForTransaction(
```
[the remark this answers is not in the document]
manuelwedler: :heavy_check_mark:

relayer/src/services/relayer/arbitrum.ts
```diff
@@ -100,38 +146,73 @@ export class ArbitrumRelayerService extends BaseRelayerService {
...
     const messages = await l2Receipt.getL2ToL1Messages<Signer>(this.l1Wallet);
     const l2ToL1Msg = messages[0];
+    return l2ToL1Msg;
+  }
+
+  private async isRelayCompleted(
+    l2TransactionHash: TransactionHash,
+  ): Promise<TransactionHash | false> {
+    if (!this.l2ToL1Message) {
+      this.l2ToL1Message = await this.getL2ToL1Message(l2TransactionHash);
```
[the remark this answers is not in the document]
manuelwedler: Well, I followed this pattern in all services to save rpc calls, exactly as you said. When I started to refactor, I wanted to keep the number of external calls the same. 

I totally agree that it's not perfect like this and will make the change.

relayer/src/services/relayer/arbitrum.ts
```diff
@@ -100,38 +146,73 @@ export class ArbitrumRelayerService extends BaseRelayerService {
...
+      }
+      console.log(`Message has already been relayed with tx hash: ${transactionHash}.\n`);
+      return transactionHash;
     }
+    return false;
+  }
+
+  private async relayTxToL1(l2TransactionHash: TransactionHash): Promise<string> {
+    if (!this.l2ToL1Message) {
+      this.l2ToL1Message = await this.getL2ToL1Message(l2TransactionHash);
```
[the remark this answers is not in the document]
manuelwedler: :heavy_check_mark:

relayer/src/services/relayer/arbitrum.ts
```diff
@@ -145,24 +226,40 @@ export class ArbitrumRelayerService extends BaseRelayerService {
     const wrappedL1TransactionReceipt = new L1TransactionReceipt(l1TransactionReceipt);
     const messages = await wrappedL1TransactionReceipt.getL1ToL2Messages(this.l2Wallet);
     const message = messages[0];
+    return message;
+  }
+
+  private async isFinalizeCompleted(l1TransactionHash: string): Promise<void | false> {
+    if (!this.l1ToL2Message) {
+      this.l1ToL2Message = await this.getL1ToL2Message(l1TransactionHash);
```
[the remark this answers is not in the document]
manuelwedler: :heavy_check_mark:

relayer/src/services/relayer/arbitrum.ts
manuelwedler: If it is redeemed, the step is completed and we return the default return value of the step (`void`).

relayer/src/services/relayer/arbitrum.ts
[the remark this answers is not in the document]
GabrielBuragev: Thats a good approach you took! 
But looking at the relayer as a service that is called not so often (hopefully), we don't need to worry too much on optimizing on RPC calls, so lets try to adopt a nice re-usable interface to make our lifes easier.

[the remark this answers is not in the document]
manuelwedler: I fully agree :)

relayer/src/services/relayer/boba.ts
```diff
@@ -65,17 +83,27 @@ export class BobaRelayerService extends BaseRelayerService {
       throw new Error("Message not ready for relaying.");
     }
 
+    return false;
+  }
+
+  private async relayTxToL1(l2TransactionHash: TransactionHash): Promise<TransactionHash> {
+    if (!this.message) {
+      this.message = await this.getMessage(l2TransactionHash);
```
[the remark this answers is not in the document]
manuelwedler: :heavy_check_mark:

relayer/src/services/relayer/boba.ts
```diff
@@ -65,17 +83,27 @@ export class BobaRelayerService extends BaseRelayerService {
       throw new Error("Message not ready for relaying.");
```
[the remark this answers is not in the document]
manuelwedler: :heavy_check_mark:

relayer/src/services/relayer/ethereum.ts
manuelwedler: :heavy_check_mark:

relayer/src/services/relayer/optimism.ts
manuelwedler: :heavy_check_mark:

relayer/src/services/relayer/polygon-zkevm.ts
manuelwedler: :heavy_check_mark:

relayer/src/services/relayer/arbitrum.ts
GabrielBuragev: aah, true. My bad

relayer/src/services/relayer/arbitrum.ts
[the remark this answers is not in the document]
GabrielBuragev: Okay, now i actually see what the problem was. Sorry that you had to go that far to fix such a small thing.
On one hand, the separation of the steps is much cleaner and easily understandable now, but we introduced 1 extra level of rpc calls but we shouldn't be too worried about it for now as this program is being ran rarely.

relayer/src/cli/programs/relay.ts
```diff
@@ -59,10 +59,35 @@ export class RelayerProgram {
...
+
+    if (relayStepCompleted) {
+      l1TransactionHash = await this.l2RelayerFrom.relayTxToL1Step.recoverL1TransactionHash(
+        this.l2TransactionHash,
+      );
+    } else {
+      if (this.l2RelayerTo.prepareStep) {
+        const prepareStepCompleted = await this.l2RelayerTo.prepareStep.isCompleted();
+
+        if (prepareStepCompleted === false) {
```
GabrielBuragev, line 75: We could now change this to `if(!prepareStepCompleted)`

relayer/src/cli/programs/relay.ts
```diff
@@ -59,10 +59,35 @@ export class RelayerProgram {
...
+
+      l1TransactionHash = await this.l2RelayerFrom.relayTxToL1Step.execute(this.l2TransactionHash);
+    }
+
+    if (this.l2RelayerTo.finalizeStep) {
+      const finalizeStepCompleted = await this.l2RelayerTo.finalizeStep.isCompleted(
+        l1TransactionHash,
+      );
+
+      if (finalizeStepCompleted === false) {
```
GabrielBuragev, line 88: Same as above

GabrielBuragev (approved):
Two more nitpicks and its good to be merged.
Thanks for addressing all the suggestions!

relayer/src/cli/programs/relay.ts
manuelwedler: Are true! thanks!

manuelwedler:
Just addressed them. Finally ready to merge!

PR #1961, opened 2023-06-20
relayer: use resolver block number per chain

GabrielBuragev:
Previously, the ethereum relayer service was only able to recover the L1 transaction hash for relays that happened on Mainnet as it was only using the `RESOLVER_DEPLOY_BLOCK_NUMBER` for L1 Mainnet.

This PR adds the capability of recovering the L1 transaction hashes both on Testnet and Local net.

Note: This is properly addressed in https://github.com/beamer-bridge/beamer/pull/1804 by pulling the deployment infos from the npm package instead of having them hardcoded, so consider this a quick & temporary fix.

### DIFF
--- a/relayer/src/services/relayer/ethereum.ts
+++ b/relayer/src/services/relayer/ethereum.ts
@@ -14,21 +14,25 @@ type RelayCallParams = {
   filler: string;
 };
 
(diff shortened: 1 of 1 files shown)


fredo (approved):
LGTM

PR #1966, opened 2023-06-21
frontend: Allow Matomo cookie inside iframe

### LINKED ISSUE #1935: frontend: Cookie consent popup decision is not stored when accepted in Safe App
> Good morning! QA from safe apps team here 😄
>
> I see that the privacy policy pop up will only disappear when it is declined, might be worth a 👀 : 
> 
> https://github.com/5afe/safe-apps-list/assets/92332710/843e7ba8-4838-41ba-bf5b-bbc16b8d0568
> 
> Cheers!

_Originally posted by @JagoFigueroa in https://github.com/5afe/safe-apps-list/issues/253#issuecomment-1587066567_

manuelwedler:
Closes #1935 

When the Beamer app is loaded in an iframe (this is for example the case with Safe), Matomo's `setCookieSameSite` must be set to `None` in order to use the consent cookie.

Inform the Safe team in the issue once this is merged.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -47,6 +47,7 @@
     <!-- Matomo -->
     <script>
       var _paq = (window._paq = window._paq || []);
+      _paq.push(['setCookieSameSite', 'None']);
       /* require user tracking consent before processing data */
       _paq.push(['requireConsent']);
       /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
@@ -60,7 +61,7 @@
           g = d.createElement('script'),
           s = d.getElementsByTagName('script')[0];
         g.async = true;
-        g.src = '//cdn.matomo.cloud/beamerbridge.matomo.cloud/matomo.js';
+        g.src = 'https://cdn.matomo.cloud/beamerbridge.matomo.cloud/matomo.js';
         s.parentNode.insertBefore(g, s);
       })();
     </script>


GabrielBuragev:
Do you think this opens the door for vulnerabilities like Cross-Site-Request-Forgery?
I only found this so far: https://cwe.mitre.org/data/definitions/1275.html but it describes the vulnerability when using sensitive cookies. I guess our cookies are not sensitive so we should be fine there.

One thing we have to check is the visualization of the traffic from Safe in the Matomo dashboard. Lets communicate with the marcomms team on this and test it out after we make a new deployment.

manuelwedler:
Well, the threat would be that someone forges website visits on another domain. But there is not a real value for an attacker I would say. 

The only other option here would be to disable Matomo tracking in Safe, but I think it would be interesting to see how many users are using the Safe integration.

GabrielBuragev (approved):
Looks good!

GabrielBuragev:
> Well, the threat would be that someone forges website visits on another domain. But there is not a real value for an attacker I would say.
> 
> The only other option here would be to disable Matomo tracking in Safe, but I think it would be interesting to see how many users are using the Safe integration.

Lets keep it this way. I will communicate with @DBecane and see how the traffic shows up on Matomo. If we need to further optimize it, we can do it apart from this issue.

PR #1967, opened 2023-06-21
frontend: Add PolygonZKEvm support

### LINKED ISSUE #1775: Support Polygon ZKEVM on frontend
# Rationale
We want user to be able to bridge to and from Polygon ZKEVM via our frontend for this we need to :
# Implementation

- [ ] Add chainconfig and tokens are added to the file

GabrielBuragev:
Closes #1775 

Also, additionally adds a way to hide certain networks before we officially announce/launch. 
This means that once we are ready to announce Polygon ZKEvm support, we just need to flip the value of the boolean flag `hidden` (or remove it completely) inside the chain config file.


frontend/config/chains/1101.json
```diff
@@ -0,0 +1,15 @@
+{
+  "identifier": 1101,
```
manuelwedler, line 2: Don't we also need a config for the polygon goerli testnet?

manuelwedler (commented):
Looks good! Just wondering why there is no configuration for goerli.

frontend/config/chains/1101.json
GabrielBuragev: Good point. Will add it.

manuelwedler (approved):
Perfect!

PR #1968, opened 2023-06-22
circleci: new e2e-test workflow

bilbeyt:


.circleci/workflows.yml
```diff
@@ -575,13 +570,17 @@ workflows:
           requires:
             - compile-contracts
             - create-relayers
-      - build-and-publish-container-image:
+      - test-container-image:
+          requires:
+            - compile-contracts
+            - create-relayers
+      - publish-container-image:
```
istankovic, line 577: I would actually move this (`publish-container-image`) out into its own workflow, and make it depend on all the other workflows.

The way this is now, we could theoretically publish a container image even if there is a failure in one of the agent and contract tests, which is not nice. Moving the publish part into its own workflow and making it depend on all tests should solve that.

istankovic (changes requested):
Very good overall, with one concern, as commented.

scripts/check-pipeline.py
```diff
@@ -0,0 +1,83 @@
...
+                if workflow.status not in ("success", "running")
+            ]
+            if len(failed_workflows) > 0:
+                print("There was at least one workflow failed")
+                sys.exit(1)
+            else:
+                print("Pipeline was successful")
+                sys.exit(0)
+        elif len(running_workflows) > 1:
+            print("There are running pipelines")
```
istankovic, line 79: Possible typo: `pipelines` -> `workflows`?

bilbeyt: pipeline consists of workflows and i am looking for statuses of workflows in a pipeline

bilbeyt: but verbose part can change ofc :D

scripts/check-pipeline.py
```diff
@@ -0,0 +1,83 @@
...
+                if workflow.status not in ("success", "running")
+            ]
+            if len(failed_workflows) > 0:
+                print("There was at least one workflow failed")
+                sys.exit(1)
+            else:
+                print("Pipeline was successful")
+                sys.exit(0)
+        elif len(running_workflows) > 1:
+            print("There are running workflows")
```
istankovic, line 79: Shouldn't we sleep inside this loop, at least a bit, e.g. at least 1s?
Otherwise we're going to be constantly issuing requests, which is not nice.

scripts/check-pipeline.py
```diff
@@ -0,0 +1,83 @@
...
+    headers = {
+        "Circle-Token": api_token,
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+    }
+    page_token = None
+    url = _WORKFLOW_URL.format(workflow_id=workflow_id)
+    res = requests.get(url, headers=headers)
+    data = res.json()
+    pipeline_id = data["pipeline_id"]
```
istankovic, line 48: Can we not get the `pipeline_id` directly via `sys.argv`, instead of `workflow_id`?

.circleci/workflows.yml
```diff
@@ -376,7 +376,29 @@ jobs:
...
           command: yarn test:unit
 
-  build-and-publish-container-image:
+  check-pipeline:
+    executor: vm
+    steps:
+      - initialize-environment
+      - run:
+          name: Check pipeline
+          command: poetry run python scripts/check-pipeline.py $CIRCLE_WORKFLOW_ID
```
istankovic, line 385: We could have something like
```
  environment:
    PIPELINE_ID: << pipeline.id >>
```
and then
```
          command: poetry run python scripts/check-pipeline.py $PIPELINE_ID
```

scripts/check-pipeline.py
bilbeyt: basically no, https://circleci.com/docs/variables/#built-in-environment-variables

.circleci/workflows.yml
bilbeyt: will try this but not sure if this is valid

scripts/check-pipeline.py
bilbeyt: will add

.circleci/workflows.yml
```diff
@@ -573,13 +577,31 @@ workflows:
...
+            - compile-contracts
+            - create-relayers
+
+  check-and-publish:
+    when: 
+      and:
+        - or: 
+          - equal: [ true, <<pipeline.parameters.backend-updated>> ]
+          - equal: [ true, <<pipeline.parameters.relayer-updated>> ]
+        - equal: [ circleci/e2e_workflow, <<pipeline.git.branch>> ]
```
istankovic, line 591: I think this line needs to be modified before the final merge.

.circleci/workflows.yml
```diff
@@ -573,13 +577,31 @@ workflows:
...
-      - build-and-publish-container-image:
+      - test-container-image:
           requires:
-            - test-relayer
-            - e2e-test-ethereum
-            - e2e-test-optimism
-            - e2e-test-arbitrum
-            - e2e-test-polygon-zkevm
+            - compile-contracts
+            - create-relayers
```
istankovic, line 583: Here we have that `test-container-image` depends on `compile-contracts`, but if you look at the definition of `test-container-image`, you will see we already compile contracts there:
```
  test-container-image:                                         
    executor: vm                                                
    environment:                                                
      IMAGE_NAME: ghcr.io/beamer-bridge/beamer                  
    steps:                                                      
      - initialize-environment                                  
      - run:                                                    
          name: Build a beamer image                            
          command: make container-image IMAGE_NAME=${IMAGE_NAME}
      - run:                                                    
          name: Compile contracts                               
          command: ape compile                                  
```

Can this then be removed from the job?

scripts/check-pipeline.py
```diff
@@ -0,0 +1,86 @@
...
+    headers = {
+        "Circle-Token": api_token,
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+    }
+    # page_token = None
+    # url = _WORKFLOW_URL.format(workflow_id=workflow_id)
+    # res = requests.get(url, headers=headers)
+    # data = res.json()
+    # pipeline_id = data["pipeline_id"]
```
istankovic, line 49: Unnecessary comments.

scripts/check-pipeline.py
```diff
@@ -0,0 +1,86 @@
...
+            ]
+            if len(failed_workflows) > 0:
+                print("There was at least one workflow failed")
+                sys.exit(1)
+            else:
+                print("Pipeline was successful")
+                sys.exit(0)
+        elif len(running_workflows) > 1:
+            print("There are running workflows")
+            time.sleep(1)
```
istankovic, line 82: Why not move this `sleep` out of the `elif` branch so that we sleep always before the next request?

scripts/check-pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+            workflow for workflow in _STATUSES.values() if workflow.status == "running"
+        ]
+        # check-and-publish should be ignored
+        if len(running_workflows) == 1 and running_workflows[0].name == "check-and-publish":
+            failed_workflows = [
+                workflow
+                for workflow in _STATUSES.values()
+                if workflow.status not in ("success", "running")
+            ]
+            if len(failed_workflows) > 0:
```
istankovic, line 68: We should make use of container bool value here, i.e. `if failed_workflows:`

scripts/check-pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+    for workflow in items:
+        created_at = parser.parse(workflow["created_at"])
+        name = workflow["name"]
+        status = workflow["status"]
+        if name not in _STATUSES:
+            _STATUSES[name] = _WorkflowStatus(created_at, name, status)
+        elif _STATUSES[name].created_at == created_at and _STATUSES[name].status != status:
+            _STATUSES[name].status = status
+        elif _STATUSES[name].created_at < created_at:
+            status[name] = _WorkflowStatus(created_at, name, status)
```
istankovic, line 34: Nicely done!

istankovic (approved):
Left a few comments.
Looks very good, this is a big improvement to our CI, thanks for doing this!

Issue #1227, opened 2022-11-27
Frontend: Migrate to WalletConnect v2


GabrielBuragev:
# Rationale
The frontend currently uses WalletConnect as one of the main interfaces, besides MetaMask, for connecting to different wallet providers (both mobile & desktop).
So far everything works as expected by using WalletConnect v1 however this version of the library will soon become deprecated and it's sunset is already in progress.

dApps need to start migrating to WalletConnect v2 beginning of January.

## Announcement
https://medium.com/walletconnect/walletconnect-v1-0-sunset-notice-and-migration-schedule-8af9d3720d2e

# Implementation
We need to migrate to WalletConnect v2

# Acceptance Criteria
- [ ] Migrated to WalletConnect v2

andfletcher:
what will happen if we do not migrate? @GabrielBuragev short term / long term?

GabrielBuragev:
Well i am sure about the following things:
- WalletConnect v1.0 gets deprecated which means no more maintenance (bugs are not being fixed)
- Many new features in WalletConnect v2.0 which might be useful for our future UX
- WalletConnect v1.0 will be shut down on March 1, 2023 which means all Wallet providers already migrated fully to WalletConnect v2.0 and support for v1.0 is not promised.

I might've missed a point or two as i cannot really predict all the outcomes in the case of not upgrading to WalletConnect v2.0.
This issue was created so we have somewhat a reminder to address this announcement.

andfletcher:
Ok. So we will need to do this before March. Got it.

andfletcher:
blocked until official launch of v2 of wallet connect

andfletcher:
set a calander entry to remind us.

andfletcher:
@GabrielBuragev scope out what needs to be done here and then deal with #1956

GabrielBuragev:
Closed in https://github.com/beamer-bridge/beamer/pull/1972

PR #1972, opened 2023-06-22
frontend: Wallet Connect v2 + Proper Disconnect

### LINKED ISSUE #1463: Frontend: Allow users to fully disconnect from wallet provider
# Rationale
Some of the wallet providers we use offer different way of connecting to our app.

Lets take WalletConnect for example: 
image

Once the user successfully connects with one of these provided options by WalletConnect, he should be able to disconnect from that specific option and have the chance to choose another option with the same connector (WalletConnect). 

At this moment, the app doesn't support this. Once you successfully connected with one of the options given by the wallet provider you can only do a soft disconnect.
A soft disconnect means that the dapp cleans its state, but doesn't take care of cleaning the state that is being used by the connector which is most of the times stored inside the browser's `localStorage` (this might differ in different connectors).

# Why should this feature exist?
As a user, i would like to be able to do a hard reset of the connection by clicking on "Disconnect" so i can choose another sub-option. 

# What are the use-cases?
Users can freely roam between the options provided by each wallet provider.

# How it currently is

https://user-images.githubusercontent.com/18466333/217293225-28953146-2116-4788-a720-895209909e54.mov

# How it should be

https://user-images.githubusercontent.com/18466333/217293477-feb24cd3-1a1a-4ff6-bde8-29714e5bba28.mov

# Implementation
Since we are sticking to implementing our own abstractions around the wallet provider libraries (such as `@coinbase/wallet-sdk`, `@walletconnect/web3-provider`) for handling the communication with the respective wallet provider extensions/apps we need to also store/have a way to access the instances of these providers inside our abstracted classes. 
Our current design transforms the instances of these providers into a `ethers.providers.Web3Provider` which means we are losing type support for the native API of the corresponding wallet provider. 
Most of the times, the instances of these native providers offer methods for disconnecting (example: https://docs.cloud.coinbase.com/wallet-sdk/docs/disconnecting-links#disconnecting-a-link). 
This will clean the state they inserted into the `localStorage` and offer the users to connect again by selecting a different sub-option.  

# Acceptance Criteria
- [ ] After successful disconnect, when trying to connect with the same wallet provider users are prompted to make a selection

GabrielBuragev:
Closes #1463 #1227 

This PR contains the work from PR https://github.com/beamer-bridge/beamer/pull/1472 that is rebased on top of latest `main`.
Plenty of merge conflicts that had to be fixed so i had to analyze the code afterwards, fix tests, linter etc. which led to a few more commits. 

Besides the `fix/merge conflict` commits i also included Base testnet support which was laying on branch `beamer-v2` together with the "proper disconnect" commits (without the rebase/fixed merge conflicts and without the wallet connect v2 support). 

How we came to this twist: 
The moment when we started splitting these features all around the repo was when we did the "shadow" Base testnet deployment to surprise the team. From that moment on, the testnet.beamerbridge.com was pointing to this branch and it just got worse after that as soon as we started using this branch to also host all the wallet connect v1/v2 changes that were being requested/tested by the WalletConnect team (for getting listed on their dashboard).
Not sure why we simply didn't merge the Base testnet support to `main` branch after the "surprise" but i guess we can resolve this now once and for all 😄


compojoom:
I was able to reproduce the switch chain bug/issue with few modifications to the walletconnect example app.
I've posted a discussion in the walletconnect repo:
https://github.com/orgs/WalletConnect/discussions/2774

Right now I'm not sure what we are supposed to do. 
If we pass all chains we have in the app and one of those chains is not configured in the wallet - the connection will fail with an error - chain unknown without explanation which chain is unknown. 

If we try to connect just a single chain - switching between the chains doesn't work. And the connected wallet doesn't behave like metamask desktop does - it doesn't offer to add the chain config in the wallet.

GabrielBuragev:
So, after updating the version of the `@walletconnect/ethereum-provider` to `2.8.4` https://github.com/WalletConnect/walletconnect-monorepo/releases/tag/2.8.4 we were able to resolve one of our problems but on the other hand, welcomed other problems.


The problem that is fixed is related to the missing network switch events. These events are now properly propagated from the MetaMask app to our DApp so we can react and reload/pre-select the source chain selection. However, this also shows flaky behavior and slow response (sometimes it takes between 5-10 seconds for the event to be emitted).

In order to solve the problem on MetaMask we had to configure the wallet connect sdk in the following way:
```ts
const provider = await WalletConnect.init({
    chains: [1],
    optionalChains: chains,
    projectId: BEAMER_PROJECT_ID,
    optionalMethods: [
      'wallet_switchEthereumChain',
      'wallet_watchAsset',
      'wallet_addEthereumChain',
    ],
    optionalEvents: ['accountsChanged', 'chainChanged', 'network', 'disconnect'],
    rpcMap: rpcList,
    showQrModal: true,
  });
  ```
We are providing L1 as the main required chain and every other chain as an optional chain (app might require it at some point - when switching between chains).
Important change here is that we are also passing the `optionalMethods` & `optionalEvents` that our app *depends* on. I know, it is weird that we are defining them as optional since our app strongly depends on them, but if we do it via `methods` & `events` property, we are simply overriding the methods & events that are configured internally in the sdk's provider.
We don't want this as we don't want to maintain the list of the basic methods & events that an RPC provider supports. Instead we want to extend the list of methods/events and therefore this approach works.
 
From my perspective, I would assume that the above described approach should be the expected integration of a wallet app integrating with WalletConnect v2. Otherwise, we would have to provide all chains that our app supports as required chains and in most of the cases this will lead to bad UX as users will need to configure all the networks our app supports inside their wallet app just to do a single transfer from one chain to another (only 1 chain is used for a single transfer) .. 

And now it gets worse ...
This approach doesn't work on BitKeep. For TokenPocket & TrustWallet its even worse, they only support connecting on one chain per WC connection and there is no way to switch between chains programatically.
As you can see, the reason why WC2 doesn't work is different for each wallet. 

Not sure how to proceed here. 

We have a few options:
- Use the above mentioned approach and wait until all the apps adapt to behave in a similar way as MM behaves
- Use the approach that was used before WalletConnect V2 (all chains as required and user has to configure all of them inside his wallet app before he can submit a transfer)

The second approach will work for BitKeep but not for TokenPocket / TrustWallet / MetaMask ...

File: frontend/src/services/web3-provider/walletconnect-provider.ts [hunk 1]
```diff
@@ -1,44 +1,64 @@
 import { hexValue } from 'ethers/lib/utils';
 
 import { EthereumProvider } from '@/services/web3-provider/ethereum-provider';
-import type { Eip1193Provider } from '@/services/web3-provider/types';
 import { WalletConnect } from '@/services/web3-provider/util-export';
 
+const BEAMER_PROJECT_ID = import.meta.env.VITE_WALLET_CONNECT_PROJECT_ID || '';
```
compojoom, line 6: Aren't we going to get an error from the walletconnect relay without a project id?

GabrielBuragev: Yeah, but since the `VITE_WALLET_CONNECT_PROJECT_ID` exists in all the env files, it shouldn't be a problem.
Unless we remove them at one point, but that is highly unlikely unless we stop supporting WC.

PR #1976, opened 2023-06-23
Polygon zkevm fix flaky test

compojoom:
The previous deployer account 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 is actually being used by polygonZkEVM as a sequencer account. Using the same account sometimes causes a “nonce too low” error since the clients can’t agree on the correct current nonce.

Instead of this account, we use one of the accounts available on the node with funds:
https://github.com/0xPolygonHermez/zkevm-node/blob/develop/docs/running_local.md

After that we use the deposit script https://github.com/0xPolygonHermez/zkevm-bridge-service/blob/develop/test/scripts/deposit/main.go to fund the account on L2.

### DIFF
--- a/docker/polygon_zkevm/polygon_zkevm.sh
+++ b/docker/polygon_zkevm/polygon_zkevm.sh
@@ -25,8 +25,8 @@ down() {
 
 configure_repo() {
     repo="https://github.com/0xPolygonHermez/zkevm-bridge-service.git"
-    commit_id="09e0247b335a19a5e16945e8e1e805e410e41ca3"
-    git init ${POLYGON_ZKEVM}
+    commit_id="59c1ba75fa19cf1156762a8c5ffa7f8fd272c60a"
+    git clone --no-checkout ${repo} ${POLYGON_ZKEVM}
     cd ${POLYGON_ZKEVM}
     git fetch --depth 1 ${repo} ${commit_id}
     git checkout ${commit_id}
@@ -79,9 +79,16 @@ e2e_test() {
         "bridgeServiceUrl": "http://localhost:8080"
     }
 EOF
-    e2e_test_fill $ARTIFACTS_DIR $l2_rpc $KEYFILE "${password}"
-    e2e_test_relayer http://localhost:8545 $l2_rpc $network_file $PRIVKEY $e2e_test_l2_txhash
-    e2e_test_verify $ARTIFACTS_DIR $l2_rpc $ADDRESS $e2e_test_request_id
+    e2e_test_fill ${DEPLOYMENT_DIR} ${KEYFILE} "${password}" $l2_rpc
+    echo Starting relayer
+    ${relayer} relay --l1-rpc-url http://localhost:8545 \
+               --l2-relay-to-rpc-url $l2_rpc \
+               --l2-relay-from-rpc-url $l2_rpc \
+               --network-to $network_file \
+               --network-from $network_file \
+               --wallet-private-key $PRIVKEY \
+               --l2-transaction-hash $e2e_test_l2_txhash
+    e2e_test_verify ${DEPLOYMENT_DIR} $l2_rpc $ADDRESS $e2e_test_request_id
 }
 
 usage() {
--- a/docker/polygon_zkevm/polygon_zkevm.sh
+++ b/docker/polygon_zkevm/polygon_zkevm.sh
@@ -2,8 +2,10 @@
 . "$(realpath $(dirname $0))/../scripts/common.sh"
 
 # Deployer's address & private key.
-ADDRESS=0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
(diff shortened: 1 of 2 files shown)


GabrielBuragev (approved):
LGTM 👍

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -43,6 +45,21 @@ create_deployment_config_file () {
...
 
+fund_account() {
+  echo "Fund account"
+  DEPOSIT_FILE="${POLYGON_ZKEVM}/test/scripts/deposit/main.go"
+  # Replace the hardcoded address and private key with our L1 account
+  sed -i'' -e 's/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266/0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC/' "${DEPOSIT_FILE}"
+  sed -i'' -e 's/0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80/0x5de4111afa1a4b94908f83103eb1f1706367c2e68ca870fc3fb9a804cdab365a/' "${DEPOSIT_FILE}"
+
+  cd ${POLYGON_ZKEVM}
+  go run test/scripts/deposit/main.go
```
bilbeyt, line 56: how this didnt fail at the first place? i am pretty sure it will fail as soon as we changed the cache key.

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -25,8 +27,8 @@ down() {
 
 configure_repo() {
     repo="https://github.com/0xPolygonHermez/zkevm-bridge-service.git"
-    commit_id="09e0247b335a19a5e16945e8e1e805e410e41ca3"
-    git init ${POLYGON_ZKEVM}
+    commit_id="59c1ba75fa19cf1156762a8c5ffa7f8fd272c60a"
+    git clone --no-checkout ${repo} ${POLYGON_ZKEVM}
```
bilbeyt, line 31: why we changed the initialization here

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -43,6 +45,21 @@ create_deployment_config_file () {
     cd ${ROOT}
 }
 
+fund_account() {
+  echo "Fund account"
+  DEPOSIT_FILE="${POLYGON_ZKEVM}/test/scripts/deposit/main.go"
+  # Replace the hardcoded address and private key with our L1 account
+  sed -i'' -e 's/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266/0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC/' "${DEPOSIT_FILE}"
+  sed -i'' -e 's/0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80/0x5de4111afa1a4b94908f83103eb1f1706367c2e68ca870fc3fb9a804cdab365a/' "${DEPOSIT_FILE}"
```
istankovic, line 53: Why duplicate the value of the `PRIVKEY` variable, why not use the variable?

docker/polygon_zkevm/polygon_zkevm.sh
```diff
@@ -43,6 +45,21 @@ create_deployment_config_file () {
     cd ${ROOT}
 }
 
+fund_account() {
+  echo "Fund account"
+  DEPOSIT_FILE="${POLYGON_ZKEVM}/test/scripts/deposit/main.go"
+  # Replace the hardcoded address and private key with our L1 account
+  sed -i'' -e 's/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266/0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC/' "${DEPOSIT_FILE}"
```
istankovic, line 52: Why duplicate the value of the `ADDRESS` variable, why not use the variable?

docker/polygon_zkevm/polygon_zkevm.sh
compojoom: @GabrielBuragev do you know? I haven't actually touched that.

docker/polygon_zkevm/polygon_zkevm.sh
compojoom: the vm we use has go in it. https://discuss.circleci.com/t/linux-machine-executor-update-2023-q1-update/47059

docker/polygon_zkevm/polygon_zkevm.sh
GabrielBuragev: Probably a mistake while merging the two versions of the script on my end. Can you try with the old approach?

compojoom: ok, I've changed it back in the new pr https://github.com/beamer-bridge/beamer/pull/1977

PR #1977, opened 2023-06-23
relayer: use variables for replace in sed

compojoom:
I was too fast to merge this PR https://github.com/beamer-bridge/beamer/pull/1976

Addressed @istankovic comments here.

### DIFF
--- a/docker/polygon_zkevm/polygon_zkevm.sh
+++ b/docker/polygon_zkevm/polygon_zkevm.sh
@@ -28,7 +28,7 @@ down() {
 configure_repo() {
     repo="https://github.com/0xPolygonHermez/zkevm-bridge-service.git"
     commit_id="59c1ba75fa19cf1156762a8c5ffa7f8fd272c60a"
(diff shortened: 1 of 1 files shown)


istankovic (approved):
Looks good.

istankovic:
@compojoom could you please rebase on top of latest `main`, I pushed a fix for the lint failure.

istankovic:
@compojoom I just noticed your commit messages are prefixed with `relayer:`, which seems wrong as we are not modifiying the relayer here. Could you please change them?

compojoom:
ok, so what? docker? I'm adding relayer as those e2e tests are for the relayer?

istankovic:
> ok, so what? docker? 

Yeah, I think most people use just `docker:` for that part.

> I'm adding relayer as those e2e tests are for the relayer?

Well, we do test the relayer, but also almost everything else (this is where the `e2e == end-to-end` part comes from). So it's not just about the relayer. (Generally, relayer changes are prefixed with `relayer:` and touch stuff under `relayer/`.)

compojoom:
k, done

Issue #1685, opened 2023-04-11
Add wrapped eth


andfletcher:
## Description
As a User, I want to be able to bridge wrapped eth to the beamer-connected rollups / chains so that I can use it there and don't need to use other bridges

## Todo

- [x] Support WETH in the Frontend https://github.com/beamer-bridge/beamer/pull/1980
- [x] Add WETH to contracts on all supported chains
- [x]  Make sure that contract configuration values are double checked by the teammates 
- [x] Update run-your-own-agent repo config file(s) with WETH
- [x] Inform agents about the need to provide WETH liquidity and the necessary agent config changes

## Acceptance criteria

Users are able to WETH between Beamer-connected chains, via the regular Beamer app.

PR #1980, opened 2023-06-23
frontend: add USDT + WETH tokens

GabrielBuragev:
Implements parts of #1969 & #1685

I implemented a way to hide the tokens as well (just like for the chains) when we are not live yet.
USDT + WETH are marked as hidden since we will probably not see the other parts of the above mentioned issues finished before this PR gets merged. 

Once all the TODOs in the issues are finished, open another PR (or edit this one if still there) to disable the "hiding" feature for these 2 tokens.


compojoom (approved):
Looks good, but please merge the whole add/hide commits and the fixup ones :)

PR #1986, opened 2023-06-27
Commands improvements

istankovic:
Small improvements to config commands:
- say whether any config updates have been found and, if yes, how many
- create state path parent directories as needed in case they don't exist
- additional check in the `read` command to make sure deployment and config chain IDs are same
- more tests for the `read` command

### DIFF
--- a/beamer/config/commands.py
+++ b/beamer/config/commands.py
@@ -30,6 +30,10 @@ log = structlog.get_logger(__name__)
 _CONFIG_UPDATE_EVENTS = frozenset((ChainUpdated, FeesUpdated, TokenUpdated, LpAdded, LpRemoved))
 
 
+def _is_config_update_event(event: Event) -> bool:
+    return type(event) in _CONFIG_UPDATE_EVENTS
+
+
 def _replay_event(w3: Web3, deployment: Deployment, config: Configuration, event: Event) -> None:
     assert deployment.chain is not None
     match event:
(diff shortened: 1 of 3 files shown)


beamer/config/commands.py
```diff
@@ -30,6 +30,10 @@
 _CONFIG_UPDATE_EVENTS = frozenset((ChainUpdated, FeesUpdated, TokenUpdated, LpAdded, LpRemoved))
 
 
+def _is_config_update_event(event: Event) -> bool:
+    return type(event) in _CONFIG_UPDATE_EVENTS
```
fredo, line 34: nitpick: this could be misleading, I could do updateChain with the exact same values resulting in `_is_config_update_event ` to be true. Or the values could have been changed and then changed back. 
I think it would be more appropriate to check the original state with the state after applying new events and evaluate based on this. Depending on how much effort it is...

beamer/config/commands.py
```diff
@@ -133,9 +137,15 @@ def read(
         w3, (request_manager, fill_manager), start_block=start_block, confirmation_blocks=0
     )
     events = fetcher.fetch()
+    events = list(filter(_is_config_update_event, events))
     for event in events:
         _replay_event(w3, deployment, config, event)
 
+    if events:
```
fredo, line 151: nitpick: this could be misleading, I could do updateChain with the exact same values resulting in _is_config_update_event  to be true. Or the values could have been changed and then changed back.
I think it would be more appropriate to check the original state with the state after applying new events and evaluate based on this. Depending on how much effort it is...

fredo: looking at the log info provided, I think this is actually good enough, so nvm

fredo (approved):
Looks all good to me, thanks for applying the feedback

PR #1994, opened 2023-06-29
fix: change fill/invalidation timestamp after prove for op

bilbeyt:
We need to wait for finality period that starts after prove operation for OP.

### DIFF
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -591,17 +591,20 @@ def _timestamp_is_l1_finalized(
 
 
 def maybe_prove(claim: Claim, context: Context) -> None:
-    # mainnet: 10, goerli: 420, local: 901
-    if context.target_chain.id not in (10, 420, 901):
-        claim.message_proved = True
-        return
-
-    if claim.message_proved:
-        return
-
     request = context.requests.get(claim.request_id)
 
     assert request is not None, "Active claim for non-existent request"
+
+    prove_tx = None
+    if request.fill_tx is not None:
+        prove_tx = request.fill_tx
+    elif claim.invalidation_tx is not None:
+        prove_tx = claim.invalidation_tx
+
+    # mainnet: 10, goerli: 420, local: 901
+    if context.target_chain.id not in (10, 420, 901):
+        claim.proved_tx = prove_tx
+        return
 
     own_challenge_stake = Wei(claim.get_challenger_stake(context.address))
     can_prove = (
@@ -611,12 +614,10 @@ def maybe_prove(claim: Claim, context: Context) -> None:
     if not can_prove:
         return
 
-    if request.fill_tx is not None:
-        proof_tx = request.fill_tx
-    elif claim.invalidation_tx is not None:
-        proof_tx = claim.invalidation_tx
+    if claim.proved_tx is not None or prove_tx in context.l1_resolutions:
+        return
 
-    if proof_tx in context.l1_resolutions:
+    if prove_tx is None:
         return
 
     future = context.task_pool.submit(
@@ -625,7 +626,7 @@ def maybe_prove(claim: Claim, context: Context) -> None:
         context.target_chain.rpc_url,
         context.source_chain.rpc_url,
         context.config.account.key,
-        proof_tx,
+        prove_tx,
         True,
     )
 
@@ -633,16 +634,22 @@ def maybe_prove(claim: Claim, context: Context) -> None:
         try:
             f.result()
         except Exception as ex:
-            context.logger.error("Optimism prove failed", ex=ex, tx_hash=proof_tx)
+            context.logger.error("Optimism prove failed", ex=ex, tx_hash=prove_tx)
         else:
-            claim.message_proved = True
+            assert request is not None
+            claim.proved_tx = prove_tx
+            if request.fill_tx == prove_tx:
+                request.fill_timestamp = Timestamp(int(time.time()))
+            elif claim.invalidation_tx == prove_tx:
+                claim.invalidation_timestamp = Timestamp(int(time.time()))
         finally:
-            context.l1_resolutions.pop(proof_tx, None)
+            assert prove_tx is not None
+            context.l1_resolutions.pop(prove_tx, None)
 
     future.add_done_callback(on_future_done)
-    context.l1_resolutions[proof_tx] = future
+    context.l1_resolutions[prove_tx] = future
 
-    context.logger.info("Initiated Optimism prove", request=request, claim=claim, tx_hash=proof_tx)
+    context.logger.info("Initiated Optimism prove", request=request, claim=claim, tx_hash=prove_tx)
 
 
 def maybe_resolve(claim: Claim, context: Context) -> bool:
@@ -650,7 +657,7 @@ def maybe_resolve(claim: Claim, context: Context) -> bool:
 
     assert request is not None, "Active claim for non-existent request"
 
-    if not claim.message_proved:
+    if claim.proved_tx is None or claim.proved_tx in context.l1_resolutions:
         return False
 
     if not _proof_ready_for_l1_relay(request) and not _invalidation_ready_for_l1_relay(claim):
@@ -658,26 +665,26 @@ def maybe_resolve(claim: Claim, context: Context) -> bool:
 
     timestamp_finalized = False
 
-    if request.fill_timestamp:
-        timestamp_finalized = _timestamp_is_l1_finalized(
-            request.fill_timestamp, context, request.target_chain_id
-        )
-        assert request.fill_tx is not None
-        proof_tx = request.fill_tx
-    elif claim.invalidation_timestamp:
-        timestamp_finalized = _timestamp_is_l1_finalized(
-            claim.invalidation_timestamp, context, request.target_chain_id
-        )
-        assert claim.invalidation_tx is not None
-        proof_tx = claim.invalidation_tx
+    match claim.proved_tx:
+        case request.fill_tx:
+            assert request.fill_timestamp is not None
+            timestamp_finalized = _timestamp_is_l1_finalized(
+                request.fill_timestamp, context, request.target_chain_id
+            )
+        case claim.invalidation_tx:
+            assert claim.invalidation_timestamp is not None
+            timestamp_finalized = _timestamp_is_l1_finalized(
+                claim.invalidation_timestamp, context, request.target_chain_id
+            )
+        case _:
+            raise ValueError(
+                f"proved transaction is neither a fill nor an invalidation: {claim.proved_tx!r}"
+            )
 
     if not timestamp_finalized:
         return False
 
     if not _l1_resolution_threshold_reached(claim, context):
-        return False
-
-    if proof_tx in context.l1_resolutions:
         return False
 
     future = context.task_pool.submit(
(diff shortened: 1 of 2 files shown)


beamer/agent/chain.py
```diff
@@ -636,6 +636,11 @@ def on_future_done(f: Future) -> None:
             context.logger.error("Optimism prove failed", ex=ex, tx_hash=proof_tx)
         else:
             claim.message_proved = True
+            assert request is not None
+            if request.fill_timestamp is not None:
+                request.fill_timestamp = Timestamp(int(time.time()))
+            elif claim.invalidation_timestamp is not None:
+                claim.invalidation_timestamp = Timestamp(int(time.time()))
```
istankovic, line 643: I would prefer if we went further than this.
For instance, we have
```
def _proof_ready_for_l1_relay(request: Request) -> bool:
    return (                                            
        not request.l1_resolved.is_active               
        and request.fill_tx is not None                 
        and request.fill_timestamp is not None          
    )                                                   
```
Checking for "fill timestamp" does not make sense in the context of proof readiness anymore.
I would consider replacing `request.fill_timestamp` with something like `request.proof_timestamp`, which would be `None` until the proof is actually done.

istankovic: Additionally, there is

```python
if request.fill_timestamp:                                      
    timestamp_finalized = _timestamp_is_l1_finalized(           
        request.fill_timestamp, context, request.target_chain_id
    )                                                           
    assert request.fill_tx is not None                          
    proof_tx = request.fill_tx                                  
```
in `maybe_resolve`. Note how we set `proof_tx` to be `request.fill_tx`. The question here is, are we still okay with naming like this? I _think_ the code should still continue to work, but it may be misleading which "proof" we are talking about.

bilbeyt: will change it to resolution_tx

istankovic: > will change it to resolution_tx

I'm not sure that's much better because that could also mean the tx on L1.
What we're talking about is really the original tx on L2 (on the transfer's target chain).

bilbeyt: resolvable_tx :D

istankovic:
Looking further, we have
```python
   if request.fill_timestamp:                                             
       timestamp_finalized = _timestamp_is_l1_finalized(                  
           request.fill_timestamp, context, request.target_chain_id       
       )                                                                  
       assert request.fill_tx is not None                                 
       proof_tx = request.fill_tx                                         
   elif claim.invalidation_timestamp:                                     
       timestamp_finalized = _timestamp_is_l1_finalized(                  
           claim.invalidation_timestamp, context, request.target_chain_id 
       )                                                                  
       assert claim.invalidation_tx is not None                           
       proof_tx = claim.invalidation_tx                                   
```
in `maybe_resolve`.

I wonder why we have timestamps on separate objects (request and claim).
Could we not have a single `proof_timestamp` on `claim`?

istankovic:
- turn `claim.message_proved` into `claim.proved_tx` to keep track of the tx hash of the proved message (either a fill or an invalidation)
- modify `maybe_resolve` to use `claim.proved_tx` to resolve the correct message; exit prematurely if `proved_tx` is `None`
- check tx hashes instead of timestamps in `maybe_prove`
- set correct `proved_tx` for all non-OP chains in `maybe_prove` (priority: fill, invalidation)

beamer/agent/chain.py
```diff
@@ -611,73 +615,76 @@ def maybe_prove(claim: Claim, context: Context) -> None:
...
-    if request.fill_tx is not None:
-        proof_tx = request.fill_tx
-    elif claim.invalidation_tx is not None:
-        proof_tx = claim.invalidation_tx
-
-    if proof_tx in context.l1_resolutions:
+    if proved_tx in context.l1_resolutions:
         return
 
+    assert proved_tx is not None
```
istankovic, line 621: We can remove this `assert` and move `if proved_tx in ...` check to just below line 609 so that we don't even do other checks if `proved_tx is None` or we already have a resolution for it:

```python
    if proved_tx is None or proved_tx in context.l1_resolutions:
        return
```

beamer/agent/chain.py
```diff
@@ -611,73 +615,76 @@ def maybe_prove(claim: Claim, context: Context) -> None:
...
 
     def on_future_done(f: Future) -> None:
         try:
             f.result()
         except Exception as ex:
-            context.logger.error("Optimism prove failed", ex=ex, tx_hash=proof_tx)
+            context.logger.error("Optimism prove failed", ex=ex, tx_hash=proved_tx)
         else:
-            claim.message_proved = True
+            assert request is not None
```
istankovic, line 639: Why is this assert needed? We already have that check outside of this closure.

bilbeyt: because mypy thinks otherwise maybe it is because this scope is different and mypy has no idea hey i already checked this

beamer/agent/chain.py
bilbeyt: sure

beamer/agent/chain.py
```diff
@@ -611,73 +615,76 @@ def maybe_prove(claim: Claim, context: Context) -> None:
...
+            claim.proved_tx = proved_tx
+            if request.fill_tx == proved_tx:
+                request.fill_timestamp = Timestamp(int(time.time()))
+            elif claim.invalidation_tx == proved_tx:
+                claim.invalidation_timestamp = Timestamp(int(time.time()))
         finally:
-            context.l1_resolutions.pop(proof_tx, None)
+            assert proved_tx is not None
```
istankovic, line 646: Do we need this assert? `proved_tx` is guaranteed not to be `None` here.

bilbeyt: same answer as above

beamer/agent/chain.py
bilbeyt: Item "None" of "Optional[Request]" has no attribute "fill_tx"

beamer/agent/chain.py
bilbeyt: Argument 1 to "pop" of "dict" has incompatible type "Optional[HexBytes]"; expected "HexBytes"

beamer/agent/chain.py
istankovic: Ugh, mypy is not smart enough to deduce that. Oh well.

beamer/agent/chain.py
```diff
@@ -625,59 +624,65 @@ def maybe_prove(claim: Claim, context: Context) -> None:
...
-        assert request.fill_tx is not None
-        proof_tx = request.fill_tx
-    elif claim.invalidation_timestamp:
+    else:
+        assert claim.invalidation_timestamp is not None
         timestamp_finalized = _timestamp_is_l1_finalized(
             claim.invalidation_timestamp, context, request.target_chain_id
         )
-        assert claim.invalidation_tx is not None
-        proof_tx = claim.invalidation_tx
```
istankovic, line 672: Perhaps consider turning this `if... else..` block into a `match` and adding a check for the unexpected case?
E.g.

```python
match claim.proved_tx:
   case request.fill_tx:
       ...
   case claim.invalidation_tx:
      ...
   case _:
      raise ValueError(f'proved transaction is neither a fill nor an invalidation: {claim.proved_tx}')
```

beamer/agent/chain.py
```diff
@@ -625,59 +624,65 @@ def maybe_prove(claim: Claim, context: Context) -> None:
...
 
     if not timestamp_finalized:
         return False
 
     if not _l1_resolution_threshold_reached(claim, context):
         return False
 
-    if proof_tx in context.l1_resolutions:
+    if claim.proved_tx in context.l1_resolutions:
         return False
```
istankovic, line 686: I would move this check earlier, e.g. right after we check that `claim.proved_tx` is not `None`. We could even combine it with that check.

beamer/agent/chain.py
```diff
@@ -686,22 +691,27 @@ def maybe_resolve(claim: Claim, context: Context) -> bool:
         context.target_chain.rpc_url,
         context.source_chain.rpc_url,
         context.config.account.key,
-        proof_tx,
+        claim.proved_tx,
         False,
     )
 
+    assert claim.proved_tx is not None
```
istankovic, line 698: I think we don't need this assert here, we check for it already above.

istankovic (approved):
Looks good from my side, but would love if @fredo also gave it a look.

beamer/agent/chain.py
bilbeyt: i dont think it needs to raise so just used return False

beamer/agent/chain.py
```diff
@@ -625,83 +624,92 @@ def maybe_prove(claim: Claim, context: Context) -> None:
...
+            timestamp_finalized = _timestamp_is_l1_finalized(
+                request.fill_timestamp, context, request.target_chain_id
+            )
+        case claim.invalidation_tx:
+            assert claim.invalidation_timestamp is not None
+            timestamp_finalized = _timestamp_is_l1_finalized(
+                claim.invalidation_timestamp, context, request.target_chain_id
+            )
+        case _:
+            return False
```
istankovic, line 680: Actually, I would raise an exception here, like I commented before, and not return `False`. Because if this happens, something seriously went wrong and we want to know about it.

beamer/agent/chain.py
```diff
@@ -591,18 +591,22 @@ def _timestamp_is_l1_finalized(
...
-    if claim.message_proved:
-        return
-
     request = context.requests.get(claim.request_id)
 
     assert request is not None, "Active claim for non-existent request"
 
+    proved_tx = None
+    if request.fill_tx is not None:
+        proved_tx = request.fill_tx
```
fredo, line 600: I think `proved_tx` is not a proper naming, as at this point it may not be proved yet. 

maybe `prove_tx` ? This is the tx which the relayer will prove

beamer/agent/chain.py
```diff
@@ -591,18 +591,22 @@ def _timestamp_is_l1_finalized(
...
+    elif claim.invalidation_tx is not None:
+        proved_tx = claim.invalidation_tx
+
+    # mainnet: 10, goerli: 420, local: 901
+    if context.target_chain.id not in (10, 420, 901):
+        if claim.proved_tx is None:
```
fredo, line 606: This is not correct. In the case of Non-OP chains the claim.proved_tx must be set unconditionally. (fill_tx should overwrite invalidation_tx)

beamer/agent/chain.py
```diff
@@ -611,12 +615,7 @@ def maybe_prove(claim: Claim, context: Context) -> None:
     if not can_prove:
         return
 
-    if request.fill_tx is not None:
-        proof_tx = request.fill_tx
-    elif claim.invalidation_tx is not None:
-        proof_tx = claim.invalidation_tx
-
-    if proof_tx in context.l1_resolutions:
+    if proved_tx is None or proved_tx in context.l1_resolutions:
```
fredo, line 618: There is an additional check missing here, that if claim.proved_tx is not None then it should exit. 
Note, in this part we are sure that we are an OP chain. If one of the claim.proved_tx is already proved, this is the one which should be resolved in the future.

fredo (approved):
looks good now 👍

Issue #1975, opened 2023-06-22
Make config commands and deployment commands use specified ABI dir 


andfletcher:
## Description
@istankovic please add some info here

PR #2003, opened 2023-06-30
ABI dir fixes

istankovic:
This PR fixes the current situation with ABI handling (i.e. not using the specfied ABI dir at all and instead relying on `contracts/.build`). A major change was to move `beamer.deploy.artifacts` one level up, which required extensive preparation itself. This was necessary due to a headache of circular imports between `beamer.deploy.artifacts`, `beamer.deploy.util` and `beamer.contracts`, the latter being in need of deployment information if it is to do anything useful. Also, it just makes sense on its own because other components need access to artifacts too.

Once this was done, things became somewhat easier, especially proceeding with `beamer.contract` changes now. Scripts and commands have been updated to make use of the ABI dir, but also the entire code-base was switched to use the new artifacts APIs. In turn, this allowed us to simplify and remove some of the code in `beamer.contracts` that has now become obsolete.

Summary of the changes
- use the newly-introduced `ABIManager` API across the codebase
- use the new artifacts API wherever appropriate
- remove old facilities for loading ABI files and deployment artifacts
- make `beamer config`, `beamer deploy` and `beamer deploy-base` commands handle ABI dir correctly
- fix fallout in scripts and tests

Closes: #1975


beamer/deploy/commands.py
```diff
@@ -182,7 +197,7 @@ def deploy(
     assert base_w3.eth.chain_id == base_deployment.base.chain_id
     log.info("Connected to base chain RPC", chain_id=base_deployment.base.chain_id, url=url)
 
-    resolver = base_deployment.obtain_contract(base_w3, "base", "Resolver")
+    resolver = obtain_contract(base_w3, abi_dir, base_deployment, "Resolver")
```
fredo: question: Is the base deployment also stored as a `Deployment` object with `chain = None` ?

istankovic: > question: Is the base deployment also stored as a `Deployment` object with `chain = None` ?

Yes, that's exactly what differentiates the base deployment from non-base deployments.

beamer/contracts.py
```diff
@@ -51,6 +52,36 @@ def prepare_deployment_infos(
...
+        self._cache: dict[str, ABIManager._CacheEntry] = {}
+
+    def get_abi(self, name: str) -> str:
+        entry = self._cache.get(name)
+        if entry is None:
+            entry = self._load_entry(name)
+            self._cache[name] = entry
+        return entry.abi
+
+    def get_bytecode(self, name: str) -> str:
```
bilbeyt, line 27: this seems like a duplicate to me, only difference is just the attribute namely abi or bytecode

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -229,19 +222,22 @@ def _mint_agent_tokens(
 
 
 def test_multiple_event_processors(
-    contracts: Contracts, token: ape.project.MintableToken, local_account: TestAccount
+    tmp_path, token: ape.project.MintableToken, local_account: TestAccount
```
bilbeyt, line 225: uups typing not here

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
```diff
@@ -274,20 +270,21 @@ def _start_agent_fee_test(config: Config):
         agent.stop()
 
 
-def test_l1_base_fees(
-    contracts: Contracts, token: ape.project.MintableToken, local_account: TestAccount
-):
+def test_l1_base_fees(tmp_path, token: ape.project.MintableToken, local_account: TestAccount):
```
bilbeyt, line 273: also here, maybe you can check this file for typings.

beamer/contracts.py
```diff
@@ -94,10 +94,19 @@ def load_deployment_info(artifacts_dir: Path, abi_dir: Path) -> DeploymentInfo:
     return deployment_info
 
 
-def contracts_for_web3(web3: Web3, artifacts_dir: Path, abi_dir: Path) -> dict[str, Contract]:
-    deployment_info = load_deployment_info(artifacts_dir, abi_dir)
-    chain_id = ChainId(web3.eth.chain_id)
-    return make_contracts(web3, deployment_info[chain_id])
+def contracts_for_web3(w3: Web3, artifacts_dir: Path, abi_dir: Path) -> dict[str, Contract]:
```
bilbeyt, line 43: Well this is not like super accurate because w3 can be basically l1 and then you need to traverse and get l1messengers as well to be more accurate

pyproject.toml
```diff
@@ -12,7 +12,7 @@ documentation = "https://docs.beamerbridge.com/"
 [tool.poetry.dependencies]
 python = ">=3.10,<3.11"
 web3 = "^6.5.0"
-click = "^8.1.4"
+click = "=8.1.3"
```
bilbeyt, line 15: do we really need = and why we are doing this?

bilbeyt (changes requested):
Generally looking really good with some small things.

pyproject.toml
istankovic: > do we really need = and why we are doing this?

I would say yes, for the time being because the click update to 8.1.4 broke a bunch of things:
https://app.circleci.com/pipelines/github/beamer-bridge/beamer/3466/workflows/db56e0fb-aa9c-48b3-8179-0e3f3d78f7a6/jobs/33514

I would prefer to address the conversion to new click later.

beamer/tests/agent/multiple_processors/test_multiple_event_processors.py
istankovic: Will add type.

beamer/tests/agent/utils.py
```diff
@@ -46,14 +47,15 @@ def generate_artifacts(artifacts_dir, contracts):
             },
         },
     }
-    artifacts_dir.mkdir()
-    with artifacts_dir.joinpath("1337-ethereum.deployment.json").open("wt") as f:
+    artifacts_dir.mkdir(exist_ok=True)
+    path = artifacts_dir / f"{chain_id}-ethereum.deployment.json"
```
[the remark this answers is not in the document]
istankovic: > not sure about this part, what if chain id changes because of an upgrade or something then we dont have config json at all.

This is not about config (artifacts), here we are generating the deployment artifacts.
Changes in chain ID should not affect anything here since we don't look up those, at least in our automated tests.
(The E2E test and container image test explicitly mention the chain ID, but those are fine and are not related to this.)

However, your comment made me look at `beamer/tests/config/util.py`, and there we could/should do the same thing as we are doing here. No reason to hardcode chain IDs.

beamer/contracts.py
fredo: I'd like to chime in here. looking at `contracts_for_web3` this function is only used in `call_contracts`,`e2e-test-fill` and `e2e-test-verify`.

Speaking for the e2e tests at least, we only need RequestManager FillManager and MintableToken. 
How far would it be away to also use the new API in the tests and get rid of contracts_for_web3 completely? It doesn't seem to have too much use in our code base.

fredo: Looked at call_contracts, here we are also only ever using the above mentioned contracts

istankovic: > I'd like to chime in here. looking at `contracts_for_web3` this function is only used in `call_contracts`,`e2e-test-fill` and `e2e-test-verify`.
> 
> Speaking for the e2e tests at least, we only need RequestManager FillManager and MintableToken. How far would it be away to also use the new API in the tests and get rid of contracts_for_web3 completely? It doesn't seem to have too much use in our code base.

This is precisely what I was debating myself before finishing the branch. `contracts_for_web3` is primarily used as a convenience function for scripts like the ones you named. In this light, it can come quite handy if you don't want to deal with loading chain-specific artifacts on your own. Also, given that it constructs an empty `ABIManager` every time it is called, it is clearly not ideal in case of multiple calls; I've even wanted to add a comment regarding that, along the lines of `to be used only in scripts`.

On the other hand, I think @bilbeyt has a valid point here and I would like to find a better solution.
Also, `beamer.artifacts.load_all` is not ideally named and I would like to improve on that.
Perhaps with an additional API to help out "the script use cases" and better naming we can get rid of this function entirely.

istankovic: Something like this, perhaps?
```diff
diff --git a/scripts/e2e-test-fill.py b/scripts/e2e-test-fill.py
index b0ceede8..394e9278 100644
--- a/scripts/e2e-test-fill.py
+++ b/scripts/e2e-test-fill.py
@@ -20,15 +20,17 @@ def main() -> None:
     deployer = account_from_keyfile(keystore_file, password)
     web3 = make_web3(l2_rpc, deployer)
 
-    l2_contracts = contracts_for_web3(web3, artifacts_dir, abi_dir)
+    abi_manager = ABIManager(abi_dir)
+    deployment = beamer.artifacts.load(web3.eth.chain_id)
+
     overriden_chain_id = os.getenv("SOURCE_CHAIN_ID")
     if overriden_chain_id is None:
         source_chain_id = ChainId(web3.eth.chain_id)
     else:
         source_chain_id = ChainId(int(overriden_chain_id))
 
-    fill_manager = l2_contracts["FillManager"]
-    token = l2_contracts["MintableToken"]
+    fill_manager = obtain_contract(web3, abi_manager, deployment, "FillManager")
+    token = obtain_contract(web3, abi_manager, deployment, "MintableToken")
 
     nonce = random.randint(1, sys.maxsize)
     request_amount = 123
```

istankovic: I have pushed a changes like above and removed both `beamer.artifacts.load_all` and `beamer.contracts.contracts_for_web3`, I think now it makes much more sense. Please take a look again.

beamer/artifacts.py
```diff
@@ -57,3 +57,12 @@ def from_file(artifact: Path) -> "Deployment":
     def to_file(self, artifact: Path) -> None:
         with open(artifact, "wt") as f:
             json.dump(apischema.serialize(self), f, indent=4)
+
+
+def load_all(artifacts_dir: Path) -> dict[ChainId, Deployment]:
```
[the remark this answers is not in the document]
istankovic: Fixed as described in https://github.com/beamer-bridge/beamer/pull/2003#discussion_r1259474457.

beamer/contracts.py
istankovic: Very little code is duplicated here, we could have a separate helper function, but I think it would end up pretty much the same code size, if not slightly bigger.

beamer/tests/agent/utils.py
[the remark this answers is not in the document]
istankovic: I've now removed hardcoded chain IDs there.

scripts/call_contracts.py
```diff
@@ -51,7 +52,13 @@ def cli(
 
     account = account_from_keyfile(keystore_file, password)
     web3 = make_web3(eth_rpc, account)
-    contracts = contracts_for_web3(web3, artifacts_dir, abi_dir)
+
+    abi_manager = ABIManager(abi_dir)
+    deployment = beamer.artifacts.load(artifacts_dir, ChainId(web3.eth.chain_id))
+
+    contracts = {}
+    for name in ("RequestManager", "FillManager", "MintableToken"):
```
bilbeyt, line 60: Just an idea here, how about using get_erc20_abi function to get token abi and getting token_address as input then we can use this freely with mainnet deployments as well

bilbeyt: but can be done in seperate pr

bilbeyt (approved):
LGTM!

beamer/contracts.py
fredo: looks much better 👍

PR #2009, opened 2023-07-04
Config write check mismatch config deployment chain ids

istankovic:


fredo:
looks good. What about desired config (in the case of write) and the actual RPC? 
Are these ids checked for consistency as well?

bilbeyt (approved):
LGTM! if you answer @fredo comment then it is ok for me.

istankovic:
> looks good. What about desired config (in the case of write) and the actual RPC? Are these ids checked for consistency as well?

Yes, we check the following:
- that the deployment chain ID matches the chain ID from RPC
- that the deployment chain ID matches the current config chain ID `[*]`
- that the current config chain ID matches the desired config chain ID

This PR adds the `[*]` check so now, from above, follows that

`deployment chain ID == RPC chain ID == current config chain ID == desired config chain ID`.

Issue #1900, opened 2023-06-05
Enable type 2 transactions for Optimism


fredo:
After Bedrock Optimism will allow type 2 transacitons. For dynamic fee chains we already have some middlewares in place and we could enable it for optimism.

Ideally it is as simple as adding optimisms chain id to the list of these chains in `make_web3()`. In the same go we could think about moving out this list to a more generic place instead of having the chain ids hardcoded in the middle of make_web3

PR #2023, opened 2023-07-06
feat: support type2 for op

bilbeyt:
Closes: #1900

### DIFF
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -17,10 +17,10 @@ from beamer.agent.models.claim import Claim
 from beamer.agent.models.request import Request
 from beamer.agent.relayer import run_relayer_for_tx
 from beamer.agent.state_machine import Context, process_event
+from beamer.chains import get_chain_descriptor
 from beamer.events import Event, EventFetcher
 from beamer.typing import BlockNumber, ChainId
 from beamer.util import TransactionFailed, get_ERC20_abi, transact
-
 
 # The time we're waiting for our thread in stop(), in seconds.
 # This is also the maximum time a call to stop() would block.
@@ -601,8 +601,10 @@ def maybe_prove(claim: Claim, context: Context) -> None:
     elif claim.invalidation_tx is not None:
         prove_tx = claim.invalidation_tx
 
-    # mainnet: 10, goerli: 420, local: 901
-    if context.target_chain.id not in (10, 420, 901):
+    chain_descriptor = get_chain_descriptor(context.target_chain.id)
+
+    assert chain_descriptor is not None, "Chain is not supported"
+    if not chain_descriptor.bedrock:
         claim.proved_tx = prove_tx
         return
 
--- a/beamer/chains.py
+++ b/beamer/chains.py
@@ -1,0 +1,100 @@
+from dataclasses import dataclass
+from typing import Any
+
+from beamer.typing import ChainId
+
+
+@dataclass(frozen=True)
+class ChainDescriptor:
+    id: ChainId
+    ecosystem_name: str
+    name: str
+    type2: bool = False
+    bedrock: bool = False
+    testnet: bool = False
+    local: bool = False
+
+
+_DESCRIPTORS = {
+    # Ethereum
+    ChainId(1): ChainDescriptor(
+        id=ChainId(1), ecosystem_name="ethereum", name="mainnet", type2=True
+    ),
+    ChainId(5): ChainDescriptor(
+        id=ChainId(5), ecosystem_name="ethereum", name="goerli", type2=True, testnet=True
+    ),
+    ChainId(1337): ChainDescriptor(
+        id=ChainId(1337), ecosystem_name="ethereum", name="local", type2=True, local=True
+    ),
+    ChainId(900): ChainDescriptor(
+        id=ChainId(900), ecosystem_name="ethereum", name="op-l1", local=True
+    ),
+    # Optimism
+    ChainId(10): ChainDescriptor(
+        id=ChainId(10), ecosystem_name="optimism", name="mainnet", type2=True, bedrock=True
+    ),
+    ChainId(420): ChainDescriptor(
+        id=ChainId(420),
+        ecosystem_name="optimism",
+        name="goerli",
+        type2=True,
+        bedrock=True,
+        testnet=True,
+    ),
+    ChainId(901): ChainDescriptor(
+        id=ChainId(901),
+        ecosystem_name="optimism",
+        name="local",
+        type2=True,
+        bedrock=True,
+        local=True,
+    ),
+    # Arbitrum
+    ChainId(42161): ChainDescriptor(id=ChainId(42161), ecosystem_name="arbitrum", name="mainnet"),
+    ChainId(421613): ChainDescriptor(
+        id=ChainId(421613), ecosystem_name="arbitrum", name="goerli", testnet=True
+    ),
+    ChainId(412346): ChainDescriptor(
+        id=ChainId(412346), ecosystem_name="arbitrum", name="local", local=True
+    ),
+    # Polygon ZkEVM
+    ChainId(1101): ChainDescriptor(
+        id=ChainId(1101), ecosystem_name="polygon_zkevm", name="mainnet"
+    ),
+    ChainId(1442): ChainDescriptor(
+        id=ChainId(1442), ecosystem_name="polygon_zkevm", name="goerli", testnet=True
+    ),
+    ChainId(1001): ChainDescriptor(
+        id=ChainId(1001), ecosystem_name="polygon_zkevm", name="local", local=True
+    ),
+    # Base
+    ChainId(84531): ChainDescriptor(
+        id=ChainId(84531),
+        ecosystem_name="base",
+        name="goerli",
+        testnet=True,
+        bedrock=True,
+        type2=True,
+    ),
+}
+
+
+def get_chain_descriptor(chain_id: ChainId) -> ChainDescriptor | None:
+    return _DESCRIPTORS.get(chain_id)
+
+
+def search(**kwargs: Any) -> tuple[ChainDescriptor, ...]:
+    found = []
+    for desc in _DESCRIPTORS.values():
+        if all(getattr(desc, field, None) == value for field, value in kwargs.items()):
+            found.append(desc)
+    return tuple(found)
+
+
+def register(chain_id: ChainId, descriptor: ChainDescriptor) -> None:
+    assert chain_id not in _DESCRIPTORS, "Chain is already added"
+    _DESCRIPTORS[chain_id] = descriptor
+
+
+def unregister(chain_id: ChainId) -> None:
+    del _DESCRIPTORS[chain_id]
--- a/beamer/tests/agent/unit/test_challenge.py
+++ b/beamer/tests/agent/unit/test_challenge.py
@@ -9,6 +9,7 @@ from web3.types import BlockData, Wei
 
 from beamer.agent.agent import Chain
 from beamer.agent.chain import process_claims
+from beamer.chains import search
 from beamer.tests.agent.unit.utils import (
     ADDRESS1,
     CLAIMER_STAKE,
@@ -21,7 +22,7 @@ from beamer.tests.agent.unit.utils import (
 )
 from beamer.tests.agent.utils import make_address
 from beamer.tests.constants import FILL_ID
-from beamer.typing import ChainId, FillId, Termination
+from beamer.typing import FillId, Termination
 
 
 @pytest.mark.parametrize("fill_id", [FILL_ID, FillId(b"cafebabe")])
@@ -99,7 +100,7 @@ def test_join_false_claim_challenge_only_when_unfilled(filler):
 def test_optimism_prove(mocked_relayer_call):
     timestamp = int(time.time())
     mocked_relayer_call.return_value = str(timestamp)
-    op_chain_id = ChainId(901)
+    op_chain_id = next(iter(search(bedrock=True, local=True))).id
     context, _ = make_context()
     context.latest_blocks[op_chain_id] = BlockData(
         {"number": BlockNumber(30), "timestamp": TIMESTAMP}
--- a/beamer/tests/agent/unit/utils.py
+++ b/beamer/tests/agent/unit/utils.py
@@ -16,6 +16,7 @@ from beamer.agent.models.request import Request
 from beamer.agent.state_machine import Context
 from beamer.agent.tracker import Tracker
 from beamer.agent.util import TokenChecker
+from beamer.chains import ChainDescriptor, register, search
 from beamer.events import ClaimMade
 from beamer.tests.agent.utils import make_address
 from beamer.tests.constants import FILL_ID
@@ -135,6 +136,13 @@ def make_claim_challenged(
 
 def make_context() -> Tuple[Context, Config]:
     checker = TokenChecker([])
+
+    target_descriptor = search(id=TARGET_CHAIN_ID)
+    source_descriptor = search(id=SOURCE_CHAIN_ID)
+    if not target_descriptor:
+        register(TARGET_CHAIN_ID, ChainDescriptor(TARGET_CHAIN_ID, "ethereum", "testnet"))
+    if not source_descriptor:
+        register(SOURCE_CHAIN_ID, ChainDescriptor(SOURCE_CHAIN_ID, "ethereum", "testnet"))
 
     chains = {}
     for chain_name in ("l2a", "l2b"):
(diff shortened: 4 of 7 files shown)


beamer/config/config.py
```diff
@@ -0,0 +1,55 @@
...
+        },
+        "polygon_zkevm": {
+            "mainnet": ChainId(1101),
+            "goerli": ChainId(1442),
+            "local": ChainId(1001),
+        },
+        "base": {"mainnet": None, "goerli": ChainId(84531), "local": None},
+    },
+    type2_chains=("ethereum", "optimism"),
+    optimistic_chains=("optimism", "base"),
```
fredo, line 54: would rename to `op_stack_chains` or even more explicit `bedrock_chains` if you check for that in `maybe_prove`

beamer/config/config.py
```diff
@@ -0,0 +1,55 @@
...
+            "local": ChainId(412346),
+        },
+        "polygon_zkevm": {
+            "mainnet": ChainId(1101),
+            "goerli": ChainId(1442),
+            "local": ChainId(1001),
+        },
+        "base": {"mainnet": None, "goerli": ChainId(84531), "local": None},
+    },
+    type2_chains=("ethereum", "optimism"),
```
fredo, line 53: would call it dynamic_fee_chains

fredo: base should also be in there

beamer/agent/chain.py
```diff
@@ -601,8 +601,7 @@ def maybe_prove(claim: Claim, context: Context) -> None:
     elif claim.invalidation_tx is not None:
         prove_tx = claim.invalidation_tx
 
-    # mainnet: 10, goerli: 420, local: 901
-    if context.target_chain.id not in (10, 420, 901):
+    if context.target_chain.id not in BEAMER_CONFIG.filter_chain_ids_by_name("optimism", False):
```
fredo, line 604: the maybe_prove should be called on bedrock chains

beamer/chains.py
```diff
@@ -0,0 +1,106 @@
...
+    testnet: bool = False
+    local: bool = False
+
+    @property
+    def supports_type_2_transactions(self) -> bool:
+        return self.type2 and not self.local
+
+    @property
+    def is_testnet(self) -> bool:
+        return self.testnet
```
istankovic, line 21: I think we don't need properties like these since we already have public fields.

beamer/chains.py
```diff
@@ -0,0 +1,106 @@
...
+@dataclass(frozen=True)
+class _BeamerConfig:
+    chains: dict[str, dict[str, ChainDescriptor]]
+
+    def get_descriptor(self, chain_id: ChainId) -> ChainDescriptor:
+        for networks in self.chains.values():
+            for descriptor in networks.values():
+                if descriptor.id == chain_id:
+                    return descriptor
+        return ChainDescriptor(id=chain_id, ecosystem_name="default", name="default")
```
istankovic, line 37: We don't really need the `_BeamerConfig` type, it does not do much anyway.

istankovic: Regarding the return value of `get_descriptor`, we have two options if the chain ID is not found:

1. returning `None`
2. returning an instance of `ChainDescriptor` that somehow indicates that this chain is not known to us

I think the agent actually drops things that are coming from unknown chains so I _think_ by the time this function is called, it should be the case that the chain is known. However, this module lives at `beamer.chains` and is expected to be used in a wider context i.e. not just the agent. So I think we should actaully return `None` here.

In fact,
```python
if get_chain_descriptor(chain_id) is not None:
```
can server as the `is this chain supported/know by us` check.

istankovic: One other thing here: I think it would be nice if, during execution of this module, the toplevel data that contains the descriptors were transformed in such a way so that it makes descriptor lookup O(1) instead of O(n).

beamer/chains.py
```diff
@@ -0,0 +1,106 @@
...
+            "mainnet": ChainDescriptor(
+                id=ChainId(42161), ecosystem_name="arbitrum", name="mainnet"
+            ),
+            "goerli": ChainDescriptor(
+                id=ChainId(421613), ecosystem_name="arbitrum", name="goerli", testnet=True
+            ),
+            "local": ChainDescriptor(
+                id=ChainId(412346), ecosystem_name="arbitrum", name="local", local=True
+            ),
+        },
```
istankovic, line 84: Seeing that there's a fair amount of repetition here, how about we use a plain tuple for the entire data, e.g.
```python
_DESCRIPTORS = (
   # Arbitrum
   ChainDescriptor(id=ChainId(42161), ecosystem_name="arbitrum", name="mainnet"),
   ChainDescriptor(id=ChainId(421613), ecosystem_name="arbitrum", name="goerli", testnet=True),
   ChainDescriptor(id=ChainId(412346), ecosystem_name="arbitrum", name="local", local=True),

...
)
_DESCRIPTORS = _prepare_for_lookups_DESCRIPTORS)
```
?

The `_prepare_for_lookups` (sorry, couldn't find a better name) would take the tuple data and convert it in a form better suited to lookups so we can have O(1) behaviour, as mentioned above.

beamer/tests/agent/unit/test_challenge.py
```diff
@@ -99,7 +100,7 @@ def test_join_false_claim_challenge_only_when_unfilled(filler):
 def test_optimism_prove(mocked_relayer_call):
     timestamp = int(time.time())
     mocked_relayer_call.return_value = str(timestamp)
-    op_chain_id = ChainId(901)
+    op_chain_id = BEAMER_CONFIG.chains["optimism"]["local"].id
```
istankovic, line 103: I think instead of that, which requires the chain descriptor data to be public, we should have an API for queries like this. For example,

```python
descriptor = beamer.chains.find_chain_descriptor(ecosystem_name="optimism", name="local")
assert descriptor is not None
op_chain_id = descriptor.id
```

istankovic (changes requested):
I think this is a step in the right direction, but some additional work is needed.

beamer/chains.py
```diff
@@ -0,0 +1,82 @@
...
+def get_chain_descriptor(chain_id: ChainId) -> ChainDescriptor | None:
+    return _CHAIN_DESCRIPTORS.get(chain_id)
+
+
+def search(**kwargs: Any) -> tuple[ChainDescriptor, ...]:
+    found = []
+    for desc in DESCRIPTORS:
+        for field, value in kwargs.items():
+            if getattr(desc, field, None) == value:
+                found.append(desc)
```
istankovic, line 81: I think we want to append to `found` only if all fields from `kwargs` match, not just one:
```python
if all(getattr(desc, field, None) == value for field, value in kwargs.items()):
   found.append(desc)
```

beamer/chains.py
```diff
@@ -0,0 +1,82 @@
...
+    ChainDescriptor(id=ChainId(412346), ecosystem_name="arbitrum", name="local", local=True),
+    # Polygon ZkEVM
+    ChainDescriptor(id=ChainId(1101), ecosystem_name="polygon_zkevm", name="mainnet"),
+    ChainDescriptor(id=ChainId(1442), ecosystem_name="polygon_zkevm", name="goerli", testnet=True),
+    ChainDescriptor(id=ChainId(1001), ecosystem_name="polygon_zkevm", name="local", local=True),
+    # Base
+    ChainDescriptor(
+        id=ChainId(84531), ecosystem_name="base", name="goerli", testnet=True, bedrock=True
+    ),
+)
```
istankovic, line 59: Nitpick: I think it's unfortunate that black reformats these lines as I think they are not _too_ long and would be easier to read if black were to just ignore this entire block. This can be done like
```python
# fmt: off
_DESCRIPTORS = ...
# fmt: on
```
per https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#code-style

beamer/chains.py
bilbeyt: :D lol i was meant to do that (appending after everything is found)

beamer/agent/chain.py
```diff
@@ -601,8 +601,12 @@ def maybe_prove(claim: Claim, context: Context) -> None:
     elif claim.invalidation_tx is not None:
         prove_tx = claim.invalidation_tx
 
-    # mainnet: 10, goerli: 420, local: 901
-    if context.target_chain.id not in (10, 420, 901):
+    chain_descriptor = get_chain_descriptor(context.target_chain.id)
+
+    if chain_descriptor is None:
+        return
```
istankovic, line 607: At this point it is guaranteed that `chain_descriptor` cannot be `None`, right?
If it was `None`, that means that a request for an unsupported chain somehow got through our machinery to `maybe_prove`, but I was convinced that we filter out such requests/events much earlier, when we get them.
I would appreciate if you could check that and turn this check rather into an `assert` statement.

beamer/chains.py
```diff
@@ -0,0 +1,82 @@
...
+    id: ChainId
+    ecosystem_name: str
+    name: str
+    type2: bool = False
+    bedrock: bool = False
+    testnet: bool = False
+    local: bool = False
+
+
+DESCRIPTORS = (
```
istankovic, line 18: Is there a reason that this is publc? If we don't need it, I'd rather that it is private; even more so since it's mutable.

beamer/chains.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def _prepare_descriptors_for_lookup() -> dict[ChainId, ChainDescriptor]:
+    descriptors = {}
+    for descriptor in DESCRIPTORS:
+        descriptors[descriptor.id] = descriptor
+    return descriptors
+
+
+_CHAIN_DESCRIPTORS = _prepare_descriptors_for_lookup()
```
istankovic, line 69: We don't actually need to have both the tuple and the dict keyed by the chain ID, it is sufficient to only have the latter. The `search` function can then just iterate over dict's `values()` and we get  the same behaviour as if we're iterating over the tuple.

beamer/tests/agent/unit/test_challenge.py
```diff
@@ -99,7 +100,7 @@ def test_join_false_claim_challenge_only_when_unfilled(filler):
 def test_optimism_prove(mocked_relayer_call):
     timestamp = int(time.time())
     mocked_relayer_call.return_value = str(timestamp)
-    op_chain_id = ChainId(901)
+    op_chain_id = next(iter(search(bedrock=True, local=True))).id
```
istankovic, line 103: This looks pretty good!

beamer/util.py
```diff
@@ -162,8 +163,8 @@ def make_web3(
     chain_id = ChainId(w3.eth.chain_id)
 
     # Apply type 2 transaction middleware for ETH2 PoS chains
-    # Currently Mainnet and Goerli
-    if chain_id in [1, 5]:
+    chain_descriptor = get_chain_descriptor(chain_id)
+    if chain_descriptor is not None and chain_descriptor.type2 and not chain_descriptor.local:
```
istankovic, line 167: Why the `not chain_descriptor.local` part?

bilbeyt: because we are not supporting base fee for local ?

beamer/chains.py
bilbeyt: but you said tuple first :D

istankovic: Right and it's totally fine to have tuple as an input as it is now, my comment above is about the fact that we have two module-level objects with the same data, just in different form. So what I was trying to say is, we only need one of those two -- the dict which is a result of the prepare function.

I meant something like
```
_DESCRIPTORS = (...)   # tuple
_DESCRIPTORS = _prepare(_DESCRIPTORS)  # now it's a dict and we only have one copy of the data
```
which could even be done as
```
_DESCRIPTORS = _prepare(
   ChainDescriptor(...),
   ChainDescriptor(...)
...
)
```

beamer/util.py
istankovic: This is because we're enforcing Istanbul fork for local ganache, right?

beamer/chains.py
bilbeyt: it is long :D at least flake8 says also :D

bilbeyt: i am keeping formatting here

beamer/util.py
bilbeyt: yes and if we ever allow it all fee calculations will break

beamer/chains.py
bilbeyt: made it dictionary directly without any tuple now

beamer/chains.py
```diff
@@ -0,0 +1,87 @@
...
+def search(**kwargs: Any) -> tuple[ChainDescriptor, ...]:
+    found = []
+    for desc in _DESCRIPTORS.values():
+        if all(getattr(desc, field, None) == value for field, value in kwargs.items()):
+            found.append(desc)
+    return tuple(found)
+
+
+def register(chain_id: ChainId, ecosystem_name: str, name: str, **kwargs: Any) -> None:
+    _DESCRIPTORS[chain_id] = ChainDescriptor(chain_id, ecosystem_name, name, **kwargs)
```
istankovic, line 87: We should `assert` that the chain ID is not already registered, in order to prevent overwriting.

bilbeyt: then we need to deregister or something in my opinion because tests may use the same chain id but maybe with different flags

istankovic: > then we need to deregister or something in my opinion because tests may use the same chain id but maybe with different flags

Hm, that's a good point, yeah.

One thing that came to my mind yesterday is that we should probably go with registering an entire `ChainDescriptor` object, rather than allowing arbitrary kwargs as we do currently.
E.g. 
```python
def register(chain_id: ChainId, descriptor: ChainDescriptor) -> None:
```

beamer/tests/agent/unit/utils.py
```diff
@@ -136,6 +137,13 @@ def make_claim_challenged(
 def make_context() -> Tuple[Context, Config]:
     checker = TokenChecker([])
 
+    target_descriptor = search(id=TARGET_CHAIN_ID)
+    source_descriptor = search(id=SOURCE_CHAIN_ID)
+    if len(target_descriptor) == 0:
+        register(TARGET_CHAIN_ID, ChainDescriptor(TARGET_CHAIN_ID, "ethereum", "testnet"))
```
istankovic, line 143: Checking `len(target_descriptor) == 0` is not really idiomatic Python, we should rather use `not ... ` here, e.g.
```python
    if not search(id=TARGET_CHAIN_ID):
        register(TARGET_CHAIN_ID, ChainDescriptor(TARGET_CHAIN_ID, "ethereum", "testnet"))
```

beamer/tests/agent/unit/utils.py
```diff
@@ -136,6 +137,13 @@ def make_claim_challenged(
...
+    if len(source_descriptor) == 0:
+        register(SOURCE_CHAIN_ID, ChainDescriptor(SOURCE_CHAIN_ID, "ethereum", "testnet"))
```
istankovic, line 145: Similarly here.

beamer/chains.py
```diff
@@ -0,0 +1,96 @@
...
+    ChainId(42161): ChainDescriptor(id=ChainId(42161), ecosystem_name="arbitrum", name="mainnet"),
+    ChainId(421613): ChainDescriptor(
+        id=ChainId(421613), ecosystem_name="arbitrum", name="goerli", testnet=True
+    ),
+    ChainId(412346): ChainDescriptor(
+        id=ChainId(412346), ecosystem_name="arbitrum", name="local", local=True
+    ),
+    # Polygon ZkEVM
+    ChainId(1101): ChainDescriptor(
+        id=ChainId(1101), ecosystem_name="polygon_zkevm", name="mainnet"
```
istankovic, line 62: Does Polygon zkEVM not support type 2 tx?

beamer/chains.py
```diff
@@ -0,0 +1,96 @@
...
+    ),
+    ChainId(1442): ChainDescriptor(
+        id=ChainId(1442), ecosystem_name="polygon_zkevm", name="goerli", testnet=True
+    ),
+    ChainId(1001): ChainDescriptor(
+        id=ChainId(1001), ecosystem_name="polygon_zkevm", name="local", local=True
+    ),
+    # Base
+    ChainId(84531): ChainDescriptor(
+        id=ChainId(84531), ecosystem_name="base", name="goerli", testnet=True, bedrock=True
```
istankovic, line 72: Does Base testnet not support type 2 tx?

beamer/chains.py
```diff
@@ -0,0 +1,96 @@
...
+    return tuple(found)
+
+
+def register(chain_id: ChainId, descriptor: ChainDescriptor) -> None:
+    assert chain_id not in _DESCRIPTORS, "Chain is already added"
+    _DESCRIPTORS[chain_id] = descriptor
+
+
+def unregister(chain_id: ChainId) -> None:
+    assert chain_id in _DESCRIPTORS, "Chain record not found"
```
istankovic, line 95: Nitpick: we can omit this `assert` since `del` below will raise if there's no such ID registered.

istankovic (approved):

istankovic:
Great work!

Issue #2022, opened 2023-07-06
Agent needs to parse timestamp from approve call of relayer


bilbeyt:
As we are changing relayer we need to set timestamp correctly in agent.

Acceptance Criteria:
* [ ] Agent uses correct approval timestamp.

PR #2027, opened 2023-07-06
fix: use timestamp from relayer

bilbeyt:
Closes: #2022

### DIFF
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -632,16 +632,16 @@ def maybe_prove(claim: Claim, context: Context) -> None:
 
     def on_future_done(f: Future) -> None:
         try:
-            f.result()
+            timestamp = f.result()
         except Exception as ex:
             context.logger.error("Optimism prove failed", ex=ex, tx_hash=prove_tx)
         else:
             assert request is not None
             claim.proved_tx = prove_tx
             if request.fill_tx == prove_tx:
-                request.fill_timestamp = Timestamp(int(time.time()))
+                request.fill_timestamp = Timestamp(int(timestamp))
             elif claim.invalidation_tx == prove_tx:
-                claim.invalidation_timestamp = Timestamp(int(time.time()))
+                claim.invalidation_timestamp = Timestamp(int(timestamp))
         finally:
             assert prove_tx is not None
             context.l1_resolutions.pop(prove_tx, None)
--- a/beamer/agent/relayer.py
+++ b/beamer/agent/relayer.py
@@ -1,12 +1,12 @@
+import re
 import subprocess
 import sys
-
 from pathlib import Path
+
 import structlog
 from hexbytes import HexBytes
 
 from beamer.typing import URL
-
 
 log = structlog.get_logger(__name__)
 
@@ -30,7 +30,7 @@ def run_relayer_for_tx(
     privkey: HexBytes,
     tx_hash: HexBytes,
     prove_tx: bool = False,
-) -> None:
+) -> str | None:
     relayer = get_relayer_executable()
 
     if not relayer.exists():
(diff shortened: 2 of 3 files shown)


beamer/agent/chain.py
```diff
@@ -632,16 +632,17 @@ def maybe_prove(claim: Claim, context: Context) -> None:
 
     def on_future_done(f: Future) -> None:
         try:
-            f.result()
+            result = f.result()
         except Exception as ex:
             context.logger.error("Optimism prove failed", ex=ex, tx_hash=prove_tx)
         else:
+            timestamp = result.stdout.split()[-1].split(": ")[-1]
```
GabrielBuragev, line 639: Does this mean, take the last line of the `stdout` and split it by ":" and take the last value from the split?

If so, i would advise to be more explicit and try to search through all the `stdout` lines to find an exact match of the line "Proof timestamp:*". before trying to split it by ":". 
If the `prove-op-message` command changes in a way that it prints anything else after the `Proof timestamp` line then this will break without us noticing.

istankovic: > Does this mean, take the last line of the `stdout` and split it by ":" and take the last value from the split?
> 
> If so, i would advise to be more explicit and try to search through all the `stdout` lines to find an exact match of the line "Proof timestamp:*". before trying to split it by ":". If the `prove-op-message` command changes in a way that it prints anything else after the `Proof timestamp` line then this will break without us noticing.

:100:

beamer/tests/agent/unit/test_challenge.py
```diff
@@ -93,8 +96,13 @@ def test_join_false_claim_challenge_only_when_unfilled(filler):
         assert not claim.transaction_pending
 
 
-@patch("beamer.agent.relayer.run_relayer_for_tx")
+@patch("beamer.agent.chain.run_relayer_for_tx")
```
istankovic, line 99: Why this change?

beamer/agent/chain.py
istankovic: I would also prefer if this timestamp parsing was handled by the `relayer` Python module, inside `run_relayer_for_tx`. Because that is where the interface to the relayer binary is.

istankovic (changes requested):
Needs updating, please see comments.

beamer/tests/agent/unit/test_challenge.py
bilbeyt: mock wasnt working before :D my bad :D

beamer/agent/relayer.py
```diff
@@ -30,7 +30,7 @@ def run_relayer_for_tx(
     privkey: HexBytes,
     tx_hash: HexBytes,
     prove_tx: bool = False,
-) -> None:
+) -> str | None:
```
istankovic, line 33: Suppose we return `None` from this function and there is no exception.
What would happen in `chain.py`?

beamer/agent/relayer.py
```diff
@@ -60,9 +60,14 @@ def run_relayer_for_tx(
...
-    subprocess.run(
+    result = subprocess.run(
         command_args,
         capture_output=True,
         encoding="utf-8",
         check=True,  # check throws an error right away
     )
+
+    if prove_tx:
+        return result.stdout.split()[-1].split(": ")[-1]
```
istankovic, line 71: This does not address @GabrielBuragev's concern.

I suggest something like the following (totally untested, please adjust accordingly):
```python
match = re.search('Proof timestamp: (\d+)', result.stdout)
if match is not None:
    return match.group(1)
raise RuntimeError("relayer failed to provide proof timestamp")
```

The reason for the exception being that code in `chain.py` can react to it.

beamer/agent/relayer.py
bilbeyt: this function is also used by l1 resolution so returning None should be fine.

bilbeyt: only one prove_tx condition is determining if it is prove or l1 resolution

beamer/tests/agent/unit/test_challenge.py
```diff
@@ -93,8 +95,10 @@ def test_join_false_claim_challenge_only_when_unfilled(filler):
...
 def test_optimism_prove(_mocked_relayer_call):
+    timestamp = int(time.time())
+    _mocked_relayer_call.return_value = str(timestamp)
```
istankovic, line 101: Nitpick: we should remove the leading underscore from the name, we are actually using it now.

istankovic (approved):
Looks good, thanks!

Issue #2020, opened 2023-07-05
Implement contract for fee subsidy


fredo:
## FeeSub.sol
* The owner needs to be able to enable/disable token addresses to be subsidized
* The owner needs to be able to change the minimum amount
* The contract needs to be able to approve the RM to take tokens from FeeSub
* User must be able to create a request if subsidizable and withdraw an expired request
* The owner must not be able to withdraw funds belonging to the user

## Implementation
* Mapping for address to minimumAmount
  * If 0, then it's not enabled to be subsidized
* setMinimumAmount(tokenAddress, minimumAmount)
  * onlyOwner
  * If set to > 0, approve RM to maximum value
  * If set to 0, remove allowance for RM
* createRequest
  * check if >= minimum amount
  * transferFrom user
  * createRequest
  * store requestId => sender
* withdrawExpiredRequest
  * check if request sender is FeeSub on RM
  * check if request withdrawnClaimId is maxInt (withdrawn by withdrawExpiredRequest)
    * if not withdrawExpiredRequest
  * check if requestId => is address(0)
    * if not set to address(0)
    * send request.amount back to user
* Funder accounting
  * TBD

andfletcher:
let's go the easy way, the owner is not able to withdraw what is in the contract. we can figure out how to get tokens out of the contract later.

PR #2028, opened 2023-07-06
FeeSub contract

bilbeyt:
Close: #2020

### DIFF
--- a/beamer/tests/contracts/test_fee_sub.py
+++ b/beamer/tests/contracts/test_fee_sub.py
@@ -1,0 +1,159 @@
+import ape
+from hexbytes import HexBytes
+from web3.constants import ADDRESS_ZERO
+
+from beamer.tests.agent.utils import make_address
+from beamer.tests.util import alloc_accounts, alloc_whitelisted_accounts, make_request
+
+
+def test_fee_sub(fee_sub, token, request_manager):
+    mint_amount = 5
+    sent_amount = 2
+    chain_id = ape.chain.chain_id
+    (sender, target) = alloc_accounts(2)
+
+    wrong_token_address = make_address()
+
+    total_fee = request_manager.totalFee(chain_id, token.address, sent_amount)
+    old_contract_balance = token.balanceOf(fee_sub.address)
+
+    with ape.reverts("Token not to be subsidized"):
+        fee_sub.createRequest(
+            chain_id,
+            wrong_token_address,
+            wrong_token_address,
+            target,
+            sent_amount,
+            1800,
+            sender=sender,
+        )
+
+    with ape.reverts("Transfer amount too small to be subsidized"):
+        fee_sub.createRequest(
+            chain_id, token.address, token.address, target, 1, 1800, sender=sender
+        )
+
+    token.mint(sender, mint_amount, sender=sender)
+
+    with ape.reverts("Insufficient allowance"):
+        fee_sub.createRequest(
+            chain_id, token.address, token.address, target, 2, 1800, sender=sender
+        )
+
+    token.approve(fee_sub.address, mint_amount, sender=sender)
+
+    assert token.balanceOf(sender) == mint_amount
+
+    request = fee_sub.createRequest(
+        chain_id, token.address, token.address, target, sent_amount, 1800, sender=sender
+    )
+
+    request_id = request.return_value
+
+    assert token.balanceOf(sender) == mint_amount - sent_amount  # no fee
+    assert token.balanceOf(fee_sub.address) == old_contract_balance - total_fee
+    assert (
+        request_manager.requests(request_id).sender == fee_sub.address
+    )  # sender is fee sub contract
+
+
+def test_expired_request_withdraw(fee_sub, token):
+    mint_amount = 5
+    sent_amount = 2
+    chain_id = ape.chain.chain_id
+    (sender, target) = alloc_accounts(2)
+
+    token.mint(sender, mint_amount, sender=sender)
+    token.approve(fee_sub.address, mint_amount, sender=sender)
+
+    request = fee_sub.createRequest(
+        chain_id, token.address, token.address, target, sent_amount, 1800, sender=sender
+    )
+
+    request_id = request.return_value
+
+    assert token.balanceOf(sender) == mint_amount - sent_amount
+
+    ape.chain.mine(deltatime=1800)
+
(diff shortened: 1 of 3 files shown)


contracts/contracts/FeeSub.sol
```diff
@@ -0,0 +1,122 @@
...
+            targetAddress,
+            amount,
+            validityPeriod
+        );
+        senders[requestId] = msg.sender;
+
+        return requestId;
+    }
+
+    function transferableAmount(
```
fredo, line 62: We figured out that we don't need this function

contracts/contracts/FeeSub.sol
```diff
@@ -0,0 +1,122 @@
...
+            ,
+            ,
+            ,
+
+        ) = requestManager.requests(requestId);
+        require(
+            sender == address(this),
+            "Request was not created by this contract"
+        );
+        require(senders[requestId] != address(0), "Already Withdrawn");
```
fredo, line 97: nitpick `Already withdrawn`

contracts/contracts/FeeSub.sol
```diff
@@ -0,0 +1,122 @@
...
+
+        IERC20 token = IERC20(sourceTokenAddress);
+        token.safeTransfer(recipient, amount);
+    }
+
+    function setMinimumAmount(
+        address tokenAddress,
+        uint256 amount
+    ) external onlyOwner {
+        if (amount > 0) {
```
fredo, line 112: We can save some gas here and ask if minimumAmounts[tokenAddress] is already > 0, then we know that we already approved and don't need to approve again.

contracts/contracts/FeeSub.sol
```diff
@@ -0,0 +1,122 @@
...
+            IERC20(tokenAddress).approve(
+                address(requestManager),
+                type(uint256).max
+            );
+        } else if (amount == 0) {
```
fredo, line 117: same here if it's already set to 0 (which is kinda nonsense)

contracts/contracts/FeeSub.sol
```diff
@@ -0,0 +1,122 @@
...
+            ,
+
+        ) = requestManager.requests(requestId);
+        require(
+            sender == address(this),
+            "Request was not created by this contract"
+        );
+        require(senders[requestId] != address(0), "Already Withdrawn");
+
+        requestManager.withdrawExpiredRequest(requestId);
```
fredo, line 99: Here we need to add another condition. It can be the case that the request was already withdrawn by calling the RequestManager directly. 
If that's the case this call would revert and the user's money would be stuck inside this contract forever. 

I suggest that we do this check by checking the withdrawnClaimId of the request. 
We should only call `requestManager.withdrawExpiredRequest(requestId);` if withdrawnClaimId is 0.

fredo: Speaking of which, we also need an additional require that the tokens have not been withdrawn by the agent. 
IOW, we should only ever proceed if we either can call `withdrawExpiredRequest` or we are sure that the money was already withdrawn, but to this contract


Suggestion: 

```sol
if(withdrawClaimId == 0) {
    // This will fail if the funds do not belong to the user (yet)
    requestManager.withdrawExpiredRequest(requestId);
}
else {
    // Make sure that funds were withdrawn by calling withdrawExpiredRequest()
    require(withdrawnClaimId == type(uint96).max);

}


```

contracts/contracts/FeeSub.sol
GabrielBuragev: I am not sure about this one. Can you remind me why we decided to not use this fn ?

contracts/contracts/FeeSub.sol
```diff
@@ -0,0 +1,114 @@
...
+            ,
+            uint96 withdrawClaimId,
+            ,
+
+        ) = requestManager.requests(requestId);
+        require(
+            sender == address(this),
+            "Request was not created by this contract"
+        );
+        require(senders[requestId] != address(0), "Already withdrawn");
```
fredo, line 80: I would rephrase it to "already refunded to the sender".
The rationale behind this is that we should have a distinction between withdrawing from the request manager and withdrawing from this contract.

Below there is `Already withdrawn by agent` -> I would this rename to `request is withdrawn by another address` (wouldnt mention the notion agent here as agent refers more to the software component)

contracts/contracts/FeeSub.sol
bilbeyt: @fredo

fredo: already explained him in DM. 
We do not need this function cause it's trivial. If we were to use this contract (which is if transfer amount >= minimumAmount) then the transferable amount = transfer amount. 
If the condition is false we are not using the contract anyway.

beamer/tests/contracts/test_fee_sub.py
```diff
@@ -0,0 +1,155 @@
...
+
+    assert token.balanceOf(sender) == mint_amount
+
+    request = fee_sub.createRequest(
+        chain_id, token.address, token.address, target, sent_amount, 1800, sender=sender
+    )
+
+    request_id = request.return_value
+
+    assert token.balanceOf(sender) == mint_amount - sent_amount  # no fee
```
fredo, line 50: can you also assert that the `fee_sub_balance == old_balance - sent_amount - total_fee(sent_amount)`

contracts/contracts/FeeSub.sol
```diff
@@ -0,0 +1,117 @@
...
+        );
+
+        if (withdrawClaimId == 0) {
+            // This will fail if the funds do not belong to the original sender (yet)
+            requestManager.withdrawExpiredRequest(requestId);
+        } else {
+            // Make sure that funds were withdrawn by calling withdrawExpiredRequest()
+            require(
+                withdrawClaimId == type(uint96).max,
+                "Request is withdrawn by another address"
```
fredo, line 92: nitpick: `was withdrawn`

beamer/tests/contracts/test_fee_sub.py
bilbeyt: is this really true? old balance - total_fee is ok to me and for the sent_amount we are already getting it and sending it so no storing in there

fredo: ah you are right

fredo (approved):
looks good 👍

PR #2044, opened 2023-07-11
Rename Backlog Item.md to Backlog_Item.md

andfletcher:
changed titel so template is shown


compojoom:
the other items are all lowercase and have a hyphen and not an underscore:
image

could you change it so that it is the same as the other items?

compojoom:
closing in favor of https://github.com/beamer-bridge/beamer/pull/2047

PR #2045, opened 2023-07-12
frontend: Update transfer success tweet text for Polygon release

manuelwedler:
See #1684

Should only be merged shortly before we go live on Polygon.

### DIFF
--- a/frontend/src/components/ShareTweet.vue
+++ b/frontend/src/components/ShareTweet.vue
@@ -7,7 +7,7 @@
 <script lang="ts" setup>
 import { computed, toRef } from 'vue';
 
-import { Transfer } from '@/actions/transfers';
+import { isSubsidizedTransfer, Transfer } from '@/actions/transfers';
 
 const props = defineProps({
   transfer: Transfer,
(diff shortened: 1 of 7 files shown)


manuelwedler:
Yes, already implemented it on top of your PR. Will updated this PR once the other is merged

manuelwedler:
Now this is ready for review.

hint: I removed the `ExtendedTransferData` because it was applied to `TransferData` objects and potentially made a `feeSubAddress` available which does not exist. I think using type guards is the safer approach.

frontend/src/components/ShareTweet.vue
```diff
@@ -25,16 +25,36 @@ const isShareable = computed(() => {
...
+
+Unlock lightning-fast and secure bridging with Beamer today 💪💫  
+https://app.beamerbridge.com/
+`;
+
+  let text: string;
+  if (isSubsidizedTransfer(transfer.value) && transfer.value.fees.uint256.isZero()) {
+    text = subsidizedTransferText;
+  } else {
+    text = defaultText;
```
GabrielBuragev, line 54: Nitpick: We can simply write `return defaultText;` and `return subsidizedTransferText;` here.

GabrielBuragev (approved):
Thanks for implementing the type guards. Indeed, this is the proper way to do the checks against the different types of transfers.

One small nitpick only.

GabrielBuragev (changes requested):
I just noticed this while testing: https://github.com/beamer-bridge/beamer/issues/1684#issuecomment-1633908079

GabrielBuragev (approved):
Looks perfect now.

PR #2072, opened 2023-07-23
frontend: extend subsidized transfers with missing functionality

GabrielBuragev:
There are still no tests for this file as it requires some effort.

### DIFF
--- a/frontend/src/actions/transfers/subsidized-transfer.ts
+++ b/frontend/src/actions/transfers/subsidized-transfer.ts
@@ -1,36 +1,72 @@
+import { sendRequestTransaction, withdrawRequest } from '@/services/transactions/request-manager';
+import { ensureTokenAllowance, isAllowanceApproved } from '@/services/transactions/token';
 import type { IEthereumProvider } from '@/services/web3-provider';
+import type { EthereumAddress } from '@/types/data';
 import { UInt256 } from '@/types/uint-256';
 
 import { type TransferData, Transfer } from './transfer';
 
 export class SubsidizedTransfer extends Transfer {
   readonly feeSubAddress: string;
-  readonly originalRequestManagerAddress: string;
 
   constructor(data: SubsidizedTransferData) {
     if (!data.sourceChain.feeSubAddress) {
       throw new Error('Please provide a fee subsidy contract address.');
     }
     super(data);
-    this.originalRequestManagerAddress = data.sourceChain.requestManagerAddress;
     this.feeSubAddress = data.sourceChain.feeSubAddress;
     Object.assign(this.fees.uint256, this.fees.uint256.multiply(new UInt256(0)));
   }
 
-  async ensureTokenAllowance(provider: IEthereumProvider): Promise<void> {
-    if (this.feeSubAddress) {
-      this.sourceChain.requestManagerAddress = this.feeSubAddress;
-    }
-    await super.ensureTokenAllowance(provider);
-    this.sourceChain.requestManagerAddress = this.originalRequestManagerAddress;
+  protected async callWithdrawRequest(
+    provider: IEthereumProvider,
+    requestIdentifier: string,
+  ): Promise<void> {
+    await withdrawRequest(provider, this.feeSubAddress, requestIdentifier);
   }
 
-  async sendRequestTransaction(provider: IEthereumProvider): Promise<void> {
-    if (this.feeSubAddress) {
-      this.sourceChain.requestManagerAddress = this.feeSubAddress;
-    }
-    await super.sendRequestTransaction(provider);
-    this.sourceChain.requestManagerAddress = this.originalRequestManagerAddress;
+  protected async callEnsureTokenAllowance(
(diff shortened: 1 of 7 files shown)


frontend/src/actions/transfers/subsidized-transfer.ts
```diff
@@ -33,6 +33,22 @@ export class SubsidizedTransfer extends Transfer {
     this.sourceChain.requestManagerAddress = this.originalRequestManagerAddress;
   }
 
+  async withdraw(provider: IEthereumProvider): Promise<void> {
+    if (this.feeSubAddress) {
+      this.sourceChain.requestManagerAddress = this.feeSubAddress;
+    }
+    await super.withdraw(provider);
+    this.sourceChain.requestManagerAddress = this.originalRequestManagerAddress;
+  }
```
istankovic, line 42: Is there something like Python's `try...finally` in JS?
What happens if `withdraw` throws an exception that is caught in one of the callers?

frontend/src/actions/transfers/subsidized-transfer.ts
```diff
@@ -33,6 +33,22 @@ export class SubsidizedTransfer extends Transfer {
     this.sourceChain.requestManagerAddress = this.originalRequestManagerAddress;
   }
 
+  async withdraw(provider: IEthereumProvider): Promise<void> {
+    if (this.feeSubAddress) {
```
fredo, line 37: Does this take into account that the fee sub contract is only called upon requests which were created by the feesub?

frontend/src/actions/transfers/subsidized-transfer.ts
GabrielBuragev: Thats a very good point Ivan. I will wrap it inside a try/catch block.

frontend/src/actions/transfers/subsidized-transfer.ts
GabrielBuragev: Well, the fact that we are inside a `SubsidizedTransfer` instance is implying that we used a FeeSub contract for that same request.
Whether we should create a normal `Transfer` or a `SubsidizedTransfer` is decided here: https://github.com/beamer-bridge/beamer/blob/main/frontend/src/composables/useTransferRequest.ts#L57-L69

fredo: ok makes sense 👍

manuelwedler:
I aligned with Gabriel and due to the issues of overwriting the request manager address, I reimplemented the `SubsidizedTransfer`. This can be reviewed again now.

GabrielBuragev (commented):
The way you implemented it looks perfect!

After thinking about the other possible ways to do it i realized we can avoid repeating ourselves (overriding the `call...` functions) by creating an abstraction around the functions inside the `request-manager.ts`. 
This abstraction will contain a `contractAddress` property that will reference to the on-chain contract that is initialized in the constructor. With this, we will only carry one instance of this class inside the `Transfer` or `SubsidizedTransfer` classes and call all the functions via that instance instead of using these functions as separate units.
This way, the `SubsidizedTransfer` will also look way leaner.
Looking forward to hear your thoughts on this. 

I anyways accept your implementation as it is perfect. If you  like my idea too, we can create an issue and optimize it in the future.

manuelwedler (approved):
Approving as @GabrielBuragev reviewed and he cannot give an approval for his own PR

manuelwedler:
> The way you implemented it looks perfect!
> 
> After thinking about the other possible ways to do it i realized we can avoid repeating ourselves (overriding the `call...` functions) by creating an abstraction around the functions inside the `request-manager.ts`. This abstraction will contain a `contractAddress` property that will reference to the on-chain contract that is initialized in the constructor. With this, we will only carry one instance of this class inside the `Transfer` or `SubsidizedTransfer` classes and call all the functions via that instance instead of using these functions as separate units. This way, the `SubsidizedTransfer` will also look way leaner. If i am not wrong, it is also more memory efficient. Looking forward to hear your thoughts on this.
> 
> I anyways accept your implementation as it is perfect. If you like my idea too, we can create an issue and optimize it in the future.

@GabrielBuragev 
This would mean, that we create a class inside the `request-manager.ts` that would wrap the transaction functions in order to hold state (the `contractAddress`), right? The transaction functions were written in a functional programming style on purpose in the first place. The reasons were simplicity and easier testing.

There are even cases when we want to use different addresses of the request manager for different functions (e.g. fee sub). I think we would run into the same problem as in my comment above regarding the `checkAndUpdateState` function of the `Transfer`.

In my opinion, it is better to have this separation: `Transfer` / `SubsidizedTransfer` holds state, `request-manager` functions (and other contract functions) are pure.

manuelwedler:
Ah ok that makes more sense. But still we would have the issue about the `checkAndUpdateState` function. It should not use the fee sub address. Or how would that be solved if you only have one instance of the new class?

GabrielBuragev:
Ah true, somehow missed that. You are right, sorry for the confusion.

PR #2084, opened 2023-07-25
Frontend: update campaign texts

GabrielBuragev:
Based on the internal discussion, we are using a similar base text for both the zebra and the polygon zkevm campaign but render the zebra one whenever a transfer has been subsidized.


manuelwedler (approved):
Looks good! Nice that you got rid of the duplicate tweet text in the test.

PR #2087, opened 2023-07-25
beamer: middleware: fix cache_get_block_by_number

istankovic:
There are two problems being fixed here.

The first one is that the type of i JSON response's `number` field is "quantity", which is represented as a string containing hex-encoded quantity. We were calling built-in function hex() on it, which was wrong an unnecessary.

The second one is comparison of block numbers in _BlockCache, which meant that we compared block numbers lexicographically. Instead, we now convert the strings to numbers and compare numbers.


GabrielBuragev (approved):
LGTM

compojoom (approved):

PR #2094, opened 2023-07-28
frontend: enable Base

GabrielBuragev:
**Note: Do not merge before we are ready to announce and launch Base.**

One thing missing here. We need to include the FeeSub contract address in the `config/chains/8453.json` file if we want to enable fee subsidy for Base as well.

### DIFF
--- a/frontend/config/chains/8453.json
+++ b/frontend/config/chains/8453.json
@@ -10,6 +10,5 @@
     "symbol": "ETH",
     "decimals": 18
   },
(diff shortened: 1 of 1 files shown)


compojoom:
@GabrielBuragev is this really everything? the pinned @beamer-bridge/deployments package is 2.0.1, but the mainnet artifacts for base are first available in [2.0.3-eb1a5011-1690562106.0](https://www.npmjs.com/package/@beamer-bridge/deployments/v/2.0.3-eb1a5011-1690562106.0)

So running yarn configure doesn't actually include base as a network despite the fact that you've removed the "hidden": true flag?

GabrielBuragev:
closing in favor of #2177

PR #2111, opened 2023-08-03
beamer: agent: take the fill mutex before doing anything else when fi…

istankovic:
…lling

Otherwise it can happen that we do the timing checks before we obtain the mutex, and we are forced to wait on the mutex long enough that the timing check performed previously becomes invalid. Same reasoning applies to other checks, e.g. the min_source_balance, token allowance check etc.

### DIFF
--- a/beamer/agent/chain.py
+++ b/beamer/agent/chain.py
@@ -337,6 +337,21 @@ def process_claims(context: Context) -> None:
 
 
 def fill_request(request: Request, context: Context) -> None:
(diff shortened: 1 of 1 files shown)


compojoom:
Looks good to me, but why is the test failing?

istankovic:
> Looks good to me, but why is the test failing?

Because it's broken. I have removed it now.

compojoom (approved):

Issue #2007, opened 2023-07-03
Research if we need a specified Base e2e test


fredo:
Depending on #2006 we need to see if it is necessary to have our own e2e test for the Base rollup. 

Looking at the repository, the node on base is a pretty up to date op-geth fork. But I'd say it depends on the codebase of the relayer if a specific local e2e for base is needed

GabrielBuragev:
@fredo can you share a link to the repository please?

PR #2120, opened 2023-08-07
Add base e2e test

bilbeyt:
Closes: #2007


deployments/config/local/901-base.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "base",
+    "chain_id": 901,
+    "l1_messenger": ["OptimismL1Messenger", ${l1_messenger_args}],
+    "l2_messenger": "OptimismL2Messenger",
+    "finality_period": 3600,
```
istankovic, line 6: Why finality period of 3600s?

bilbeyt: just same with op

istankovic (approved):
Looks good!

Issue #2102, opened 2023-08-02
Relayer: rework the way relayer is passed the key by the agent


istankovic:
Currently the agent just hands out the private key as an option to the relayer. However, when reporting errors, the command will end up in agent's log, which is not good. We need to find a way to invoke the relayer without the key (nor the password) ending up in the log.

istankovic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-62442a69c5e6e60010e31cf3/issues/beamer-bridge/beamer/2102?planning-poker) with Zenhub @bilbeyt @compojoom @fredo

PR #2126, opened 2023-08-09
feat: use account path for relayer

bilbeyt:
Closes: #2102


beamer/agent/relayer.py
```diff
@@ -37,40 +53,49 @@ def run_relayer_for_tx(
...
+                capture_output=True,
+                encoding="utf-8",
+                check=True,  # check throws an error right away
+            )
+        except subprocess.CalledProcessError as ex:
+            relayer_args = ex.args[1]
+            for i, arg in enumerate(relayer_args):
+                if arg == "--password":
+                    ex.args[1][i + 1] = "<REDACTED>"
+                    break
```
istankovic, line 93: Could be more concise, without the `for`:
```python
i = relayer_args.find('--password')
relayer_args[i] = ...
```

beamer/agent/relayer.py
```diff
@@ -23,11 +29,21 @@ def get_relayer_executable() -> Path:
     return path.resolve()
 
 
+@contextlib.contextmanager
+def account_store(account: LocalAccount) -> Generator[tuple[str, str], None, None]:
+    password = uuid.uuid4().hex
+    data = account.encrypt(password)
+    with tempfile.TemporaryFile() as f:
+        f.write(json.dumps(data).encode("utf-8"))
+        f.seek(0)
```
istankovic, line 38: This `seek` is unnecessary.

beamer/agent/relayer.py
```diff
@@ -23,11 +29,21 @@ def get_relayer_executable() -> Path:
     return path.resolve()
 
 
+@contextlib.contextmanager
+def account_store(account: LocalAccount) -> Generator[tuple[str, str], None, None]:
+    password = uuid.uuid4().hex
+    data = account.encrypt(password)
+    with tempfile.TemporaryFile() as f:
+        f.write(json.dumps(data).encode("utf-8"))
```
istankovic, line 37: You can directly call 
```python
json.dump(data, f)
```

docker/scripts/common.sh
```diff
@@ -102,23 +102,25 @@ e2e_test_relayer() {
...
                                          --network-to $network_config \
                                          --network-from $network_config \
-                                         --wallet-private-key $privkey \
+                                         --keystore-file $keystore_path \
+                                         --password '' \
                                          --l2-transaction-hash $txhash; \
                         do sleep 1s; done"
 }
 
 e2e_test_op_proof(){
```
istankovic, line 111: Nitpick: while we're here, can you please insert one space between `()` and `{` to make it consistent with the rest of the functions?

beamer/tests/agent/unit/test_relayer.py
```diff
@@ -0,0 +1,47 @@
...
+
+from beamer.agent.relayer import run_relayer_for_tx
+from beamer.tests.agent.unit.utils import make_address, make_context
+
+
+@patch("beamer.agent.relayer.subprocess.run")
+@patch("beamer.agent.relayer.get_relayer_executable")
+def test_redacted_private_key_log(mocked_relayer_executable, mocked_subprocess):
+    context, _ = make_context()
+    fill_tx = HexBytes(make_address())
```
istankovic, line 15: Why do we need these? We can just  provide dummy values for both the RPC URL and the fill tx, right?
Since the relayer never gets to be executed, those don't need to be valid anyway.

beamer/tests/agent/unit/test_relayer.py
```diff
@@ -0,0 +1,47 @@
...
+from unittest.mock import patch
+
+from hexbytes import HexBytes
+
+from beamer.agent.relayer import run_relayer_for_tx
+from beamer.tests.agent.unit.utils import make_address, make_context
+
+
+@patch("beamer.agent.relayer.subprocess.run")
+@patch("beamer.agent.relayer.get_relayer_executable")
```
istankovic, line 12: I don't like the fact that we use this much mocking. We need to try to avoid messing with the implementation so much.

My thinking is, we should be able to get the same effect (check that the password was redacted) by e.g. providing a path to a nonexistent keyfile, or giving a wrong password etc. In both cases the relayer should exit with an error and we should be able to get the `CalledProcessError` and check for the password. And we can avoid patching our functions.

istankovic (changes requested):
Overall the Python part looks good, but the test needs more work (please see the comment).

beamer/tests/agent/unit/test_relayer.py
bilbeyt: in theory we dont need but i wanted to have a little bit more realistic values but ofc can change

compojoom:
the js part from my perspective looks ok. I have one remark though:
const wallet = Wallet.fromEncryptedJsonSync(encryptedWalletJson, password);

Back in the day when I joined the trustlines team the app was loading the key from keystore json - this very expensive operation and was locking the UI thread for 30+ seconds on some devices. I know that here we run on a server and we can wait, but want to make sure that we are aware of this. 
https://docs.ethers.org/v5/api/signer/#Wallet-fromEncryptedJsonSync
```
This operation will operate synchronously which will lock up the user interface, possibly for a non-trivial duration. Most applications should use the asynchronous fromEncryptedJson instead.
```

Switching to creating the wallet from privatekey made a huge difference in the speed. Not sure if it matters for our use case, but just keep that in mind. Are the other python threads which are working with the other chains going to work?

bilbeyt:
> 

When we need to call relayer from agent, we will create a keystore file and random password then will pass this path and password to the relayer and then relayer will decrypt it. It will be done only once when relayer initialized, so we wouldnt have any bad effects considering this is not frontend directly.

compojoom:
What do you mean once? The relayer will have to decode the json file on every start. I don't have a benchmark, maybe on the server it is way faster than on crappy phones... I'm just saying that the relayer is going to become slower when called. That's all

bilbeyt:
> What do you mean once? The relayer will have to decode the json file on every start. I don't have a benchmark, maybe on the server it is way faster than on crappy phones... I'm just saying that the relayer is going to become slower when called. That's all

Yes, comparing doing this on fronted using on relayer with agent server is much faster, but yes it will be called every start

istankovic:
> What do you mean once? The relayer will have to decode the json file on every start. I don't have a benchmark, maybe on the server it is way faster than on crappy phones... I'm just saying that the relayer is going to become slower when called. That's all

Thanks for the note, @compojoom. I don't think this matter here though because the relayer will never be invoked by the frontend, only by the agent, and since it does not happen too often, the slower operation is quite acceptable.

beamer/tests/agent/unit/test_relayer.py
```diff
@@ -0,0 +1,25 @@
...
+
+import pytest
+from eth_account.account import Account
+from hexbytes import HexBytes
+
+from beamer.agent.relayer import run_relayer_for_tx
+from beamer.typing import URL
+
+
+@pytest.mark.usefixtures("setup_relayer_executable")
```
istankovic, line 11: Nice idea!

istankovic (approved):
Looks good, with one comment regarding a place where we should use `pytest.raises`.

.circleci/workflows.yml
```diff
@@ -254,6 +254,8 @@ jobs:
     executor: vm
     environment:
       ARBITRUM_FOLDER: "docker/arbitrum"
+      DOCKER_BUILDKIT: 0
+      COMPOSE_DOCKER_CLI_BUILD: 0
```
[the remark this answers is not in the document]
bilbeyt: Just googled and wanted to try out, and it worked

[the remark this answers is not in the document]
bilbeyt: https://stackoverflow.com/a/66695181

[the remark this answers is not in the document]
istankovic: But do we know _why_ this works? And how this interacts with CircleCI's DLC (Docker Layer Caching)?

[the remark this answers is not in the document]
bilbeyt: With BuildKit, you don't need to pull the remote images before building since it caches each build layer in your image registry. Then, when you build the image, each layer is downloaded as needed during the build.

So i guess it will disable it for arbitrum e2e test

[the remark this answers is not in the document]
bilbeyt: Because this is disabling buildkit simply

Issue #2110, opened 2023-08-02
Improve log output


compojoom:
When a tx fails to go through we output an error like this:
```
fillRequest failed Context exc TransactionFailed('unknown error') request_id RequestId('0x...') source_chain_id 10 target_chain_id 1101 
```

since we don't output the tx_hash, but only the request id, it is hard to know which transaction failed.

compojoom:
Also, right now we don't know when a fill request was submitted. I think that we should output a "fill request" start event right before submitting the request to the chain:
https://github.com/beamer-bridge/beamer/blob/main/beamer/agent/chain.py#L404

Right now through the logs I can see that a request ended up on chain 30mins after a "requestcreated" event was seen by the agent, but I don't know when the agent actually submitted the fill.

PR #2128, opened 2023-08-10
fix: add more logs for failed tx and fill request

bilbeyt:
Closes: #2110


beamer/agent/chain.py
```diff
@@ -402,6 +402,8 @@ def _fill_request_exclusive(request: Request, context: Context) -> None:
         )
         return
 
+    context.logger.info("fillRequest started", request_id=request.id)
```
istankovic, line 405: I think this should rather be `debug` log level.

istankovic (approved):
Looks good, with one comment.

PR #2130, opened 2023-08-10
health: change unfilled behavior

compojoom:
Right now we output "request expired" if the request expires with no fill. We are changing this behavior to output "requst unfilled" until the request expires. After it has expired it would trigger "request expired". Of course provided the request was no actually filled.


fredo (approved):
gud

PR #2131, opened 2023-08-10
use claim fill id during invalidation resolution

fredo:
The PR fixes the correct resolution during resolution by invalidation

### DIFF
--- a/contracts/contracts/RequestManager.sol
+++ b/contracts/contracts/RequestManager.sol
@@ -726,7 +726,7 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
         if (filler != address(0)) {
             // Claim resolution via 1)
             claimValid = filler == claimer && fillId == claim.fillId;
-        } else if (request.invalidFillIds[fillId]) {
+        } else if (request.invalidFillIds[claim.fillId]) {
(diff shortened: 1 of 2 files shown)


bilbeyt (approved):
LGTM

bilbeyt:
@fredo please fix lint

contracts/contracts/RequestManager.sol
```diff
@@ -726,7 +726,7 @@ contract RequestManager is Ownable, LpWhitelist, RestrictedCalls, Pausable {
         if (filler != address(0)) {
             // Claim resolution via 1)
             claimValid = filler == claimer && fillId == claim.fillId;
-        } else if (request.invalidFillIds[fillId]) {
+        } else if (request.invalidFillIds[claim.fillId]) {
```
istankovic, line 729: I would use this opportunity to remove this part above:
```
        bytes32 fillId = request.fillId;
```
because it is not needed, the value is only used once and we remove the potential for confusion.
Sadly, github does not let me comment on the line in question.

fredo: It will be more costly when removing the filler, but since it is the unhappy case (L1 resolution), I'm fine with it.

Issue #1912, opened 2023-06-07
Contracts: allow removing support for a chain


istankovic:
Currently we have this in our request manager:
```
function createRequest(                           
    uint256 targetChainId,                        
    address sourceTokenAddress,                   
    address targetTokenAddress,                   
    address targetAddress,                        
    uint256 amount,                               
    uint256 validityPeriod                        
) external whenNotPaused returns (bytes32) {      
    require(                                      
        chains[targetChainId].finalityPeriod != 0,
        "Target rollup not supported"             
    );                                            
```
and this:
```
function updateChain(                                                     
    uint256 chainId,                                                      
    uint256 finalityPeriod,                                               
    uint256 transferCost,                                                 
    uint256 targetWeightPPM                                               
) external onlyOwner {                                                    
    require(finalityPeriod > 0, "Finality period must be greater than 0");
```

This means we don't have an effective way to remove support for a chain.
We should fix this so that we allow setting the finality period of a chain to zero so that the check in `createRequest` becomes functional.

istankovic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-62442a69c5e6e60010e31cf3/issues/beamer-bridge/beamer/1912?planning-poker) with Zenhub @bilbeyt @compojoom @fredo

PR #2136, opened 2023-08-11
feat: allow removing chain

bilbeyt:
Closes: #1912

### DIFF
--- a/beamer/config/commands.py
+++ b/beamer/config/commands.py
@@ -189,8 +189,12 @@ def _generate_chain_updates(
     current_chains = current_config.request_manager.chains
     desired_chains = desired_config.request_manager.chains
 
-    # We don't support chain removals yet.
-    assert set(current_chains).issubset(desired_chains)
+    # First check for removed chains.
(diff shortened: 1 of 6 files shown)


istankovic (approved):

beamer/config/commands.py
```diff
@@ -189,8 +189,12 @@ def _generate_chain_updates(
     current_chains = current_config.request_manager.chains
     desired_chains = desired_config.request_manager.chains
 
-    # We don't support chain removals yet.
-    assert set(current_chains).issubset(desired_chains)
+    # First check for removed chains.
+    for chain_id in current_chains:
+        if chain_id not in desired_chains:
+            # The chain is disabled by settings its finality_period
+            # target_weight_ppm and transfer_cost.
```
istankovic, line 196: Nitpick: I would be a bit more specific here, e.g.
```
            # The chain is disabled by setting its finality_period to zero.
            # At the same time, set target_weight_ppm and transfer_cost to zero too.
```

fredo (approved):

Issue #2132, opened 2023-08-10
Remove unnecessary require in withdraw


fredo:
https://github.com/beamer-bridge/beamer/blob/d15fc9f22491ffd1779aa7ba0ebb7e554639c962/contracts/contracts/RequestManager.sol#L658-L661)

I think this require is pretty unnecessary and complicates claim resolution in some edge cases. Since we can do claim on behalf, I don't see actually the need to require that it must be an active participant in the claim/challenge. If the address is not a participant then it would end up in the following situations: 

* Claim not finished. It would revert anyway if the claim is not resolvable.
* Valid claim. Regardless of the participant, the payouts would go to the claimer
* invalid claim. This is the only situation where we specifically need the address. As there can be multiple challengers, the address is used to pay out the challenger stake of this very addres. So if the address wasn't an active participant, it would revert in [here](https://github.com/beamer-bridge/beamer/blob/d15fc9f22491ffd1779aa7ba0ebb7e554639c962/contracts/contracts/RequestManager.sol#L758)

This leads me to the thinking that the require isn't really necessary.

Furthermore, it complicates things. Imagine the following situation: 

There is a claim with no challenger, but an L1 resolution made this claim invalid. Per code, since there is no challenger, the stakes go to the protocol. But this means, there is also only one participant in this claim - the claimer.
It is kind of confusing that one would have to withdraw the claim on behalf of the (malicious) claimer, in order to resolve the claim to make the claimer lose.
If we'd remove the require, we could just call withdraw with any address (or even with the wrapper function which uses msg.sender as the participant) in order to make the claimer lose and resolve the claim.
And since I don't see a reason why the require should be there, I think it would be an optimization to simply remove it.

PR #2137, opened 2023-08-11
feat: remove require from withdraw

bilbeyt:
Close: #2132


beamer/tests/contracts/test_request_manager.py
```diff
@@ -1436,3 +1436,29 @@ def test_transfer_limit_requests(request_manager, token):
     make_request(request_manager, token, requester, requester, transfer_limit + 1)
 
     assert token.balanceOf(requester) == 0
+
+
+def test_withdraw_invalid_fill(
```
fredo, line 1441: I would suggest to change the name to something like `test_withdraw_invalid_fill_no_challenger`

fredo (approved):

Issue #1646, opened 2023-03-31
Add a script to automate creating fill invalidations 


istankovic:
Currently we have to manually send fill invalidations between supported chains, which is cumbersome and error-prone.
It would be good to have this automated so that we can be more efficient after making a new deployment.
To that end, write a script that

- loads the chain RPC URLs from a file, for all supported chains
- uses the provided keyfile and deployment dir info to invoke the `call_contract` script with the `invalidate-fill` command
- auto-generates meaningful fill IDs for later debugging
-  outputs tx hashes for all sent transactions

istankovic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-62442a69c5e6e60010e31cf3/issues/beamer-bridge/beamer/1646?planning-poker) with Zenhub @fredo @GabrielBuragev

fredo:
How do you address the issue with the token? It would be good to have a configurable token. I can imagine one of the following

* add a token option and define by symbol. The script could read the state files and resolve the symbol to the correct address (I already have code for that). 
* Add a separate token.json file to the artifacts dir, which in the future will represent our mintable test token. The script could expect a `token.json` which includes a mapping from chainid to token address. For now we could build a token.json for USDC but in the future when we have the infra it could be our mintable token.

istankovic:
> How do you address the issue which token to use? It would be good to have a configurable token. I can imagine one of the following
> 
>     * add a token option and define by symbol. The script could read the state files and resolve the symbol to the correct address (I already have code for that).
> 
>     * Add a separate token.json file to the artifacts dir, which in the future will represent our mintable test token. The script could expect a `token.json` which includes a mapping from chainid to token address. For now we could build a token.json for USDC but in the future when we have the infra it could be our mintable token.

[...]

I spent quite a while thinking about this. At first, I immediately liked the idea of just reusing the config state files, because... why not? They are already there, contain the necessary info, we can then just reference the token via its symbol etc.

But then after thinking some more about it, I came to dislike the fact that we're now coupling ourselves (or, rather, the challenge test command) to the configuration state and this would necessitate adding yet another option to the already big options set. (In fact I'm not sure how it would work without some convention of naming the configuration state files and then depending on that convention....)

The idea with a separate `token.json` file is somewhat better in that respect, however, I still don't like the fact that we
- have to maintain that file and
- have to add yet another options

IMO, we already have a number of different input and output files so we should really think thrice before adding another kind of files.

[...]

PR #2144, opened 2023-08-17
Beamer check: add initiate-l1-invalidations command

istankovic:
This PR is part of #1646, adding the `beamer check initiate-l1-invalidations` command.

### DIFF
--- a/beamer/check/commands.py
+++ b/beamer/check/commands.py
@@ -1,0 +1,267 @@
+import json
+import random
+import sys
+import time
+from collections import Counter
+from dataclasses import dataclass, field
+from datetime import datetime, timedelta
+from pathlib import Path
+from subprocess import CalledProcessError
+
+import apischema
+import click
+import structlog
+from apischema import schema
+from eth_utils import to_checksum_address
+from hexbytes import HexBytes
+from web3.constants import ADDRESS_ZERO
+from web3.contract import Contract, ContractConstructor
+from web3.contract.contract import ContractFunction
+from web3.types import TxReceipt
+
+import beamer.chains
+import beamer.contracts
+import beamer.util
+from beamer.contracts import ABIManager, obtain_contract
+from beamer.relayer import run_relayer_for_tx
+from beamer.typing import ChainId, FillId, RequestId, TokenAmount
+from beamer.util import ChainIdParam, create_request_id, make_web3
+
+log = structlog.get_logger(__name__)
+
+
+@dataclass
+class Invalidation:
+    proof_source: ChainId = field(metadata=schema(min=1))
+    proof_target: ChainId = field(metadata=schema(min=1))
+    request_id: RequestId
+    fill_id: FillId
+    txhash: HexBytes
+    finalization_timestamp: datetime | None
+
+
+def _transact(func: ContractConstructor | ContractFunction) -> TxReceipt:
+    try:
+        receipt = beamer.util.transact(func)
+    except beamer.util.TransactionFailed as exc:
+        log.error("Transaction failed", exc=exc)
+        sys.exit(1)
+
+    txhash = receipt.transactionHash.hex()  # type: ignore
+    log.info("Transaction sent", block=receipt.blockNumber, txhash=txhash)  # type: ignore
+    return receipt
+
+
+def _make_invalidation(
+    fill_manager: Contract, proof_source: ChainId, proof_target: ChainId
+) -> Invalidation:
+    # Mark check-related fill IDs so that we can distiguish them
+    # on chain from other invalidations, if the need arises.
+    fill_id = FillId(b"beamer check " + random.randbytes(19))
+    assert len(fill_id) == 32
+
+    # The nonce is only 96 bits.
+    nonce = random.randint(0, 2**96 - 1)
+
+    func = fill_manager.functions.invalidateFill(
+        proof_target, ADDRESS_ZERO, ADDRESS_ZERO, TokenAmount(0), nonce, fill_id
+    )
+    receipt = _transact(func)
+
+    request_id = create_request_id(
+        proof_target,
+        proof_source,
+        to_checksum_address(ADDRESS_ZERO),
+        to_checksum_address(ADDRESS_ZERO),
+        TokenAmount(0),
+        nonce,
+    )
+    return Invalidation(
+        proof_source=proof_source,
+        proof_target=proof_target,
+        request_id=request_id,
+        fill_id=fill_id,
+        txhash=receipt.transactionHash,  # type: ignore
+        finalization_timestamp=None,
+    )
+
+
+@click.group()
+def check() -> None:
+    pass
+
+
+@check.command("initiate-l1-invalidations")
+@click.option(
+    "--keystore-file",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, path_type=Path),
+    required=True,
+    metavar="PATH",
+    help="Path to the keystore file.",
+)
+@click.password_option(
+    "--password",
+    type=str,
+    default="",
+    prompt=False,
+    help="The password needed to unlock the keystore file.",
+)
+@click.option(
+    "--rpc-file",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, path_type=Path),
+    required=True,
+    help="Path to the RPC config file.",
+)
+@click.option(
+    "--artifacts-dir",
+    type=click.Path(exists=True, file_okay=False, dir_okay=True, path_type=Path),
+    required=True,
+    help="The directory containing deployment artifacts.",
+)
+@click.option(
+    "--abi-dir",
+    type=click.Path(exists=True, file_okay=False, dir_okay=True, path_type=Path),
+    required=True,
+    help="Path to the directory with contract ABIs.",
+)
+@click.option(
+    "--output",
+    type=click.Path(file_okay=True, dir_okay=False, path_type=Path),
+    required=True,
+    help="Path to store the invalidation info at, which can be later used for verification.",
+)
+@click.option(
+    "--count",
+    type=click.IntRange(1),
+    default=1,
+    show_default=True,
+    help="Number of invalidations to create, per (PROOF_SOURCE, PROOF_TARGET) pair.",
+)
+@click.argument("proof-source", type=ChainIdParam())
+@click.argument("proof-target", type=ChainIdParam(), nargs=-1, required=True)
+def initiate_l1_invalidations(
+    keystore_file: Path,
+    password: str,
+    rpc_file: Path,
+    artifacts_dir: Path,
+    abi_dir: Path,
+    output: Path,
+    count: int,
+    proof_source: ChainId,
+    proof_target: tuple[ChainId],
+) -> None:
+    """Initiate L1 invalidations from chain PROOF_SOURCE to each PROOF_TARGET.
+    PROOF_SOURCE and PROOF_TARGET are chain IDs."""
+    beamer.util.setup_logging(log_level="DEBUG", log_json=False)
+
+    account = beamer.util.account_from_keyfile(keystore_file, password)
+    log.info("Loaded keystore file", address=account.address)
+
+    abi_manager = ABIManager(abi_dir)
+    deployment = beamer.artifacts.load(artifacts_dir, proof_source)
+
+    rpc_info = beamer.util.load_rpc_info(rpc_file)
+    url = rpc_info[proof_source]
+    w3 = make_web3(url, account)
+    assert w3.eth.chain_id == proof_source
+    log.info("Connected to RPC", url=url)
+
+    fill_manager = obtain_contract(w3, abi_manager, deployment, "FillManager")
+    request_manager = obtain_contract(w3, abi_manager, deployment, "RequestManager")
+
+    # Step 1: issue invalidateFill transactions.
+    # This will produce invalidations without finalization_timestamp.
+    #
+    # Note that if the output file already contains invalidations from
+    # proof_source to a specific proof_target, we will only issue  as many new
+    # invalidations as is necessary to reach `count`.
+    if output.exists():
+        with output.open("rt") as f:
+            data = json.load(f)
+        invalidations = apischema.deserialize(list[Invalidation], data)
+    else:
+        invalidations = []
+
+    existing_counts: Counter[ChainId] = Counter()
+    for invalidation in invalidations:
+        if invalidation.proof_source == proof_source:
+            existing_counts[invalidation.proof_target] += 1
+
+    sent = 0
+    for chain_id in proof_target:
+        for _ in range(existing_counts[chain_id], count):
+            invalidation = _make_invalidation(fill_manager, proof_source, chain_id)
+            invalidations.append(invalidation)
+            log.info(
+                "Sent invalidation",
+                idx=len(invalidations) - 1,
+                proof_source=proof_source,
+                proof_target=chain_id,
+            )
+            sent += 1
+
+            # Write the output immediately so that we have at least partial data
+            # in case an error occurs during invalidation for the next chain.
+            data = apischema.serialize(invalidations)
+            with output.open("wt") as f:
+                json.dump(data, f, indent=4)
+
+    # Step 2: complete invalidations with finalization_timestamp.
+    # This will prove invalidations if necessary (Optimism chains).
+    descriptor = beamer.chains.get_chain_descriptor(invalidation.proof_source)
+    assert descriptor is not None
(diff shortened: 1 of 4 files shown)


beamer/check/commands.py
```diff
@@ -0,0 +1,152 @@
...
+            fill_id=fill_id.hex(),
+            txhash=receipt.transactionHash.hex(),  # type: ignore
+            finalization_timestamp=timestamp.isoformat(timespec="seconds"),
+        )
+
+        # Write the output immediately so that we have at least partial data
+        # in case an error occurs during invalidation for the next chain.
+        with output.open("wt") as f:
+            json.dump(data, f, indent=4)
+        log.info("Invalidation info stored", path=str(output))
```
bilbeyt, line 152: Maybe we can at least add chain_id as log here like proof_source -> chain_id or something like that so we dont see duplicates.

beamer/tests/util.py
```diff
@@ -270,9 +270,9 @@ class CommandFailed(Exception):
     pass
 
 
-def run_command(*args):
+def run_command(cmd, *args):
     runner = CliRunner()
-    result = runner.invoke(*args)
+    result = runner.invoke(cmd, args, catch_exceptions=False)
```
bilbeyt, line 275: why catch_exceptions=False ?

istankovic:
Still missing; docs. :facepalm:

beamer/check/commands.py
```diff
@@ -0,0 +1,190 @@
...
+    txhash: HexBytes
+    finalization_timestamp: datetime
+
+
+def _transact(func: ContractConstructor | ContractFunction) -> TxReceipt:
+    try:
+        receipt = beamer.util.transact(func)
+    except beamer.util.TransactionFailed as exc:
+        log.error("Transaction failed", exc=exc)
+        sys.exit(1)
```
fredo, line 43: What's the motivation behind the strict exit here?

beamer/check/commands.py
```diff
@@ -0,0 +1,190 @@
...
+    for chain_id in proof_target:
+        for n in range(count):
+            invalidation = _make_invalidation(
+                fill_manager, proof_source, chain_id, finality_period
+            )
+            invalidations.append(invalidation)
+
+            # Write the output immediately so that we have at least partial data
+            # in case an error occurs during invalidation for the next chain.
+            data = apischema.serialize(invalidations)
```
fredo, line 181: Related to my comment above. I think we can do a bit better here. 

Imagine we have 5 chains and 3 invalidations each. This means we have 60 invalidations in total (If I'm not mistaken). Ideally, at the end we have a json file with 60 invalidation information, right? 

Imagine somewhere in the middle at invalidation 32, we exit due to TransactionFailed (we can only speculate, but my gut feeling tells me that with 60 transactions over 5 chains it's not unlikely that this happens once in a while). 

Now we have a partial invalidation file, which we cannot extend by resuming the command. It can get a bit tricky now. Of course, we could run a modified command with the chain combinations which haven't been executed yet, which requires some manual work. Then we can only write it into a new file and merge the two by hand. 
Lastly where the command broke we end up with a partial state, where we might have to again modify the command cause what we are gonna essentially do is running over two nested for loops where the outer chain at which the command broke will be partially done and partially not.

It's a good thinking to save the state at each run, but it would be good if the command is able to "resume" the work in case it crashed. Otherwise I see unnecessary manual work coming towards us.

One idea could be, similarly as with the state file, to read the output if the file exists and skip all invalidations which are already there and extend the file.

fredo: OTOH, if we want to make new invalidations we need to give a new output file (or remove the old one).

fredo:
All in all, looks pretty good. I'd like to be aligned on my comment above though. 

What's also missing (but part of the description in #1646) are the prove() calls for OP Stack chains.

beamer/tests/check/test_invalidations.py
```diff
@@ -0,0 +1,58 @@
...
+        "--abi-dir",
+        f"{root}/contracts/.build/",
+        "--artifacts-dir",
+        Path(artifact).parent,
+        "--output",
+        output,
+        "--count",
+        count,
+        str(chain_id),
+        str(chain_id),
```
fredo, line 42: Since you test the count variable here, you could also test that it works with multiple proof targets (I think the existence of the chain is not checked on a contract level during invalidation).

beamer/check/commands.py
istankovic: Two things:
- consistency with the other command(s)
- "fail early and complain loudly" so the user has a much higher chance of noticing that something went wrong

beamer/check/commands.py
istankovic: Heh, when thinking about this, I was like "well, cat is a good enough tool". :rofl: 
(In the sense, just like you mentioned above, you could stitch together multiple output file to have everything in one, but the downside is, of course, you would need to manually track where you left off when runnig additional commands.)

> Imagine we have 5 chains and 5 invalidations each. This means we have 20 invalidations in total (If I'm not mistaken). >Ideally, at the end we have a json file with 20 invalidation information, right?

The answer here depends on how you count.
If you count also the proof source chain into the 5 chains then you will get 20, yes. (4 proof target chains times 5 invalidations each).
If the 5 chains are all proof target chains, then you will get 25 invalidations in total. (5 proof target chains times 5 invalidations each).

> In any case, if our goal is to have a file containing all invalidations from one chain to all others, then we should have  the tools to create this file in case of a crash

Yeah, I can see the benefit of just being able to resume where you left off in case something failed. The data is already there. I will extend it so that if the output file exists already, we continue so that one does not need to adjust the command. :+1:

istankovic:
> What's also missing (but part of the description in #1646) are the prove() calls for OP Stack chains.

Yeah, I realized that just after marking this ready for review. :facepalm: 
Will add, of course. (I'm not sure how to even approach testing this part so I'm probably going to leave it out of this PR. Maybe something based on top of our existing OP E2E tests, but even that does not seem fully straightforward...)

beamer/tests/util.py
istankovic: Part of the previously merged batch, but I want to answer anyway: this is so that exceptions propagate to pytest/ape and are visible immediately when running tests, instead of us having to run PDB when `CommandFailed` is raised and inspecting `result.exc_info` manually.

beamer/check/commands.py
istankovic: We now have more info in the logs.

fredo:
> > What's also missing (but part of the description in #1646) are the prove() calls for OP Stack chains.
> 
> Yeah, I realized that just after marking this ready for review. 🤦 Will add, of course. (I'm not sure how to even approach testing this part so I'm probably going to leave it out of this PR. Maybe something based on top of our existing OP E2E tests, but even that does not seem fully straightforward...)

I would be fine to push this into another PR cause we need to think a bit more about the information being passed here. For OP chains the finality period only starts after the prove() call was made, so we cannot really add the timestamp if the prove() didnt happen. 

With regards to testing, in order to really test it we would need an actual OP setup like in the e2e tests...

beamer/tests/check/test_invalidations.py
istankovic: Ah, good point; added a new, dummy proof target.

beamer/check/commands.py
istankovic: Made it so the command reads the output file if it exists and continues.

istankovic:
Alright, the docs are now also ready. :scroll:

beamer/check/commands.py
```diff
@@ -0,0 +1,256 @@
...
+    if output.exists():
+        with output.open("rt") as f:
+            data = json.load(f)
+        invalidations = apischema.deserialize(list[Invalidation], data)
+    else:
+        invalidations = []
+
+    existing_counts: Counter[ChainId] = Counter()
+    for invalidation in invalidations:
+        if invalidation.proof_source == proof_source:
```
fredo, line 188: Question: 

Do you intent to batch multiple proof sources into one global end file or do you plan to leave the generated outputs (invalidation files), which are separated by proof_source.

If one or the other it would be ideal to either keep the if condition here, or to actually assert if the consumed output file is compatible.

beamer/check/commands.py
```diff
@@ -0,0 +1,256 @@
...
+
+    rpc_info = beamer.util.load_rpc_info(rpc_file)
+    url = rpc_info[proof_source]
+    w3 = make_web3(url, account)
+    assert w3.eth.chain_id == proof_source
+    log.info("Connected to RPC", url=url)
+
+    fill_manager = obtain_contract(w3, abi_manager, deployment, "FillManager")
+    request_manager = obtain_contract(w3, abi_manager, deployment, "RequestManager")
+    finality_period = request_manager.functions.chains(proof_source).call()[0]
```
fredo, line 171: nitpick: you can probably move this call down to line 212, right before the for loop where the timestamps will be set. 

This would safe us the call if no invalidations would have been made.

beamer/tests/check/test_invalidations.py
```diff
@@ -0,0 +1,63 @@
...
+    for invalidation, proof_target in zip(invalidations, targets):
+        assert invalidation.proof_source == chain_id
+        assert invalidation.proof_target == proof_target
+
+        receipt = w3.eth.get_transaction_receipt(invalidation.txhash)
+        fill_manager = ape.project.FillManager.at(receipt.to)
+        event = fill_manager.FillInvalidated.from_receipt(receipt)[0]
+        assert event.event_arguments == dict(
+            requestId=invalidation.request_id, fillId=invalidation.fill_id
+        )
```
fredo, line 63: Suggestion: A test which continues on an existing invalidation file would be nice. 

option a) run the command again with an increased count
option b) run the command again with one more dummy chain

docs/source/commands.rst
```diff
@@ -161,3 +162,48 @@ command will notify the user by sending everything either to Telegram or RocketC
      - Logging level, one of ``debug``, ``info``, ``warning``, ``error``, ``critical``.
        Default: ``error``.
 
+
+.. _command-check-initiate-l1-invalidations:
+
+``beamer check initiate-l1-invalidations``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``beamer check initiate-l1-invalidations read [OPTIONS] PROOF_SOURCE PROOF_TARGET...``
```
fredo, line 171: typo? `... initiate-l1-invalidations read...`

docs/source/commands.rst
```diff
@@ -161,3 +162,48 @@ command will notify the user by sending everything either to Telegram or RocketC
...
+
+The command initiates L1 invalidations from ``PROOF_SOURCE`` to each
+``PROOF_TARGET``. ``PROOF_SOURCE`` and ``PROOF_TARGET`` are chain IDs. By
```
fredo, line 174: Could you extend with a sentence that the last N arguments (but at least 1) are PROOF_TARGETS. At least it is not clearly understandable that you can add many proof targets.

beamer/check/commands.py
istankovic: > Do you intent to batch multiple proof sources into one global end file or do you plan to leave the generated outputs (invalidation files), which are separated by proof_source.

The intent was/is to allow for invalidations with different proof sources in the same file, i.e. I would not prohibit having different proof sources in the same file. This lets the user decide the policy with respect to how those files are organized, named etc. Most likely, for the toplevel script, we'd just have 1 file per proof_source to all other chains, but I wanted to leave that up to the script, that is, the user of the command.

> If one or the other it would be ideal to either keep the if condition here,

Yeah, I'd like for this condition to stay there.

Your question made me realize that for a differing proof_source we will do nothing in step 1, while in step 2 we may actually run the relayer, regardless of both proof_source and proof_target. So I'm thinking step 2 about skipping step 2 too if proof_source of the invalidation does not match and proof_target of the invalidation is not in the provided list. What's your opinion on this?

docs/source/commands.rst
istankovic: Definitely, thanks! :eyes:

docs/source/commands.rst
istankovic: Eh, this is just using the `X...` syntax with triple dots to describe arbitrary number of args (at least one), which is pretty standard, see e.g. `cat` manpage:
```
NAME
       cat - concatenate files and print on the standard output

SYNOPSIS
       cat [OPTION]... [FILE]...
```

But I will add some wording to make it more explicit.

beamer/check/commands.py
fredo: > Your question made me realize that for a differing proof_source we will do nothing in step 1, while in step 2 we may actually run the relayer, regardless of both proof_source and proof_target. So I'm thinking step 2 about skipping step 2 too if proof_source of the invalidation does not match and proof_target of the invalidation is not in the provided list. What's your opinion on this?

That's a good point.. I see two arguments for the two sides. If we automatically do prove calls, we always require all RPCs in the rpc file, even though the command was about different chains. This could be a problem of not being able to successfully finish a command cause of an unrelated missing rpc.
On the other hand, it's easy to miss finalizations. 

My suggestion would be to clearly indicate that invalidations for proof_source X have been completed, to give at least some feedback to the user in this direction, but only handle those invalidations which are part of the command

docs/source/commands.rst
fredo: Yes, I see it in the command, of course. But I was specifically referring to the wording.

beamer/tests/check/test_invalidations.py
istankovic: Done. Thanks for making me do the tests.

We now have nice coverage:
image

(The uncovered part is mostly OP-prove-related code.)

beamer/check/commands.py
istankovic: > My suggestion would be to clearly indicate that invalidations for proof_source X have been completed, to give at least some feedback to the user in this direction, but only handle those invalidations which are part of the command

Yeah, that sounds good. Will add something to the logs and adjust the current behaviour to only consider specified proof source. :+1:

istankovic: Here's an example of (1) running with all invalidations already sent, and (2) running with 1 invalidation sent, the rest already done before (makes for 4 in total).

```
beamer-py3.10) [ivan@beta4 beamer]$ beamer check initiate-l1-invalidations --keystore-file 0x1CEE82EEd89Bd5Be5bf2507a92a755dcF1D8e8dc.json --password '' --rpc-file .local.rpc.json --artifacts-dir deployments/artifacts/test/ --abi-dir contracts/.build/ -
output /tmp/foo.invalidations 1337 1337 --count 3 
023-08-21 14:04:59.899602 [info     ] Loaded keystore file           [beamer.check.commands] address=0x1CEE82EEd89Bd5Be5bf2507a92a755dcF1D8e8dc
023-08-21 14:04:59.966677 [info     ] Connected to RPC               [beamer.check.commands] url=http://127.0.0.1:8545
023-08-21 14:05:00.008475 [info     ] Invalidations initiated successfully [beamer.check.commands] path=/tmp/foo.invalidations sent=0 total=3
beamer-py3.10) [ivan@beta4 beamer]$ beamer check initiate-l1-invalidations --keystore-file 0x1CEE82EEd89Bd5Be5bf2507a92a755dcF1D8e8dc.json --password '' --rpc-file .local.rpc.json --artifacts-dir deployments/artifacts/test/ --abi-dir contracts/.build/ -
output /tmp/foo.invalidations 1337 1337 --count 4 
023-08-21 14:05:14.263204 [info     ] Loaded keystore file           [beamer.check.commands] address=0x1CEE82EEd89Bd5Be5bf2507a92a755dcF1D8e8dc
023-08-21 14:05:14.319749 [info     ] Connected to RPC               [beamer.check.commands] url=http://127.0.0.1:8545
023-08-21 14:05:14.422818 [info     ] Transaction sent               [beamer.check.commands] block=85 txhash=0x040cca501a506d62aa6922ddae1bd0e8901a724b6fb165463916c08504a77145
023-08-21 14:05:14.423418 [info     ] Sent invalidation              [beamer.check.commands] idx=3 proof_source=1337 proof_target=1337
023-08-21 14:05:14.441107 [info     ] Invalidation complete          [beamer.check.commands] finalization_timestamp=2023-08-21T14:05:14 idx=3 proof_source=1337 proof_target=1337
023-08-21 14:05:14.441248 [info     ] Invalidations initiated successfully [beamer.check.commands] path=/tmp/foo.invalidations sent=1 total=4
```

fredo: looks good!

fredo (approved):
very nice 🍷

PR #2149, opened 2023-08-17
beamer: config: remove chain from dictionary if reset to default values

fredo:
The same thing was missing for chains. 

Removing chains should result in removing it from the dicitonary.

### DIFF
--- a/beamer/tests/config/test_read.py
+++ b/beamer/tests/config/test_read.py
@@ -51,15 +51,34 @@ def test_config_read_request_manager(tmp_path, token, deployer):
     assert token_config.eth_in_token == 5
     assert config.token_addresses == {symbol: token.address}
 
+    # check token removal
+    request_manager.updateToken(token.address, 0, 0)
+    config = read_config_state(rpc_file, artifact)
+
+    assert config.block > old_block
+    assert symbol not in config.token_addresses
+    assert len(config.token_addresses) == 0
+    assert symbol not in config.request_manager.tokens
+    assert len(config.request_manager.tokens) == 0
+
     # check chain update
-    request_manager.updateChain(123, 6, 7, 8)
+    chain_id = 123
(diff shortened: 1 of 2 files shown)


bilbeyt (approved):
Seems good to me, checked if we have any tests regarding Token removal and couldnt find any so this should be also ok

istankovic:
> Seems good to me, checked if we have any tests regarding Token removal and couldnt find any so this should be also ok

Should we not then use this opportunity to add new tests, both for token and chain removal? :smile: 
Especially since it seems like an easy thing to do.

fredo:
> > Seems good to me, checked if we have any tests regarding Token removal and couldnt find any so this should be also ok
> 
> Should we not then use this opportunity to add new tests, both for token and chain removal? 😄 Especially since it seems like an easy thing to do.
> 
> It would also be interesting to see test coverage now for modules under `beamer.config`.

There you go

beamer/tests/config/test_read.py
```diff
@@ -61,6 +71,14 @@ def test_config_read_request_manager(tmp_path, token, deployer):
     assert chain_config.transfer_cost == 7
     assert chain_config.target_weight_ppm == 8
 
+    # check chain removal
+    request_manager.updateChain(123, 0, 0, 0)
+    config = read_config_state(rpc_file, artifact)
+
+    assert config.block > old_block
+    assert symbol not in config.request_manager.chains
+    assert len(config.request_manager.tokens) == 0
```
istankovic, line 80: This should have been
```suggestion
    assert len(config.request_manager.chains) == 0
```
right?

istankovic (approved):

beamer/tests/config/test_read.py
fredo: it was late last night... pushed a fix

PR #2151, opened 2023-08-17
Remove config parameters from deployment config

fredo:
fees, and chain parameters which are changable are removed from deployment config as they are not needed there. They are configure by beamer config write commands along with desired state files


istankovic (approved):
:parrot:

PR #2174, opened 2023-08-24
contracts: use safeApprove for FeeSub

bilbeyt:


fredo (approved):

fredo:
safeApprove

contracts/contracts/FeeSub.sol
```diff
@@ -130,12 +130,12 @@ contract FeeSub is Ownable {
         uint256 amount
     ) external onlyOwner {
         if (amount > 0 && minimumAmounts[tokenAddress] == 0) {
-            IERC20(tokenAddress).approve(
+            IERC20(tokenAddress).safeApprove(
```
istankovic, line 133: Why are we using `safeApprove` here?

To quote https://github.com/code-423n4/2022-05-aura-findings/issues/35
```
OpenZeppelin's safeApprove implementation is deprecated. Using this deprecated function can lead to unintended reverts and potentially the locking of funds.
```

The doc comment from OpenZeppelin:
```
    /**
     * @dev Deprecated. This function has issues similar to the ones found in
     * {IERC20-approve}, and its usage is discouraged.
     *
     * Whenever possible, use {safeIncreaseAllowance} and
     * {safeDecreaseAllowance} instead.
     */
```

fredo: The reason for the change was USDT not implementing the ERC20 standard which leads to not being able to use ERC20.approve(). `safeApprove` handles it, but if there is an alternative and safeApprove is deprecated, then we might want to use that.

PR #2177, opened 2023-08-24
frontend: use v3 contracts

GabrielBuragev:


fredo:
Yes let's remove the parenthesis in the banner 
@GabrielBuragev

PR #2178, opened 2023-08-24
Beamer check verify invalidations

istankovic:
Part of #1646, this introduces a command to verify L1 invalidations.

Coverage:
image

### DIFF
--- a/beamer/check/commands.py
+++ b/beamer/check/commands.py
@@ -2,15 +2,18 @@ import json
 import random
 import sys
 import time
-from collections import Counter
+from collections import Counter, defaultdict
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
+from enum import Enum
 from pathlib import Path
+from typing import NamedTuple
 
 import apischema
 import click
 import structlog
 from apischema import schema
+from eth_account.account import LocalAccount
 from eth_utils import to_checksum_address
 from hexbytes import HexBytes
 from web3.constants import ADDRESS_ZERO
@@ -23,7 +26,7 @@ import beamer.contracts
 import beamer.util
 from beamer.contracts import ABIManager, obtain_contract
 from beamer.relayer import RelayerError, run_relayer_for_tx
-from beamer.typing import ChainId, FillId, RequestId, TokenAmount
+from beamer.typing import URL, ChainId, FillId, RequestId, TokenAmount
 from beamer.util import ChainIdParam, create_request_id, make_web3
 
 log = structlog.get_logger(__name__)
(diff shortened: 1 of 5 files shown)


beamer/check/commands.py
```diff
@@ -264,3 +264,166 @@ def initiate_l1_invalidations(
...
+        except RelayerError as exc:
+            log.error(
+                "Relayer failed", idx=idx, exc=exc, proof_source=pair[0], proof_target=pair[1]
+            )
+            errors += 1
+        else:
+            log.info("Relayer succeeded", thxash=invalidation.txhash)
+
+            # Check the request manager, the invalidation should have arrived.
+            if request_manager.functions.isInvalidFill(
```
fredo, line 397: This will only work after the relayer is checking the request manager, right?

beamer/check/commands.py
```diff
@@ -264,3 +264,166 @@ def initiate_l1_invalidations(
...
+        try:
+            run_relayer_for_tx(
+                l1_rpc=rpc_info[deployment.base.chain_id],
+                l2_relay_from_rpc_url=rpc_info[invalidation.proof_source],
+                l2_relay_to_rpc_url=rpc_info[invalidation.proof_target],
+                account=account,
+                tx_hash=invalidation.txhash,
+            )
+        except RelayerError as exc:
+            log.error(
```
fredo, line 389: I noticed an edge case here: 

In line 367 you check if the invalidation is already resolved, and if so, the respective chain pair is gonna be marked as verified. Thus any further invalidations will not be tried to be resolved, which is fine for now.

However, there is a situation, where the first invalidation for the chain pair fails on the relayer, and if I'm not mistaken, the loop will continue and try the next one. This means that the second invalidation is the one which verified the chain pair. 
If you then run the command again, it would try to resolve the first one again even though the chain pair is already verified by the second invalidation. 
It's just a nitpick.

fredo: Which leads me to the thinking, what is the intended algorithm for this command? Especially if you run the command more often (either due to a crash or because the file contains multiple chain pairs, with different finalization timestamps). 

I understood the code in a way, that it would always check if it was already verified and if so then don't do anything for this chain pair. Is that correct? 

Then I see the following situations, which are really just nitpicks IMO:

1) The situation above where not the first invalidation for a chain pair resolved correctly but a later one did. Then the code will always invalidate the first one again

2) We cannot really do another invalidation with this command if we wanted to. (probably deleting the resolved invalidation from the file would help). However, I don't have a real scenario at hand where we would want to do this, but for sure for some reason we always do multiple invalidations in order to safe another cycle of finality periods.
A tradeoff could be to introduce a flag, which will not check for existing invalidations and thus requires one invalidation for each chain pair again.

beamer/check/commands.py
```diff
@@ -264,3 +264,166 @@ def initiate_l1_invalidations(
...
+                    proof_target=pair[1],
+                )
+                errors += 1
+
+    if errors:
+        log.error(
+            "Error(s) occurred during verification",
+            errors=errors,
+            verified=len(verified),
+            chain_pairs=chain_pairs,
```
fredo, line 421: is this really true? I think it would confuse me, because chain_pairs contains all chain_pairs which have been touched regardless of error or success

beamer/check/commands.py
istankovic: > I understood the code in a way, that it would always check if it was already verified and if so then don't do anything for this chain pair. Is that correct?

Yes. Which is why I would not consider your comments here nitpics -- it's a bug that needs to be fixed.
We need to 
- group all invalidations by (source, target) chain pair
- drop the whole list for a particular pair if any were had been verified
- verify invalidations of remaining chain pairs

istankovic: To provide more info on how I am thinking about the command:

> 1. The situation above where not the first invalidation for a chain pair resolved correctly but a later one did. Then the code will always invalidate the first one again

This should not happen. Once an invalidation is done, we should not waste time/money for checking the same chain pair again.

> 2\. We cannot really do another invalidation with this command if we wanted to. (probably deleting the resolved invalidation from the file would help). However, I don't have a real scenario at hand where we would want to do this, but for sure for some reason we always do multiple invalidations in order to safe another cycle of finality periods.
>         A tradeoff could be to introduce a flag, which will not check for existing invalidations and thus requires one invalidation for each chain pair again.

That, or a flag to remove the ones which were already verified (perhaps saving the original in a backup file).

beamer/check/commands.py
istankovic: Yes, that assumes the relayer change is done, as discussed with Gabriel.

beamer/check/commands.py
istankovic: Yeah, that's confusing, I wanted to give info like "verified M out of N chain pairs in total", something like that....
Maybe something like this?
```python
            total_chain_pairs=len(chain_pairs)
```

fredo: That would be better, But I probably would also like to see the failing chain pairs, for further debugging

istankovic: Addressed this in the new version, by showing all unverified chain pairs. The logs will also contain specific errors that occurred.

docs/source/commands.rst
```diff
@@ -208,3 +209,40 @@ pair is equal or greater to the count, no new invalidations will be sent.
...
+
+``beamer check verify-l1-invalidations``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``beamer check verify-l1-invalidations [OPTIONS] FILE``
+
+The command verifies L1 invalidations contained in ``FILE``. In case of
+multiple invalidations for the same pair ``(proof_source, proof_target)``, if
+an invalidation is verified successfully, remaining invalidations for the same
+chain pair will be skipped as they then not considered necessary. If a
```
fredo, line 224: missing `are` ? `as they then not considered necessary`

fredo (approved):
very nice!

docs/source/commands.rst
istankovic: > missing `are` ? `as they then not considered necessary`

Nice catch! :eyes: 
Fixed, plus I rewored the last sentence to be more clear:
```diff
diff --git a/docs/source/commands.rst b/docs/source/commands.rst
index cdace38c..a5f2955c 100644
--- a/docs/source/commands.rst
+++ b/docs/source/commands.rst
@@ -221,10 +221,11 @@ pair is equal or greater to the count, no new invalidations will be sent.
 The command verifies L1 invalidations contained in ``FILE``. In case of
 multiple invalidations for the same pair ``(proof_source, proof_target)``, if
 an invalidation is verified successfully, remaining invalidations for the same
-chain pair will be skipped as they then not considered necessary. If a
+chain pair will be skipped as they are then not considered necessary. If a
 verification fails for any invalidation, for whatever reason, an error will be
 emitted and the process will continue with the next invalidation. The command
-will exit with a success code only if there were no errors.
+will exit with a success code only if all chain pairs had at least one
+successful invalidation.
```

istankovic:
Found also another typo in a comment:
```diff
--- a/beamer/check/commands.py                                                 
+++ b/beamer/check/commands.py                                                 
@@ -470,8 +470,8 @@ def verify_l1_invalidations(
     # If pair is present in both failed and verified, that's considered fine
     # since different invalidations for the same pair can have sufficiently
     # different finalization timestamps such that one is finalized and the
-    # other is. As long as all chain pairs in failed are also in verified, we 
-    # should report success.                                                  
+    # other is not. As long as all chain pairs in failed are also in verified,
+    # we should report success.                                               
```

PR #2189, opened 2023-08-31
Beamer check initiate-challenges

istankovic:
This PR adds a new command, `beamer check initiate-challenges` in order to automate challenge tests.

Part of #1646.


beamer/check/commands.py
```diff
@@ -478,3 +478,246 @@ def verify_l1_invalidations(
         sys.exit(1)
 
     log.info("Invalidations for all chain pairs successful", verified=len(verified))
+
+
+@dataclass
+class Challenge:
+    request_chain: ChainId = field(metadata=schema(min=1))
```
bilbeyt, line 485: as it is ChainId maybe we should name this like request_chain_id, same true for next line for fill_chain_iid

beamer/check/commands.py
```diff
@@ -478,3 +478,246 @@ def verify_l1_invalidations(
...
+    fill_chain: ChainId = field(metadata=schema(min=1))
+    request_id: RequestId
+    claim_id: ClaimId | None
+    txhash_createRequest: HexBytes
```
bilbeyt, line 489: shouldnt we change this like create_request_tx_hash ?

beamer/check/commands.py
```diff
@@ -478,3 +478,246 @@ def verify_l1_invalidations(
...
+@click.option(
+    "symbol",
+    "--token",
+    type=str,
+    required=True,
+    help="Symbol of the token to be used for challenges (e.g. USDC).",
+)
+@click.option(
+    "--stake",
+    type=click.FloatRange(0.125),
```
bilbeyt, line 649: 0.1 is enough

beamer/check/commands.py
```diff
@@ -478,3 +478,246 @@ def verify_l1_invalidations(
...
+        chain_id=challenge.request_chain,
+        claim_id=challenge.claim_id,
+        txhash=txhash,
+    )
+
+    block = request_manager.w3.eth.get_block(receipt["blockNumber"])
+    finality_period = request_manager.functions.chains(challenge.fill_chain).call()[0]
+    timestamp = datetime.utcfromtimestamp(block["timestamp"])
+    challenge.txhash_challengeClaim = HexBytes(txhash)
+    challenge.finalization_timestamp = timestamp + timedelta(seconds=finality_period)
```
bilbeyt, line 598: i think there is a problem here related with OP chains

bilbeyt (commented):
Generally looking good with some questions

beamer/check/commands.py
istankovic: Eh, I was doing following the same pattern as in the L1 invalidation-related commands... it's a bit shorter and perhaps a tiny bit more readable, but OTOH it does differ from the usual pattern in e.g. `beamer.agent`.
I'd prefer to leave it as is, if for nothing else, then for for the similarity with the existing check commands.

beamer/check/commands.py
istankovic: > i think there is a problem here related with OP chains

Could you please be more specific?

beamer/check/commands.py
istankovic: Changed.

beamer/check/commands.py
istankovic: After a lengthy discussion, we decided to go with the current finalization timestamp computation. Ideally, we would start the finalization period from the moment the proof is executed on L1 (this is done by the agent internally), however, it is not straightforward to get the proof tx timestamp and possible solutions (e.g. watching for events on L1, querying the related OP contracts) would add much complexity.

beamer/check/commands.py
```diff
@@ -478,3 +478,246 @@ def verify_l1_invalidations(
...
+    fill_chain: ChainId = field(metadata=schema(min=1))
+    request_id: RequestId
+    claim_id: ClaimId | None
+    create_request_txhash: HexBytes
+    claim_request_txhash: HexBytes | None
+    challenge_claim_txhash: HexBytes | None
+    finalization_timestamp: datetime | None
+
+
+def _find_token(ctx: Context, chain_id: ChainId, symbol: str) -> Contract | None:
```
fredo, line 495: I think the solution to find the token via the symbol is suboptimal. Especially with the discussion in mind we had during the config commands, that there can easily by more tokens with the same symbol.
Secondly, fetching events from deployment on can become quite costly (It may even crash below, if the RPC restricts a large range of blocks to fetch logs from).

However, I currently fail to provide an alternative. Currently our test token is integrated into the deployment artifacts. We could make it a requirement. Or we provide an alternative list of the test token for the purpose of testing this command. 
I see that running the test agent is out of scope here, so we need to know of the token anyways to fund the agent. 
(I think what we had aligned on was that we will run these tests with a specific test token)

istankovic: >  Secondly, fetching events from deployment on can become quite costly (It may even crash below, if the RPC restricts a large range of blocks to fetch logs from).

That is a simple matter of policy: we need to just make sure the test token is deployed soon after the Beamer contracts. Finding the token then should be neither costly nor cause the RPC to restrict us.

>  I see that running the test agent is out of scope here, so we need to know of the token anyways to fund the agent.

Yup.

>  (I think what we had aligned on was that we will run these tests with a specific test token)

Yes, we discussed all of the above already in #1646.

fredo: > That is a simple matter of policy: we need to just make sure the test token is deployed soon after the Beamer contracts. Finding the token then should be neither costly nor cause the RPC to restrict us.

I'm not sure I understand this comment. Calling `get_Logs(from=RequestManagerDeploymentBlock)` without specifying the `to` parameter will result in a call fetching events from X to `latest`. This is regardless of when the test token was deployed. (In any case the deployment of the token doesn't matter, I think you are referring to the first time when the token was updated on the Request Manager).
However, this call still grows with time so the calls we will be doing will always be from deployment of the request manager to `latest`, which as a result can crash if the provider doesn't allow fetching logs > X blocks where `latest - deployment > X`.

fredo: Maybe my problem with using the symbol is that we are resolving by the first ever seen event and not by the latest, which would be more accurate IMO.

istankovic: > > That is a simple matter of policy: we need to just make sure the test token is deployed soon after the Beamer contracts. Finding the token then should be neither costly nor cause the RPC to restrict us.
> 
> I'm not sure I understand this comment. Calling `get_Logs(from=RequestManagerDeploymentBlock)` without specifying the `to` parameter will result in a call fetching events from X to `latest`. This is regardless of when the test token was deployed. (In any case the deployment of the token doesn't matter, I think you are referring to the first time when the token was updated on the Request Manager). However, this call still grows with time so the calls we will be doing will always be from deployment of the request manager to `latest`, which as a result can crash if the provider doesn't allow fetching logs > X blocks where `latest - deployment > X`.

Hmm, I see your point.

```python
    logs = request_manager.events.TokenUpdated.get_logs(fromBlock=from_block)  # type: ignore
```
will only look for `TokenUpdated` events, but the fact is, `get_logs` is still an eager call and it will search until the latest block, as you describe. Ugh. :woozy_face: 

Would have been nice to have a streaming/iterator-style call, but oh well.
I'm not really sure what to do here; a few options come to mind:
# specify "good enough" `to` block number
# use our `EventFetcher`

istankovic: After some discussion, we decided not to use the `EventFetcher` as that would require relatively risky changes to it. Instead, we're going with the assumption that the first token with the symbol we look for is the correct one.

fredo (approved):
thanks for the research on the alternative ❤️

PR #2198, opened 2023-09-06
Add relayer documentation

### LINKED ISSUE #1301: Relayer: Add documentation
# Rationale

Since the relayer grew in size and therefore in complexity we need to properly document the setup/usage of it.

GabrielBuragev:
Closes #1301

### DIFF
--- a/docs/source/relayer.rst
+++ b/docs/source/relayer.rst
@@ -1,0 +1,107 @@
+Cross-chain message relayer
+===========================
+
+This component is responsible for handling the cross-chain contract state updates. 
+
+It is mainly used by the Agent software when resolving requests as a result of a running challenge game 
+or for sending fill invalidation messages (L1 resolutions). However, it can also be used manually by running
+the CLI commands.
+If you haven't read about the role of :ref:`l1-resolutions` in Beamer, make sure you get familiar 
+with this concept first.
+
+Every **request fill** action creates a **proof** on the request destination chain.
+Since these proofs are created on a different chain than the chain where the requests originated from,
+we need a way to transport these proofs from one chain to another in order to be able 
+to prove that a certain state is correct. This is where the cross-chain relayer comes into play.
+
+The agent utilizes these proofs if and when needed. 
+In normal operation, agent's claims will not be challenged so the fill proofs are not needed, therefore the agent will not invoke the relayer.
+However, if a challenge game occurs, the agent has to make use of the proof to ensure it wins the game. 
+Typically, that is done by invoking the relayer at an appropriate moment, i.e. when the proof becomes available for execution on Ethereum L1.
+
+For instructions on how to setup, compile and run the relayer, make sure you take a look at the `/relayer/README.md` file.
+
+Command reference
+-----------------
+
+The relayer CLI currently supports the following commands:
+
+
+* :ref:`command-relay` relays a fill proof or a fill invalidation message from chain A to chain B.
+* | :ref:`command-prove-op-message` proves on L1 that a message exists on L2. 
+  | Used only for messages traveling from chains that are based on the *Optimism Bedrock* stack.
+
+.. _command-relay:
+
+``relay``
+^^^^^^^^^
+
+The ``relay`` command takes care of relaying a message from one chain to another (end-to-end).
+In cases where a message needs to be relayed from an "Optimism Bedrock"-like chain, then one first needs to run
+the :ref:`command-prove-op-message` command, for the same message that needs to be relayed, before one can run the :ref:`command-relay` command.
+
+
+.. list-table::
+   :header-rows: 1
+
+   * - Command-line option 
+     - Description
+
+   * - ``--l1-rpc-url URL``
+     - RPC URL to be used for communicating with the base chain.
+
+   * - ``--l2-relay-to-rpc-url URL``
+     - RPC URL to be used for communicating with the destination chain (chain where the proof is travelling *to*).
+
+   * - ``--l2-relay-from-rpc-url URL``
+     - RPC URL to be used for communicating with the source chain (chain where the proof is travelling *from*).
+
+   * - ``--keystore-file PATH``
+     - Path to the keystore file.
+
+   * - ``--password PASSWORD``
+     - Password of the keystore file.
+
+   * - ``--l2-transaction-hash TX_HASH``
+     - Transaction hash of the submitted message that needs to be relayed.
+
+   * - ``--network-from PATH``
+     - Path to a file with custom network configuration. This option is mainly used for development purposes.
+
+   * - ``--network-to PATH``
+     - Path to a file with custom network configuration. This option is mainly used for development purposes.
+
+
+.. _command-prove-op-message:
+
+``prove-op-message``
+^^^^^^^^^^^^^^^^^^^^
+The ``prove-op-message`` command is used only when a message needs to travel *from* "Optimism Bedrock"-like networks.
+On such networks, there is a need of proving the message up front before it is treated as a message
+that should be included in a batch and submitted to L1 (after the proof is finalized on L2).
+The finality period starts only after the message was proven on L2.
+
+
+.. list-table::
+   :header-rows: 1
+
+   * - Command-line option 
+     - Description
+
+   * - ``--l1-rpc-url URL``
+     - RPC URL to be used for communicating with the base chain.
+
+   * - ``--l2-rpc-url URL``
+     - RPC URL to be used for communicating with the L2 chain.
+
+   * - ``--keystore-file PATH``
+     - Path to the keystore file.
+
+   * - ``--password PASSWORD``
+     - Password of the keystore file.
+
+   * - ``--l2-transaction-hash TX_HASH``
+     - Transaction hash of the submitted message that needs to be proven.
+
+   * - ``--custom-network PATH``
+     - Path to a file with custom L2 network configuration. This option is mainly used for development purposes.
--- a/relayer/README.md
+++ b/relayer/README.md
@@ -1,0 +1,59 @@
+# Beamer's cross-chain message relayer
+
+This software component provides a CLI that can be used to relay messages between two networks.
+
+## Setup
+
+Pre-requisites:
+
+- [Node.js >=18.x](https://nodejs.org/en)
+
+Start by installing the dependencies
+
+```
+yarn install
+```
+
+## Compile the TS code
+
+```
+yarn build
+```
+
+## Run
+
+There are two ways to run the CLI commands.
+
+1. By using the TS runtime enviroment `ts-node` (preferred in dev mode)
+
+```
+yarn ts-node ./src/cli.ts <command> [options]
+```
+
+2. By using node after running the [compilation step above](#compile-the-ts-code)
+
+```
+node ./build/src/cli.js <command> [options]
+```
+
+You can find more info on the available commands and their respective options [here](https://docs.beamerbridge.com/relayer).
+
+Additionally, these commands and options can be listed by invoking the CLI with the `--help` option.
+
+## Run unit tests
+
+```
+yarn test:unit
+```
+
+## Lint files
+
+```
+yarn lint
+```
+
+## Fix lint errors
+
+```
+yarn lint:fix
+```
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -14,6 +14,7 @@ Welcome to Beamer's documentation!
    contracts-api-reference.rst
    protocol.rst
    l1_resolution.rst
+   relayer.rst
    contributing.rst
    releasing.rst
    branching-strategy.rst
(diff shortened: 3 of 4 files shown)


relayer/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+yarn build
+```
+
+## Run
+
+There are two ways to run the CLI commands.
+
+1. By using the TS runtime enviroment `ts-node` (preffered in dev mode)
```
istankovic, line 27: Typo: should be `preferred` instead of `preffered`.

relayer/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+2. By using node after running the [compilation step above](#compile-the-ts-code)
+
+```
+node ./build/src/cli.js <command> [options]
+```
+
+You can find more info on the available commands and their respective options [here](https://docs.beamerbridge.com/relayer).
+
+If you forgot the commands & options, you can always consult yourself via the CLI by running it with `--help`.
```
istankovic, line 41: I'd drop the "forgetting" part and reformulate in a neutral tone, e.g
```
Available commands and options can be listed using the `--help` option.
```
or something like that.

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+  The agent utilizes these proofs in an optimistic way. If a challenge game ever occurs, or if a
+  fill was invalid, the agent has the chance to prove that a misbehavior occured and in the case of a challenge game,
+  get rewarded for it.  
+|
+| Since these **proofs** lay on a different chain than the chain where the requests originated from and where the challenge
+  game is running, we need a way to transport this **proof** from one chain to another in order to be able 
+  to prove that a certain state is correct. Thats where the cross-chain relayer comes into play.
+|
+| For instructions on how to setup, compile and run the relayer, make sure you take a look at the `/relayer/README.md` file.
+
```
istankovic, line 22: Why is `|` at the beginning of some lines above?

Could you please remove that and also the trailing whitespace in this file?

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+| This component is responsible for handling the cross-chain contract state updates. 
+|
+| It is mainly used by the Agent software when resolving requests as a result of a running challenge game
+  or for sending fill invalidation messages (L1 resolutions). However, it can also be used manually by running
+  the CLI commands.
+| If you haven't read about the role of :ref:`l1-resolutions` in Beamer, make sure you get familiar 
+| with this concept first.
+|
+| Basically, every **request fill** action creates a **proof** on the request destination chain.
```
istankovic, line 12: Nitpick: no need for `Basically` at the beginning.

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+  The agent utilizes these proofs in an optimistic way. If a challenge game ever occurs, or if a
+  fill was invalid, the agent has the chance to prove that a misbehavior occured and in the case of a challenge game,
+  get rewarded for it.
```
istankovic, line 15: To someone coming from outside, it would not be quite clear what is meant by `optimistic` above. I think we should clarify that a bit more (perhaps at the risk of repeating some things from the L1 resolution part).

E.g.
```
The agent utilizes these if and when needed. In normal operation, agent's claims will not be challenged so fill proofs are not needed, therefore the agent will not invoke the relayer.

However, if a challenge game occurs, the agent has to make use of the proof to ensure it wins the game. Typically, that is done by invoking the relayer at an appropriate moment, i.e. when the proof becomes available for execution on Ethereum L1.
```

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+  the CLI commands.
+| If you haven't read about the role of :ref:`l1-resolutions` in Beamer, make sure you get familiar 
+| with this concept first.
+|
+| Basically, every **request fill** action creates a **proof** on the request destination chain.
+  The agent utilizes these proofs in an optimistic way. If a challenge game ever occurs, or if a
+  fill was invalid, the agent has the chance to prove that a misbehavior occured and in the case of a challenge game,
+  get rewarded for it.  
+|
+| Since these **proofs** lay on a different chain than the chain where the requests originated from and where the challenge
```
istankovic, line 17: Is `lay` a typo? It looks like `lie` would be more fitting here. Or perhaps a rephrasing along the lines of `Since these proofs are created on a different...`

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+  game is running, we need a way to transport this **proof** from one chain to another in order to be able
```
istankovic, line 18: Here we use "this proof" (singular), while in the previous part of the sentence we use plural.

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+| with this concept first.
+|
+| Basically, every **request fill** action creates a **proof** on the request destination chain.
+  The agent utilizes these proofs in an optimistic way. If a challenge game ever occurs, or if a
+  fill was invalid, the agent has the chance to prove that a misbehavior occured and in the case of a challenge game,
+  get rewarded for it.  
+|
+| Since these **proofs** lay on a different chain than the chain where the requests originated from and where the challenge
+  game is running, we need a way to transport this **proof** from one chain to another in order to be able 
+  to prove that a certain state is correct. Thats where the cross-chain relayer comes into play.
```
istankovic, line 19: Typo: `Thats` -> `That's` or `That is`

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+
+* :ref:`command-relay` relays a fill proof or a fill invalidation message from chain A to chain B.
+* :ref:`command-prove-op-message` proves on L1 that a message exists on L2. Used only for messages traveling from chains that are based on the *OP Bedrock* stack.
+
+.. _command-relay:
+
+``relay``
+^^^^^^^^^^^^^^^^
+
+The ``relay`` command takes care of relaying a message from one chain to another (E2E).
```
istankovic, line 37: What is `E2E` here?

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+In cases where a message needs to be relayed from an "OP Bedrock"-like chain, then one first needs to run
```
istankovic, line 38: I would prefer if instead of `OP` we used `Optimism` everywhere in order to be fully clear what we are talking about.

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+the :ref:`command-prove-op-message` command, for the same message that needs to be relayed, before he can run the :ref:`command-relay` command.
```
istankovic, line 39: Since you are using the impersonal pronoun `one` in the previous part of the sentence, you should also use the same prnoun here, i.e. `before one can run...`.

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+     - RPC URL to be used for communicating with the base chain.
+
+   * - ``--l2-relay-to-rpc-url URL``
+     - RPC URL to be used for communicating with the destination chain (chain where the proof is travelling *to*).
+
+   * - ``--l2-relay-from-rpc-url URL``
+     - RPC URL to be used for communicating with the source chain (chain where the proof is travelling *from*).
+
+   * - ``--keystore-file PATH``
+     - Path to the account keyfile.
```
istankovic, line 58: We have decided to avoid usage of `keyfile` and instead use `keystore file`.
The other commands document this as 
```
Path to the keystore file.
```
It would be good to be in line with how the other commands do it, please see `docs/source/commands.rst`.

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+
+
+.. _command-prove-op-message:
+
+``prove-op-message``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The ``prove-op-message`` command is used only when a message needs to travel *from* "OP Bedrock"-like networks.
+On such networks, there is a need of proving the message upfront before it is treated as a message
+that should be included in a batch and submitted to L1 (after it reaches the finality period).
+The finality period starts counting only after the message was proven on L2.
```
istankovic, line 80: Nitpick: I would drop `counting` here (the finality period starts and ends, it does not count).

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+     - Path to a file with custom network configuration. This option is mainly used for development purposes.
+
+
+.. _command-prove-op-message:
+
+``prove-op-message``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The ``prove-op-message`` command is used only when a message needs to travel *from* "OP Bedrock"-like networks.
+On such networks, there is a need of proving the message upfront before it is treated as a message
+that should be included in a batch and submitted to L1 (after it reaches the finality period).
```
istankovic, line 79: How about
```
after it reaches the end of its finality period
```
or
```
after the proof is finalized on L2
```
?

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+     - RPC URL to be used for communicating with the base chain.
+
+   * - ``--l2-rpc-url URL``
+     - RPC URL to be used for communicating with the L2 chain.
+
+   * - ``--keystore-file PATH``
+     - Path to the account keyfile.
+
+   * - ``--password PASSWORD``
+     - Password of the keystore file.
```
istankovic, line 99: Similarly to above, it would be good to use the same docs for same options across all our commands.

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
...
+| For instructions on how to setup, compile and run the relayer, make sure you take a look at the `/relayer/README.md` file.
+
+Command reference
+-----------------
+
+The relayer CLI currently supports the following commands:
+
+
+* :ref:`command-relay` relays a fill proof or a fill invalidation message from chain A to chain B.
+* :ref:`command-prove-op-message` proves on L1 that a message exists on L2. Used only for messages traveling from chains that are based on the *OP Bedrock* stack.
```
istankovic, line 30: Nitpick: would it be possible to break this long line somehow?

istankovic (changes requested):
Some more work is needed still.

docs/source/relayer.rst
bilbeyt: I think lay is better as lie also means like some is cheating lying to somebody, but generally the best solution is just using ``Since these proofs are created on a different...`` like @istankovic suggested.

docs/source/relayer.rst
```diff
@@ -0,0 +1,105 @@
+Cross-chain message relayer
```
bilbeyt, line 1: Well in overview section you only mentioned our l1 resolution relay system but we are also using op chains proof mechanism with every fill, you may also include it here to overview.

docs/source/relayer.rst
GabrielBuragev: What about "This is ..."?

docs/source/relayer.rst
GabrielBuragev: I used it so i can preserve the line-breaks on some of the sentences.
For example, i wanted the sentence: "If you haven't read about the role ...." to stay in the same paragraph block but in a new line. When this paragraph block is not prefixed with "|", this sentence will continue on the same line as the previous sentence.

If you have a suggestion on how to avoid using "|" and achieve the desired thing, please let me know.

docs/source/relayer.rst
GabrielBuragev: `end-to-end = chain-to-chain`
Maybe i'll just write it as `end-to-end`

docs/source/relayer.rst
GabrielBuragev: Good point

docs/source/relayer.rst
GabrielBuragev: `after the proof is finalized on L2` sounds better. Thanks

docs/source/relayer.rst
GabrielBuragev: Uhm, not sure. I will give it a try.

docs/source/relayer.rst
istankovic: Yes, it's much better to be explicit in our docs, because we cannot assume that everyone reading it will be familiar with terms that we've been using for a long time. For us, E2E is pretty much clean, for anyone outside our team, there would be head scratching.

docs/source/relayer.rst
istankovic: > For example, i wanted the sentence: "If you haven't read about the role ...." to stay in the same paragraph block but in a new line. 

The question is, why would we want to do that?
Why not just start a new paragraph?

GabrielBuragev: I feel like that sentence should be kept with the others in that paragraph due to the context but if its on the same line the text becomes really hard to read. I noticed that in general with our whole documentation.

GabrielBuragev: But, i will remove it so we keep the same structure everywhere :)

bilbeyt (approved):

istankovic (approved):
Looks good, thanks!
(You'll need to get rid of the fixup commits, though.)

docs/source/relayer.rst
istankovic: I'm fine with whatever you choose here, it's just that it drew my attention due to line length, but in the end it's only a nitpick.

PR #2200, opened 2023-09-06
relayer: Wait until message has arrived on the destination RequestManager

### LINKED ISSUE #1867: Relayer: Wait until message has arrived on the destination RequestManager
# Rationale
An L1 resolution (coming from the `FillManager` on rollup B) is considered to be completed only after the message successfully arrived on the rollup A's `RequestManager` contract and an event (either `FillInvalidatedResolved` or `RequestResolved`) has been emitted by the same contract.

Currently, the responsibility of relaying the L1 resolution message (between the rollups) is handed over to the relayer program.
While the current relayer implementation does its job very well, there is one small step missing in its implementation. That is, checking if the message really arrived by scanning/listening on the events of the rollups A's `RequestManager` (for a message travelling in the direction Rollup B -> Rollup A) before the program returns/exits with a successful status code.  

We decided this way so we can separate the message relaying logic and the business logic of the Agent software. There is no reason for this logic to not be incorporated into the relayer. In fact, this will provide completeness to the relayer software.
 
# Implementation
For each network-specific relayer service an additional step is implemented for waiting on the event to show up on the destination `RequestManager`.

# Acceptance Criteria
- [ ] Each relayer service awaits for `FillInvalidatedResolved` or `RequestResolved` to show up before exiting the program with a successful status code

manuelwedler:
closes #1867 

fetchFirstMatchingEvent is taken from frontend. Added the possibility to check for event argument values which are not indexed.


relayer/src/services/types.ts
```diff
@@ -106,7 +143,88 @@ export abstract class BaseRelayerService {
...
+    const resolver = Resolver__factory.connect(resolverAddress, this.l1RpcProvider);
+    const requestManagerAddress = (await resolver.sourceChainInfos(this.l2ChainId)).requestManager;
+    const requestManager = RequestManager__factory.connect(
+      requestManagerAddress,
+      this.l2RpcProvider,
+    );
+
+    const timeout = new Promise<undefined>((resolve) => setTimeout(resolve, timeoutMs)).then(
+      () => {
+        requestManager.removeAllListeners();
```
GabrielBuragev, line 161: Nitpick: I think we can move this additional clean-up step inside a function that will be provided to the `setTimeout()` call (where the call to `resolve` will happen too).

relayer/src/common/events/utils.ts
```diff
@@ -0,0 +1,52 @@
+import type { Event, EventFilter } from "ethers";
+
+function sleep(time: number) {
+  return new Promise((resolve) => setTimeout(resolve, time));
```
GabrielBuragev, line 4: We already have a `sleep` util function defined inside `src/common/util.ts`

relayer/src/cli/programs/relay.ts
```diff
@@ -92,5 +92,18 @@ export class RelayerProgram {
         await this.l2RelayerTo.finalizeStep.execute(l1TransactionHash);
       }
     }
+
+    const relayParameters = await this.l2RelayerFrom.parseFillEventDataFromTxHash(
```
GabrielBuragev, line 96: Can we print a status message here before this additional step starts?
Something like: `Verifying successful relay operation...`

Would be nice for debugging purposes.

GabrielBuragev (changes requested):
A few nitpicks and optimisation suggestions. The main logic and placement of the code looks perfect.

Btw, did you test this either on testnet or on mainnet? I would recommend testing it with at least one FillInvalidation message either on testnet or on mainnet.

GabrielBuragev (approved):
Looks good now.

